PR #201, opened 2018-01-26
Initial PR for calling PaymentsFlow from Offers

dparikh:
TODO:
Add unit tests

### DIFF
--- a/src/runtime/offers-flow.js
+++ b/src/runtime/offers-flow.js
@@ -20,6 +20,20 @@ import {
   PayStartFlow,
   PayCompleteFlow,
 } from './pay-flow';
+import {assert} from '../utils/log';
+
+/**
+ * @typedef {{
+ *   code: string,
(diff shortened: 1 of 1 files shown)


src/runtime/offers-flow.js
```diff
@@ -30,41 +46,53 @@ const offersUrl = '$frontend$/subscribewithgoogleclientui/offersiframe';
...
-        offersUrl,
+        OFFERS_URL,
         {
-          'publicationId': pageConfig.getPublicationId(),
-          'label': pageConfig.getLabel(),
+          'publicationId': deps.pageConfig().getPublicationId(),
+          'label': deps.pageConfig().getLabel(),
         });
+
+    PayCompleteFlow.configurePending(this.deps_);
```
dvoytenko: @dparikh quick note: `configurePending` can only be called from main runtime since only one callback per page is needed.

PR #242, opened 2018-02-13
Support additional {details} format for pay response

dvoytenko:
/to @kaipakartik

### DIFF
--- a/src/runtime/pay-flow.js
+++ b/src/runtime/pay-flow.js
@@ -166,29 +166,33 @@ export function validatePayResponse(port) {
  * @return {!SubscribeResponse}
  */
 export function parseSubscriptionResponse(data) {
+  let swgData = null;
   let raw = null;
   if (data) {
     if (typeof data == 'string') {
       raw = /** @type {string} */ (data);
     } else {
       // Assume it's a json object in the format:
-      // `{integratorClientCallbackData: "..."}`.
-      let json = /** @type {!Object} */ (data);
-      if ('details' in json) {
-        json = /** @type {!Object} */ (json['details']);
+      // `{integratorClientCallbackData: "..."}` or `{swgCallbackData: "..."}`.
+      const json = /** @type {!Object} */ (data);
+      if ('swgCallbackData' in json) {
+        swgData = /** @type {!Object} */ (json['swgCallbackData']);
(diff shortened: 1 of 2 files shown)


src/runtime/pay-flow.js
```diff
@@ -173,7 +173,10 @@ export function parseSubscriptionResponse(data) {
     } else {
       // Assume it's a json object in the format:
       // `{integratorClientCallbackData: "..."}`.
-      const json = /** @type {!Object} */ (data);
```
kaipakartik: You weren't base64 decoding the data anywhere?

src/runtime/pay-flow.js
```diff
@@ -166,26 +166,33 @@ export function validatePayResponse(port) {
...
+      const parsed = parseJson(raw);
+      swgData = parsed['swgCallbackData'];
+    }
   }
   if (!swgData) {
     throw new Error('unexpected payment response');
   }
+  raw = JSON.stringify(/** @type {!JsonObject} */ (swgData));
   return new SubscribeResponse(
       raw,
```
kaipakartik: Return the json object instead of Json.stringify?
It's fine to do this in another pull request if needed

src/runtime/pay-flow.js
```diff
@@ -166,26 +166,33 @@ export function validatePayResponse(port) {
...
   }
-
-  let swgData = null;
-  if (raw) {
-    const parsed = parseJson(atob(raw));
-    swgData = parsed['swgCallbackData'];
+  if (raw && !swgData) {
+    raw = atob(raw);
+    if (raw) {
+      const parsed = parseJson(raw);
```
kaipakartik: Why not Json.Parse?

src/runtime/pay-flow.js
dvoytenko: By current format, `raw` is a string to simplify posting it to the 3p site.

src/runtime/pay-flow.js
dvoytenko: That's pretty much that - with some nuances around types.

dvoytenko: Also. That call is inlined into `JSON.parse()` in the binary.

PR #250, opened 2018-02-25
Remove additional param 'showCloseAction' on dialog

dparikh:


src/ui/ui.css
```diff
@@ -23,8 +23,7 @@ body {
   position: absolute;
   left: 0;
   top: 0;
-  padding: 16px;
-  margin: 0;
+  margin: 16px;
```
dparikh: In firefox, when `padding: 16px;`, the Close button does not show with 16px padding on left.

dvoytenko: The padding was used not for positioning, but to enlarge the tap target IIRC. Margin doesn't give us bigger targets.

dparikh:
Reverted the ui.css changes, since the CSS is now moved from here.

PR #252, opened 2018-02-27
Offers page: keep alive until canceled or completed

dparikh:


src/components/dialog-manager.js
```diff
@@ -53,6 +53,10 @@ export class DialogManager {
    * @return {!Promise}
    */
   openView(view) {
+    view.acceptResult().catch(reason => {
+      this.completeView(view);
```
dvoytenko: For the time being, can we only close when "cancel", e.g. `reason.name == 'AbortError'`? It's useful for debugging to keep the view option for other failures.

src/components/view.js
```diff
@@ -48,6 +48,12 @@ export class View {
     // Do nothing by default. Override if needed.
   }
 
+  /*
+   * Accept the result.
+   * @return {!Promise}
+   */
+  acceptResult() {}
```
dvoytenko: I think we need a `whenComplete` as `!Promise` here instead. Otherwise I don't see how we can reconcile types give that acceptResult is `!Promise<ActivityResult>` or what not.

src/ui/activity-iframe-view.js
```diff
@@ -70,6 +70,9 @@ export class ActivityIframeView extends View {
     /** @private {?web-activities/activity-ports.ActivityIframePort} */
     this.port_ = null;
 
+    /* @private {Promise<!Object<string, string|number>} */
```
dvoytenko: It's not a promise.

src/components/dialog-manager.js
dparikh: Done.

src/ui/activity-iframe-view.js
dparikh: Done

src/components/view.js
dparikh: Done.

src/components/dialog-manager.js
```diff
@@ -53,6 +53,12 @@ export class DialogManager {
    * @return {!Promise}
    */
   openView(view) {
+    view.whenComplete().catch(reason => {
+      if (reason.name && reason.name === 'AbortError') {
```
dvoytenko: No need for `reason.name` part, just `if (reaso.name == 'AbortError') {}`

src/ui/activity-iframe-view.js
```diff
@@ -129,6 +159,14 @@ export class ActivityIframeView extends View {
     return this.promise_;
   }
 
+  /**
+   * Completes the flow.
+   * @return {!Promise}
+   */
+  whenComplete() {
+    return this.whenCompletePromise_;
```
dvoytenko: Simply return `this.promise_` here. There's no need for a new object, we just needed API separation to preserve the types.

src/components/dialog-manager.js
dparikh: Done

src/ui/activity-iframe-view.js
dparikh: Got it. Done

PR #261, opened 2018-03-03
Move build-system/eslint-rules into a plugin

dparikh:
Inspired by: https://github.com/ampproject/amphtml/pull/13474

Some facts:
1. We use `gulp-eslint` as a wrapper around `eslint` to run our lint checks via `gulp lint`
2. We have some custom linter rules in `build-system/eslint-rules` (in addition to the rules in `.eslintrc`)
3. `gulp-eslint` uses the `rulePaths` option to specify custom rules
4. `rulePaths` is translated to a `--rulesdir` argument to `eslint`
5. `--rulesdir` is being deprecated by `eslint` in favor of [eslint plugins](https://eslint.org/docs/developer-guide/working-with-plugins) 
6. Our current way of specifying custom rules breaks the `eslint` [plugins](https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint) of popular IDEs like VSCode

This PR does the following:
1. Gets rid of deprecated flags and options for custom rules
2. Exports the `amphtml` custom lint rules into an internal plugin called `eslint-plugin-amphtml-internal`
3. Applies all custom `eslint` rules directly from `.eslintrc` via the new internal plugin
4. This enables a fully working `eslint` [plugin](https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint) in IDEs like VSCode (with no additional settings, since all rules are pulled in by `.eslintrc`)

With this PR, users of the [VS Code](https://code.visualstudio.com/) editor (and likely, other editors that support eslint integration) can install the [ESLint](https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint) extension and see real-time linting of source code, a.k.a. red squiggly lines pointing out lint errors as you type. This includes  support for custom `swg-internal` rules.

Steps for VS Code:
1. If you haven't already done so, sync your local `master` branch and run `yarn`
2. Install VS Code from https://code.visualstudio.com
3. Enable the ESLint extension (View -> Extensions -> ESLint -> Install -> Reload)
4. Start editing a js file to see the linter in action


package.json
```diff
@@ -26,7 +26,8 @@
   },
   "dependencies": {
     "promise-pjs": "1.1.3",
-    "web-activities": "1.4.0"
+    "web-activities": "1.4.0",
+    "yarn": "^1.5.1"
```
dvoytenko: Why is it in deps? Can it go to `devDependencies` instead?

dparikh: Yes, Moved.

dparikh: In fact, this is not required at all. Removed.

package.json
```diff
@@ -26,7 +26,7 @@
   },
   "dependencies": {
     "promise-pjs": "1.1.3",
-    "web-activities": "1.4.0"
+    "web-activities": "1.4.0",
```
dvoytenko: I don't think you are allowed terminating commas in package.json.

dparikh: Done.

PR #263, opened 2018-03-04
Switch from self.SUBSCRIPTIONS to self.SWG

dvoytenko:


src/runtime/runtime-test.js
```diff
@@ -48,6 +48,128 @@ describes.realWin('installRuntime', {}, env => {
...
+
+describes.realWin('installRuntime legacy', {}, env => {
+  let win;
+
+  beforeEach(() => {
+    win = env.win;
+  });
+
   function dep(callback) {
     (win.SUBSCRIPTIONS = win.SUBSCRIPTIONS || []).push(callback);
```
dparikh: should this be changed to win.SWG?

dvoytenko: It tests both versions at this time.

PR #280, opened 2018-03-07
Switch to /swg/_/api urls

dvoytenko:


dvoytenko:
Merging per cl/188236436

PR #282, opened 2018-03-08
GetOffers API

dvoytenko:


src/runtime/runtime.js
```diff
@@ -387,6 +397,11 @@ export class ConfiguredRuntime {
     this.callbacks_.setOnEntitlementsResponse(callback);
   }
 
+  /** @override */
+  getOffers(opt_options) {
```
dparikh: duplicate of line 237?

dvoytenko: No, that's the distinction between pending and configured runtime. The pending runtime is a simple proxy.

PR #285, opened 2018-03-09
Toast iframe implementation

dparikh:
TODO:

Use the 'source' parameter passed to the Activity to show correct publisher.

### DIFF
--- a/src/runtime/entitlements-manager.js
+++ b/src/runtime/entitlements-manager.js
@@ -61,15 +61,6 @@ export class EntitlementsManager {
 
     /** @private @const {!./storage.Storage} */
     this.storage_ = deps.storage();
-
-    /** @private @const {string} */
-    this.toastSrc_ = feUrl('/toastiframe');
-
-    /** @private @const {string} */
-    this.publicationId_ = this.deps_.pageConfig().getPublicationId();
-
-    /** @private @const {!Object<string, ?>} */
-    this.toastArgs_ = feArgs({'publicationId': this.publicationId_});
   }
 
   /**
@@ -194,8 +185,11 @@ export class EntitlementsManager {
    */
   showToast_(entitlement) {
     const source = entitlement.source || 'google';
-    const args = Object.assign(this.toastArgs_, {'source': source});
-    return new Toast(this.deps_, this.toastSrc_, args).open();
+
+    return new Toast(this.deps_, feUrl('/toastiframe'), feArgs({
+      'publicationId': this.publicationId_,
+      'source': source,
+    })).open();
   }
 
   /**
--- a/src/runtime/entitlements-manager.js
+++ b/src/runtime/entitlements-manager.js
@@ -183,7 +183,7 @@ export class EntitlementsManager {
    * @private
    */
   showToast_(entitlement) {
-    const source = entitlement.source || 'Google';
+    const source = entitlement.source || 'google';
     return new Toast(this.deps_, {'source': source}).open();
   }
 
--- a/src/runtime/entitlements-manager.js
+++ b/src/runtime/entitlements-manager.js
@@ -18,6 +18,7 @@ import {Entitlement, Entitlements} from '../api/entitlements';
 import {JwtHelper} from '../utils/jwt';
 import {Toast} from '../ui/toast';
 import {serviceUrl} from './services';
+import {feArgs, feUrl} from '../runtime/services';
 
 const SERVICE_ID = 'subscribe.google.com';
 const TOAST_STORAGE_KEY = 'toast';
@@ -60,6 +61,15 @@ export class EntitlementsManager {
 
     /** @private @const {!./storage.Storage} */
     this.storage_ = deps.storage();
+
+    /** @private @const {string} */
+    this.toastSrc_ = feUrl('/toastiframe');
+
+    /** @private @const {string} */
+    this.publicationId_ = this.deps_.pageConfig().getPublicationId();
+
+    /** @private @const {!Object<string, ?>} */
+    this.toastArgs_ = feArgs({'publicationId': this.publicationId_});
   }
 
   /**
@@ -184,7 +194,8 @@ export class EntitlementsManager {
    */
   showToast_(entitlement) {
     const source = entitlement.source || 'google';
-    return new Toast(this.deps_, {'source': source}).open();
+    const args = Object.assign(this.toastArgs_, {'source': source});
+    return new Toast(this.deps_, this.toastSrc_, args).open();
   }
 
   /**
--- a/src/runtime/entitlements-manager.js
+++ b/src/runtime/entitlements-manager.js
@@ -202,7 +202,7 @@ export class EntitlementsManager {
   fetch_() {
     const url = serviceUrl(
         '/publication/' +
-        encodeURIComponent(this.config_.getPublicationId()) +
+        encodeURIComponent(this.publicationId_) +
         '/entitlements');
     return this.fetcher_.fetchCredentialedJson(url).then(json => {
       const signedData = json['signedEntitlements'];
--- a/src/runtime/entitlements-manager.js
+++ b/src/runtime/entitlements-manager.js
@@ -183,20 +183,8 @@ export class EntitlementsManager {
    * @private
    */
   showToast_(entitlement) {
-    const toast = new Toast(this.win_, {
-      text:
-          (entitlement.source || 'google') == 'google' ?
-          'Access via Google Subscriptions' :
-          // TODO(dvoytenko): display name instead.
-          'Access via [' + entitlement.source + ']',
-      action: {
-        label: 'View',
-        handler: function() {
-          // TODO(dparikh): Implementation.
-        },
-      },
-    });
-    toast.open();
+    const source = entitlement.source || 'Google';
+    return new Toast(this.deps_, {'source': source}).open();
   }
 
   /**
--- a/src/runtime/entitlements-manager.js
+++ b/src/runtime/entitlements-manager.js
@@ -41,6 +41,9 @@ export class EntitlementsManager {
     /** @private @const {!../model/page-config.PageConfig} */
     this.config_ = config;
 
+    /** @private @const {string} */
+    this.publicationId_ = this.config_.getPublicationId();
+
     /** @private @const {!./fetcher.Fetcher} */
     this.fetcher_ = fetcher;
 
--- a/src/ui/toast-test.js
+++ b/src/ui/toast-test.js
@@ -21,7 +21,7 @@ import {PageConfig} from '../model/page-config';
 import {getStyle} from '../utils/style';
 import * as sinon from 'sinon';
 
-const src = '$frontend$/swglib/toastiframe';
+const src = '$frontend$/swglib/toastiframe?_=_';
 
 const args = {
   _client: 'SwG $internalRuntimeVersion$',
--- a/src/ui/toast-test.js
+++ b/src/ui/toast-test.js
@@ -44,7 +44,6 @@ describes.realWin('Toast', {}, env => {
 
   it('should have created Notification View', function() {
     const iframe = toast.getElement();
-    console.log('iframe', toast);
     toast.whenReady().then(() => {
       expect(iframe.nodeType).to.equal(1);
       expect(iframe.nodeName).to.equal('IFRAME');
--- a/src/ui/toast-test.js
+++ b/src/ui/toast-test.js
@@ -34,7 +34,7 @@ describes.realWin('Toast', {}, env => {
     pageConfig = new PageConfig('pub1:label1');
     runtime = new ConfiguredRuntime(win, pageConfig);
     activitiesMock = sandbox.mock(runtime.activities());
-    toast = new Toast(runtime, {'source': 'Google'});
+    toast = new Toast(runtime, {'source': 'google'});
     toast.whenReady = () => Promise.resolve();
     port = new ActivityPort();
     port.onResizeRequest = () => {};
@@ -49,8 +49,6 @@ describes.realWin('Toast', {}, env => {
       expect(iframe.nodeName).to.equal('IFRAME');
 
       expect(getStyle(iframe, 'opacity')).to.equal('1');
-      expect(getStyle(iframe, 'font-family'))
-          .to.equal('"Google sans", sans-serif');
       expect(getStyle(iframe, 'bottom')).to.equal('0px');
       expect(getStyle(iframe, 'display')).to.equal('block');
 
@@ -67,8 +65,7 @@ describes.realWin('Toast', {}, env => {
         {
           _client: 'SwG $internalRuntimeVersion$',
           publicationId: 'pub1',
-          showNative: false,
-          'source': 'Google',
+          source: 'google',
         })
         .returns(Promise.resolve(port));
     return toast.open();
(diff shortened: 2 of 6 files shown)


src/runtime/entitlements-manager.js
```diff
@@ -183,20 +183,8 @@ export class EntitlementsManager {
...
-          'Access via [' + entitlement.source + ']',
-      action: {
-        label: 'View',
-        handler: function() {
-          // TODO(dparikh): Implementation.
-        },
-      },
-    });
-    toast.open();
+    const source = entitlement.source || 'Google';
```
dvoytenko: I think it's lowercase actually.

src/ui/toast-test.js
```diff
@@ -15,78 +15,62 @@
...
+      expect(iframe.nodeType).to.equal(1);
+      expect(iframe.nodeName).to.equal('IFRAME');
 
-    expect(getStyle(iframe, 'opacity')).to.equal('1');
-    expect(getStyle(iframe, 'font-family'))
-        .to.equal('"Google sans", sans-serif');
-    expect(getStyle(iframe, 'bottom')).to.equal('0px');
-    expect(getStyle(iframe, 'display')).to.equal('block');
+      expect(getStyle(iframe, 'opacity')).to.equal('1');
+      expect(getStyle(iframe, 'font-family'))
```
dvoytenko: Does font actually matter here?

src/ui/toast-test.js
```diff
@@ -15,78 +15,62 @@
...
-    expect(detail.getAttribute('class')).to.equal('swg-detail');
-    expect(detail.getAttribute('aria-label')).to.equal(spec.action.label);
-    expect(detail.textContent).to.equal(spec.action.label);
+    activitiesMock.expects('openIframe').withExactArgs(
+        sinon.match(arg => arg.tagName == 'IFRAME'),
+        '$frontend$/swglib/toastiframe',
+        {
+          _client: 'SwG $internalRuntimeVersion$',
+          publicationId: 'pub1',
+          showNative: false,
```
dvoytenko: There shouldn't be `showNative` here.

src/ui/toast-test.js
```diff
@@ -15,78 +15,62 @@
...
+          'source': 'Google',
```
dvoytenko: Ditto: lowercase.

src/ui/toast.js
```diff
@@ -53,121 +40,99 @@ export const toastImportantStyles = {
...
 /**
  * The class Notification toast.
  */
 export class Toast {
 
   /**
-   * @param {!Window} win
-   * @param {!ToastSpecDef} spec
+   * @param {!../runtime/deps.DepsDef} deps
+   * @param {{'source': string}} args
```
dvoytenko: `!Object<string, ?>`

dvoytenko: I'd say let's pass URL here as well.

src/ui/toast.js
```diff
@@ -53,121 +40,99 @@ export const toastImportantStyles = {
...
-
-    this.addItems_(iframeDoc, iframeBody);
-
-    return this;
+  buildToast_() {
+    return this.activityPorts_.openIframe(
+        this.iframe_, feUrl('/toastiframe'),
+        feArgs({
+          'publicationId': this.deps_.pageConfig().getPublicationId(),
+          'showNative': this.deps_.callbacks().hasSubscribeRequestCallback(),
```
dvoytenko: Not needed.

src/ui/toast.js
```diff
@@ -53,121 +40,99 @@ export const toastImportantStyles = {
...
-    injectStyleSheet(iframeDoc, TOAST_CSS);
-
-    this.addItems_(iframeDoc, iframeBody);
-
-    return this;
+  buildToast_() {
+    return this.activityPorts_.openIframe(
+        this.iframe_, feUrl('/toastiframe'),
+        feArgs({
+          'publicationId': this.deps_.pageConfig().getPublicationId(),
```
dvoytenko: I think we should just pass these all from outside as constructor args.

src/ui/toast.js
```diff
@@ -53,121 +40,99 @@ export const toastImportantStyles = {
...
+    this.iframe_ =
+        /** @type {!HTMLIFrameElement} */ (
+            createElement(this.doc_, 'iframe', iframeAttributes));
 
-    setImportantStyles(this.iframe_.getElement(), toastImportantStyles);
-    setStyles(this.iframe_.getElement(), topFriendlyIframePositionStyles);
+    setImportantStyles(this.iframe_, toastImportantStyles);
+    setStyles(this.iframe_, topFriendlyIframePositionStyles);
+
+            /** @private @const {!Promise} */
```
dvoytenko: indent

src/ui/toast.js
```diff
@@ -53,121 +40,99 @@ export const toastImportantStyles = {
...
+    });
   }
 
-  /**
-   * Gets the attached iframe instance.
-   * @return {!FriendlyIframe}
-   */
-  getIframe() {
-    return this.iframe_;
+  getElement() {
```
dvoytenko: jsdoc

src/ui/toast.js
```diff
@@ -53,121 +40,99 @@ export const toastImportantStyles = {
...
+    return this.iframe;
   }
 
   /**
-   * Gets the Iframe element.
-   * @return {!HTMLIFrameElement}
+   * When promise is resolved.
```
dvoytenko: Rewrite the description or remove it.

src/ui/toast.js
```diff
@@ -53,121 +40,99 @@ export const toastImportantStyles = {
...
+   * @return {!Promise}
    */
-  getElement() {
-    return this.iframe_.getElement();
+  whenReady() {
```
dvoytenko: Do we have to expose this method? It seems a bit ambiguous given that there are several notions of readiness. I think the promise returned by `open()` should be enough?

src/ui/toast.js
```diff
@@ -53,121 +40,99 @@ export const toastImportantStyles = {
...
+          'publicationId': this.deps_.pageConfig().getPublicationId(),
+          'showNative': this.deps_.callbacks().hasSubscribeRequestCallback(),
+          'source': this.args_.source,
+        })).then(port => {
+          resetStyles(this.iframe_, ['height']);
+          setImportantStyles(this.iframe_, {
+            'animation': 'swg-notify 1s ease-out normal backwards, '
+                  + 'swg-notify-hide 1s ease-out 7s normal forwards',
+          });
+          return port.acceptResult();
```
dvoytenko: I think there's no need to return `port.acceptResult()` here - it will never return anythign useful.

src/ui/toast.js
```diff
@@ -53,121 +40,99 @@ export const toastImportantStyles = {
...
+    return this.activityPorts_.openIframe(
+        this.iframe_, feUrl('/toastiframe'),
+        feArgs({
+          'publicationId': this.deps_.pageConfig().getPublicationId(),
+          'showNative': this.deps_.callbacks().hasSubscribeRequestCallback(),
+          'source': this.args_.source,
+        })).then(port => {
+          resetStyles(this.iframe_, ['height']);
+          setImportantStyles(this.iframe_, {
+            'animation': 'swg-notify 1s ease-out normal backwards, '
```
dvoytenko: Can animation times be made shorter? We usually try to keep to 0.3s.

src/ui/toast.js
```diff
@@ -53,121 +40,99 @@ export const toastImportantStyles = {
...
+                  + 'swg-notify-hide 1s ease-out 7s normal forwards',
```
dvoytenko: Once the toast is done animating out, shouldn't it be removed from DOM?

src/runtime/entitlements-manager.js
```diff
@@ -183,20 +183,8 @@ export class EntitlementsManager {
...
+    return new Toast(this.deps_, {'source': source}).open();
```
dvoytenko: Path URL and all the rest of args here.

src/ui/toast-test.js
dparikh: Not really. Removed.

src/runtime/entitlements-manager.js
dparikh: Done

src/ui/toast-test.js
dparikh: Unit tests failing without showNative.

src/ui/toast-test.js
dparikh: Done

src/ui/toast-test.js
dparikh: Done

src/ui/toast.js
dparikh: linter complaining when using Object<string, ?>
when trying to refer: args.source.

src/ui/toast.js
dparikh: Done

src/ui/toast.js
dparikh: Done

src/ui/toast.js
dparikh: Done

src/ui/toast.js
dparikh: Removed.

src/ui/toast.js
dparikh: Removed.

src/ui/toast.js
dparikh: .3 seems little quicker/shorter since the height of the toast is only 48px.
What about .5 sec.?

src/ui/toast.js
dparikh: Removed whenReady() function call.

src/ui/toast.js
dvoytenko: You have to use `args['source']`. But please refer to my follow up comments on passing all arguments together as a single map here.

src/ui/toast.js
dvoytenko: Let's try .5

src/ui/toast.js
dparikh: Ack.

src/ui/toast.js
dparikh: Done

src/ui/toast.js
dparikh: Done

src/runtime/entitlements-manager.js
```diff
@@ -60,6 +61,15 @@ export class EntitlementsManager {
 
     /** @private @const {!./storage.Storage} */
     this.storage_ = deps.storage();
+
+    /** @private @const {string} */
+    this.toastSrc_ = feUrl('/toastiframe');
```
dvoytenko: These are rarely used. I recommend you just build them right before showing the toast - no values to have them as state params.

src/runtime/entitlements-manager.js
```diff
@@ -183,20 +193,9 @@ export class EntitlementsManager {
...
-        label: 'View',
-        handler: function() {
-          // TODO(dparikh): Implementation.
-        },
-      },
-    });
-    toast.open();
+    const source = entitlement.source || 'google';
+    const args = Object.assign(this.toastArgs_, {'source': source});
+    return new Toast(this.deps_, this.toastSrc_, args).open();
```
dvoytenko: Basically, just: 
```
return new Toast(this.deps_, feUrl('/toastiframe'), feArgs({
  'publicationId': ...,
  'source': ...,
}))
```

src/ui/toast-test.js
```diff
@@ -15,78 +15,67 @@
...
-} from '../utils/style';
+import {ActivityPort} from 'web-activities/activity-ports';
+import {ConfiguredRuntime} from '../runtime/runtime';
+import {PageConfig} from '../model/page-config';
+import {getStyle} from '../utils/style';
+import * as sinon from 'sinon';
+
+const src = '$frontend$/swglib/toastiframe';
+
+const args = {
```
dvoytenko: This should go to `entitlements-manager-test.js`

src/ui/toast-test.js
```diff
@@ -15,78 +15,67 @@
...
-  getStyle,
-  googleFontsUrl,
-} from '../utils/style';
+import {ActivityPort} from 'web-activities/activity-ports';
+import {ConfiguredRuntime} from '../runtime/runtime';
+import {PageConfig} from '../model/page-config';
+import {getStyle} from '../utils/style';
+import * as sinon from 'sinon';
+
+const src = '$frontend$/swglib/toastiframe';
```
dvoytenko: This should go to `entitlements-manager-test.js`

src/ui/toast.js
```diff
@@ -53,121 +39,96 @@ export const toastImportantStyles = {
...
 
-    /** @private @const {!ToastSpecDef} */
-    this.spec_ = spec;
+    /** @private @const {!web-activities/activity-ports.ActivityPorts} */
+    this.activityPorts_ = deps.activities();
 
-    /** @private @const {!FriendlyIframe} */
-    this.iframe_ = new FriendlyIframe(this.doc_, {'class': 'swg-toast'});
+    /** @private @const {!../runtime/deps.DepsDef} */
+    this.deps_ = deps;
```
dvoytenko: Remove

src/ui/toast.js
```diff
@@ -53,121 +39,96 @@ export const toastImportantStyles = {
...
-      childElements.push(linkButton);
-    }
-
-    // Create container element and add 'label' and/or 'linkButton' to it.
-    this.container_ = createElement(iframeDoc, 'div', {
-      'class': 'swg-toast-container',
-    }, childElements);
-
-    iframeBody.appendChild(this.container_);
+  close() {
```
dvoytenko: So, I'm still missing this answer: after animation is done animating out, the toast should somehow be removed from DOM. Where does it happen?

src/runtime/entitlements-manager.js
dparikh: Moved

src/runtime/entitlements-manager.js
dparikh: Done

src/ui/toast.js
dparikh: Done

src/ui/toast.js
dparikh: Done. Added a callback after 8 seconds to remove the toast from dom. (Since toast is shown for 7 seconds).

src/runtime/entitlements-manager.js
```diff
@@ -183,20 +184,12 @@ export class EntitlementsManager {
...
-        handler: function() {
-          // TODO(dparikh): Implementation.
-        },
-      },
-    });
-    toast.open();
+    const source = entitlement.source || 'google';
+
+    return new Toast(this.deps_, feUrl('/toastiframe'), feArgs({
+      'publicationId': this.publicationId_,
```
dvoytenko: Hmm. Is there such a variable in this class? I don't recall and not seeing it added. Perhaps something also worth to cover in tests?

dparikh: Updated.

dvoytenko: One request. There are uses in this class with `this.config_.getPublicationId()`. Since you added this as a variable, please either switch to this variable, or use `this.config_.getPublicationId()` everywhere.

dparikh: Found only one instance, replaced with `this.publicationId_`

PR #289, opened 2018-03-09
Setup caching rules for FE URLs

dvoytenko:


src/utils/url.js
```diff
@@ -177,6 +177,31 @@ export function parseQueryString(query) {
...
+ * @param {string} param
+ * @param {string} value
+ * @return {string}
+ */
+export function addQueryParam(url, param, value) {
+  const queryIndex = url.indexOf('?');
+  const fragmentIndex = url.indexOf('#');
+  let fragment = '';
+  if (fragmentIndex != -1) {
+    fragment = fragmentIndex != -1 ? url.substring(fragmentIndex) : '';
```
dparikh: this line is executed only if `fragmentIndex != -1`
so could just set: `fragment = url.substring(fragmentIndex);`

dvoytenko: Thanks! Missed it. Fixed.

PR #298, opened 2018-03-15
Preconnect payments

dvoytenko:


src/runtime/pay-flow.js
```diff
@@ -33,6 +33,13 @@ const PAY_REQUEST_ID = 'swg-pay';
  */
 export class PayStartFlow {
 
+  /**
+   * @param {!../utils/preconnect.Preconnect} pre
+   */
+  static preconnect(pre) {
+    pre.prefetch(feUrl('/pay'));
```
cramforce: Please make sure this actually results in cache hits. Is this for an iframe? I've seen this not work before (at least in terms of not coalescing request if it hasn't come back by iframe usage.`prefetch` is semantically for future navigations.

I'd recommend to also preconnect

dvoytenko: This is for a plain navigation and I definitely see cache hits everywhere, except Safari. But Safari doesn't do the fetch anyway - so not a net-negative.

Do you know if I can do `<link rel="preconnect prefetch">`?

cramforce: Yes you can! Sounds good for plain navigation!!!

dvoytenko: Thanks! Updated.

PR #312, opened 2018-03-22
Parse Microdata to discover page config

sohanirao:


src/model/page-config-resolver-test.js
```diff
@@ -266,7 +270,143 @@ describes.realWin('PageConfigResolver', {}, env => {
...
+      readyState = 'loading';
+      Object.defineProperty(doc, 'readyState', {get: () => readyState});
+    });
+
+    function getTestDiv(options) {
+      const access = (options['isAccessibleForFree'] != undefined);
+      const productId = options['productId'];
+      const topItemtype = options['newsArticleType'];
+      const divElement = createElement(doc, 'div');
+      if (!!topItemtype) {
```
dvoytenko: Just `if (topItemType) {}`. Here and below.

src/model/page-config-resolver.js
```diff
@@ -291,6 +358,64 @@ class MicrodataParser {
     this.win_ = win;
   }
 
+  discoverAccess_(prop) {
+    if (prop.getAttribute('itemprop') == 'isAccessibleForFree') {
+      const value = prop.content || prop.textContent;
+      if (typeof value == 'boolean') {
```
dvoytenko: Can it ever really be a `boolean`?

src/model/page-config-resolver.js
```diff
@@ -291,6 +358,64 @@ class MicrodataParser {
...
+    }
+    return null;
+  }
+
+  discoverProductId_(prop) {
+    if (prop.getAttribute('itemprop') == 'productID')	{
+      return prop.content || prop.textContent;
+    }
+    return null;
+  };
```
dvoytenko: no semicolon

src/model/page-config-resolver.js
```diff
@@ -291,6 +358,64 @@ class MicrodataParser {
...
+          if (element.getAttribute('itemtype') == 'http://schema.org/NewsArticle') {
+            const props = element.querySelectorAll('[itemprop]');
+            const items = [];
+            props.forEach(prop => {
+              let type = 'default';
+              if (prop.matches('[itemscope]')) {
+                type = prop.getAttribute('itemtype');
+              }
+              const item = {
+                'type': type,
```
dvoytenko: Why are these properties in quotes?

src/model/page-config-resolver.js
```diff
@@ -291,6 +358,64 @@ class MicrodataParser {
...
+    }
+    return null;
+  };
+
+  tryExtractConfig_(doc) {
+    let itemId = 0;
+    const results = [];
+    doc.querySelectorAll('[itemscope]')
+        .forEach(element => {
+          if (element.getAttribute('itemtype') == 'http://schema.org/NewsArticle') {
```
dvoytenko: Can we combine `itemtype` with selector above? That'd be a lot faster.

src/model/page-config-resolver.js
dvoytenko: JSON-LD is in the JSON format that has a native support for booleans. But microdata is parsed from attributes/text content, so my question was how is boolean possible there?

src/model/page-config-resolver.js
sohanirao: What if itemscope is defined in the parent div and itemtype is defined as a meta of a child div? We will end up missing it.

dvoytenko: But that's not what you're doing now, is it?

PR #318, opened 2018-03-27
Microdata 2

sohanirao:


src/model/page-config-resolver-test.js
```diff
@@ -268,7 +272,337 @@ describes.realWin('PageConfigResolver', {}, env => {
...
+              Paid content possibly. \
+            </div> \
+          </div>';
+      addMicrodata(divElement);
+      readyState = 'complete';
+
+      const resolver = new PageConfigResolver(gd);
+      resolver.resolveConfig().then(config => {
+        expect(config.isLocked()).to.be.true;
+        expect(config.getProductId()).to.equal('pub1:premium');
```
dvoytenko: Since this is within a promise - there's no guarantee that these checks will actually execute within the test. To guarantee it, you can either return this promise or use `yield`.

src/model/page-config-resolver-test.js
```diff
@@ -268,7 +272,337 @@ describes.realWin('PageConfigResolver', {}, env => {
...
+      const validAccessElement = createElement(doc, 'div');
+      validAccessElement.innerHTML =
+          '<div itemprop="hasPart" itemscope itemtype="http://schema.org/WebPageElement"> \
+            <meta itemprop="isAccessibleForFree" content=false/> \
+            <meta itemprop="cssSelector" content=".paywalled-section"/> \
+          </div>';
+      doc.getElementById('top').appendChild(validAccessElement);
+
+      return resolver.resolveConfig().then(config => {
+        expect(config.isLocked()).to.be.true;
```
dvoytenko: Ditto: why is it `true`?

src/model/page-config-resolver-test.js
```diff
@@ -268,7 +272,337 @@ describes.realWin('PageConfigResolver', {}, env => {
...
+              Paid content possibly. \
+            </div> \
+          </div>';
+      addMicrodata(divElement);
+      readyState = 'complete';
+      const resolver = new PageConfigResolver(gd);
+      expect(resolver.check()).to.be.null;
+    });
+
+    it('malformed microdata no access info default to unlocked', () => {
```
dvoytenko: That's not "malformed" - an expect case. Though I thought you tested it plenty above.

src/model/page-config-resolver-test.js
```diff
@@ -268,7 +272,337 @@ describes.realWin('PageConfigResolver', {}, env => {
...
+            </div> \
+            <div itemprop="articleBody" class="paywalled-section"> \
+              Paid content possibly. \
+            </div> \
+          </div>';
+      addMicrodata(divElement);
+      readyState = 'complete';
+      return new PageConfigResolver(gd).resolveConfig().then(config => {
+        expect(config.isLocked()).to.be.true;
+        expect(config.getProductId()).to.equal('pub1:premium');
```
dvoytenko: Hmm. I think the first one should win...

src/model/page-config-resolver-test.js
```diff
@@ -268,7 +272,337 @@ describes.realWin('PageConfigResolver', {}, env => {
...
+      const divElement = createElement(doc, 'div');
+      divElement.innerHTML =
+          '<div itemscope itemtype="http://schema.org/NewsArticle"> \
+            <div itemprop="isPartOf" itemscope itemtype="http://schema.org/CreativeWork http://schema.org/Product"> \
+              <meta itemprop="name" content="New York Times"/> \
+              <meta itemprop="productID" content="pub1:premium"/> \
+            </div> \
+            <div itemscope itemtype="http://schema.org/Section"> \
+              <div itemprop="hasPart" itemscope itemtype="http://schema.org/WebPageElement"> \
+                <meta itemprop="isAccessibleForFree" content="False"/> \
```
dvoytenko: I think I understand the confusion now. The `isAccessibleForFree` is indeed allowed for `WebPageElement` and `NewsArticle`. However, the value within `WebPageElement` is not the right one for us (that's what I meant by "section" when we talked offline). We are only interested in whether the whole document is considered to be `isAccessibleForFree`.

src/model/page-config-resolver.js
```diff
@@ -283,6 +283,49 @@ class JsonLdParser {
   }
 }
 
+/**
+ * Class the describes the microdata found in the page
+ */
+class Microdata {
+  /**
+  * @global @type {Object<string, (string|boolean)>} Property
```
dvoytenko: What does this mean?

src/model/page-config-resolver.js
```diff
@@ -292,6 +335,120 @@ class MicrodataParser {
...
   }
 
+  /**
+   * Returns true if access is restricted, otherwise false
+   * @param {Node} root A node that is an item of type 'NewsArticle'
+   * @return {?boolean} locked access
+   * @private
+   */
+  discoverLocked_(root) {
+    const ALREADY_SEEN = 'alreadySeenForAccessInfo';
```
dvoytenko: There's already `ALREADY_SEEN` constant defined at the top of the file. Can we just use that?

src/model/page-config-resolver.js
```diff
@@ -292,6 +335,120 @@ class MicrodataParser {
...
+    const nodeList = root
+        .querySelectorAll("[itemprop='isAccessibleForFree']");
+    for (let i = 0; nodeList[i]; i++) {
+      const element = nodeList[i];
+      const content = element.content || element.textContent;
+      if (!content) {
+        continue;
+      }
+      if (content.toLowerCase() == 'true') {
+        value = true;
```
dvoytenko: Where's the `value` defined?

src/model/page-config-resolver.js
```diff
@@ -292,6 +335,120 @@ class MicrodataParser {
...
+        return !value;
+      }
+    }
+    return null;
+  }
+
+  /**
+   * Verifies if a node is valid based on the following
+   * - child of an item of type 'NewsArticle'
+   * - not a child of an item of type 'Section'
```
dvoytenko: As described before - this is not correct. It should really be "not a child of any other itemtype, including WebPageElement". Confusion probably happened b/c I mistakenly called `WebPageElement` as a "section".

src/model/page-config-resolver.js
```diff
@@ -292,6 +335,120 @@ class MicrodataParser {
...
+      if (!content) {
+        continue;
+      }
+      if (content.toLowerCase() == 'true') {
+        value = true;
+      } else if (content.toLowerCase() == 'false') {
+        value = false;
+      }
+      if (this.isValidElement_(element, root, ALREADY_SEEN)) {
+        return !value;
```
dvoytenko: This is kind of hard to read: if element is valid, reverse the value? It'd probably be better to just call it `locked` and calculate it accordingly.

src/model/page-config-resolver.js
```diff
@@ -292,6 +335,120 @@ class MicrodataParser {
...
+   * - not a child of an item of type 'Section'
+   * - not seen before, marked using the alreadySeen tag
+   * @param current the node to be verified
+   * @param root the parent to track up to
+   * @param alreadySeen used to tag already visited nodes
+   * @return {!boolean} valid node
+   * @private
+   */
+  isValidElement_(current, root, alreadySeen) {
+    while (current && !current[alreadySeen]) {
```
dvoytenko: I think the only thing that matters is whether or not the "value" element itself has been seen. It's too dangerous to mark the whole chain up - that could cause some incorrect interpretation. As such, you can simply move that check in the `discoverLocked_`.

src/model/page-config-resolver.js
```diff
@@ -292,6 +335,120 @@ class MicrodataParser {
...
+      current[alreadySeen] = true;
+      if (current.hasAttribute('itemscope')) {
+        const type = current.getAttribute('itemtype');
+        if (type == 'http://schema.org/NewsArticle') {
```
dvoytenko: By Schema.org rules itemtypes can always be concatenated. So, we need to test that `NewsArticle` is part of the value, but not necessarily the only value.

src/model/page-config-resolver.js
```diff
@@ -292,6 +335,120 @@ class MicrodataParser {
...
+          // This parent may need to be check for other nodes
+          current[alreadySeen] = false;
+          return true;
+        } else if (type.indexOf('Section') >= 0) {
```
dvoytenko: See above.

src/model/page-config-resolver.js
```diff
@@ -292,6 +335,120 @@ class MicrodataParser {
...
+   * - not seen before, marked using the alreadySeen tag
+   * @param current the node to be verified
+   * @param root the parent to track up to
+   * @param alreadySeen used to tag already visited nodes
+   * @return {!boolean} valid node
+   * @private
+   */
+  isValidElement_(current, root, alreadySeen) {
+    while (current && !current[alreadySeen]) {
+      current[alreadySeen] = true;
```
dvoytenko: Can you switch to `for` loop? Way safer in the long run.

src/model/page-config-resolver.js
```diff
@@ -292,6 +335,120 @@ class MicrodataParser {
...
+   * Obtains the product ID that meets the requirements
+   * - child of an item of type 'NewsArticle'
+   * - Not a child of an item of type 'Section'
+   * - child of an item of type 'productID'
+   * @param {Node} root A node that is an item of type 'NewsArticle'
+   * @return {?string} product ID, if found
+   * @private
+   */
+  discoverProductId_(root) {
+    const ALREADY_SEEN = 'alreadySeenForProductInfo';
```
dvoytenko: Ditto: one constant should be enough.

src/model/page-config-resolver.js
```diff
@@ -292,6 +335,120 @@ class MicrodataParser {
...
+    const nodeList = root
+        .querySelectorAll("[itemprop='productID']");
```
dvoytenko: Reverse single/double quotes.

src/model/page-config-resolver.js
```diff
@@ -292,6 +335,120 @@ class MicrodataParser {
...
+    for (let i = 0; nodeList[i]; i++) {
+      const element = nodeList[i];
+      const content = element.content || element.textContent;
```
dvoytenko: To confirm: this could only be a `<meta>` element? B/c other elements would not have a `.content` property and you'd have to read the attribute directly instead.

src/model/page-config-resolver.js
```diff
@@ -292,6 +335,120 @@ class MicrodataParser {
...
+      const item = element.closest('[itemtype][itemscope]');
+      const type = item.getAttribute('itemtype');
+      if (type.indexOf('Product') <= -1) {
```
dvoytenko: Please use the full itemtype as URL for this check.

src/model/page-config-resolver.js
```diff
@@ -292,6 +335,120 @@ class MicrodataParser {
...
+   */
+  tryExtractConfig_() {
+    let itemId = 0;
+    const results = [];
+    const nodeList = this.doc_.getRootNode().querySelectorAll(
+        "[itemscope][itemtype='http://schema.org/NewsArticle']");
+    const domReady = this.doc_.isReady();
+    for (let i = 0; nodeList[i]; i++) {
+      const element = nodeList[i];
+      if (!domReady && !hasNextNodeInDocumentOrder(element)) {
```
dvoytenko: Hmm. That'd likely block us for the remainder of parsing. Why can't we search within partial DOM?

src/model/page-config-resolver.js
```diff
@@ -292,6 +335,120 @@ class MicrodataParser {
...
+        continue;
+      }
+      const locked = this.discoverLocked_(element);
+      const productInfo = this.discoverProductId_(element);
+      const props = {};
```
dvoytenko: If both are found, why not just return the config right away? What value do we get from storing a set of them? If we want to keep the track of `locked` or `productId` in case one of them already found - we can just store them as private vars in this class. E.g.
```
/** @private {?boolean} */
this.locked_ = null;
```

Then we can avoid discovering it many times, i.e.
```
if (this.locked_ === null) {
  this.locked_ = discoverLocked(...);
}
```

sohanirao: Done.

src/model/page-config-resolver.js
sohanirao: Done.

src/model/page-config-resolver.js
sohanirao: Done

src/model/page-config-resolver.js
sohanirao: As discussed, we are keeping two different tags.

src/model/page-config-resolver.js
sohanirao: Done.

src/model/page-config-resolver.js
sohanirao: Done

src/model/page-config-resolver.js
sohanirao: It wasn't. addressed.

src/model/page-config-resolver.js
sohanirao: This is no longer necessary

src/model/page-config-resolver-test.js
sohanirao: The first one happens to be under "Section"

src/model/page-config-resolver-test.js
sohanirao: will remove this case.

src/model/page-config-resolver-test.js
```diff
@@ -268,7 +272,337 @@ describes.realWin('PageConfigResolver', {}, env => {
...
+          </div>';
+      addMicrodata(divElement);
+      readyState = 'complete';
+      return new PageConfigResolver(gd).resolveConfig().then(config => {
+        expect(config.isLocked()).to.be.true;
+        expect(config.getProductId()).to.equal('pub1:premium');
+      });
+    });
+
+    it('should parse unlocked access', () => {
```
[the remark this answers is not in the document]
sohanirao: i dont intend this case to use default value. the content of the meta actually indicates free access.

src/model/page-config-resolver-test.js
```diff
@@ -268,7 +272,337 @@ describes.realWin('PageConfigResolver', {}, env => {
...
+
+    beforeEach(() => {
+      readyState = 'loading';
+      Object.defineProperty(doc, 'readyState', {get: () => readyState});
+    });
+
+    it('should parse locked access', () => {
+      const divElement = createElement(doc, 'div');
+      divElement.innerHTML =
+          '<div itemscope itemtype="http://schema.org/NewsArticle"> \
```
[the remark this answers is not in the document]
sohanirao: yes, it is necessary when we split across lines. But reverse quotes looks and is cleaner. will change.

src/model/page-config-resolver-test.js
```diff
@@ -268,7 +272,337 @@ describes.realWin('PageConfigResolver', {}, env => {
...
+
+    it('should wait for element ready', () => {
+      const resolver = new PageConfigResolver(gd);
+      const element = createElement(doc, 'div');
+      addMicrodata(element);
+
+      // Empty content.
+      let config = resolver.check();
+      readyState = 'complete';
+      expect(config).to.be.null;
```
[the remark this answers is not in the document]
sohanirao: Done

src/model/page-config-resolver.js
sohanirao: I derived these requirements from the tests in the JSON LD parser. Regardless of the parser, I assumed that those test cases imply similar requirements on Microdata parser. In this case if dom is not ready, it ensure if there's atleast one more node to process before proceeding? Is that not expected?

src/model/page-config-resolver.js
sohanirao: as per schema.org itemprop can be a part of span, anchor tags and meta tags.

src/model/page-config-resolver-test.js
```diff
@@ -268,7 +272,337 @@ describes.realWin('PageConfigResolver', {}, env => {
...
+    it('should parse unlocked access', () => {
+      const divElement = createElement(doc, 'div');
+      divElement.innerHTML =
+          '<div itemscope itemtype="http://schema.org/NewsArticle"> \
+            <div itemprop="isPartOf" itemscope itemtype="http://schema.org/CreativeWork http://schema.org/Product"> \
+              <meta itemprop="name" content="New York Times"/> \
+              <meta itemprop="productID" content="pub1:premium"/> \
+            </div> \
+            <div itemprop="hasPart" itemscope itemtype="http://schema.org/WebPageElement"> \
+              <meta itemprop="isAccessibleForFree" content="true"/> \
```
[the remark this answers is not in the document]
sohanirao: some of your comments are based on the assumption that I used WebPageElement to invalidate a meta property. This wasn't true. I have removed WebpageElement unless it is intended to discard a node. Please revisit the new patch.

src/model/page-config-resolver-test.js
```diff
@@ -268,7 +272,337 @@ describes.realWin('PageConfigResolver', {}, env => {
...
+
+      // Add content.
+      const divElement = createElement(doc, 'div');
+      divElement.innerHTML =
+          '<div itemscope itemtype="http://schema.org/NewsArticle"> \
+            <div itemprop="isPartOf" itemscope itemtype="http://schema.org/CreativeWork http://schema.org/Product"> \
+            <meta itemprop="name" content="New York Times"/> \
+            <meta itemprop="productID" content="pub1:premium"/> \
+          </div> \
+          <div itemprop="hasPart" itemscope itemtype="http://schema.org/WebPageElement"> \
```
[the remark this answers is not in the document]
sohanirao: do you believe line 335 is enough to complete this test?

src/model/page-config-resolver-test.js
```diff
@@ -268,7 +272,337 @@ describes.realWin('PageConfigResolver', {}, env => {
...
+            <div itemprop="articleBody" class="paywalled-section"> \
+              Paid content possibly. \
+            </div> \
+          </div>';
+      addMicrodata(divElement);
+      readyState = 'complete';
+
+      const resolver = new PageConfigResolver(gd);
+      resolver.resolveConfig().then(config => {
+        expect(config.isLocked()).to.be.true;
```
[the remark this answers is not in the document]
sohanirao: fixed.

src/model/page-config-resolver-test.js
sohanirao: my previous patch WebPageElement was a valid parent.

src/model/page-config-resolver-test.js
sohanirao: this has been addressed overall.

src/model/page-config-resolver.js
sohanirao: offline discussion, we'll keep alreadySeen tag depending on what we are searching for.

src/model/page-config-resolver.js
sohanirao: addressed.

src/model/page-config-resolver.js
sohanirao: discussed this offline.

src/model/page-config-resolver.js
sohanirao: Done

src/model/page-config-resolver.js
dvoytenko: Yeah, this is a very different case. The issue with LD parse is that it parses JSON text, which must be complete to be parseable, otherwise it completely fails. So, I used `hasNextNodeInDocumentOrder` as an indication that JSON text is fully available. We don't need this for microdata. As a matter of fact, I believe, this is one of its benefits. And you don't need `domReady` here either.

src/model/page-config-resolver.js
dvoytenko: In this case, you'll have to drop element.content and replace it with element.getAttribute('content'). Span/anchor elements do not expose content property - so this will always be undefined.

src/model/page-config-resolver-test.js
```diff
@@ -268,7 +272,350 @@ describes.realWin('PageConfigResolver', {}, env => {
...
+            <div>
+              <meta itemprop="isAccessibleForFree" content="true"/>
+              <meta itemprop="cssSelector" content=".paywalled-section"/>
+            </div>
+            <div itemprop="articleBody" class="paywalled-section">
+              Paid content possibly.
+            </div>
+          </div>`;
+      addMicrodata(divElement);
+      readyState = 'complete';
```
dvoytenko: I'd generally reduce `readyState = 'complete'` use to only few tests where it really matters.

src/model/page-config-resolver-test.js
```diff
@@ -268,7 +272,350 @@ describes.realWin('PageConfigResolver', {}, env => {
...
+          </div> \
+          <div itemprop="articleBody" class="paywalled-section"> \
+            Paid content possibly. \
+          </div> \
+        </div>';
+      addMicrodata(divElement);
+      readyState = 'complete';
+      config = resolver.check();
+      expect(config).to.be.ok;
+      expect(config.getProductId()).to.equal('pub1:premium');
```
dvoytenko: Check the `isLocked()` here as well.

src/model/page-config-resolver-test.js
```diff
@@ -268,7 +272,350 @@ describes.realWin('PageConfigResolver', {}, env => {
...
+      return expect(resolver.resolveConfig()).to.eventually.equal(config);
+    });
+
+    /*
```
dvoytenko: Uncomment or remove.

src/model/page-config-resolver-test.js
dvoytenko: Understood.

src/model/page-config-resolver.js
```diff
@@ -283,13 +283,180 @@ class JsonLdParser {
   }
 }
 
+/**
+ * Class the describes the microdata found in the page
+ */
+class Microdata {
```
dvoytenko: Do we still really need this class? It looks like we'd operate totally great w/o it?

src/model/page-config-resolver.js
```diff
@@ -283,13 +283,180 @@ class JsonLdParser {
...
   /**
    * @param {!Doc} doc
    */
   constructor(doc) {
     /** @private @const {!Doc} */
     this.doc_ = doc;
+    /** @private {?boolean} */
+    this.access_ = null;
+    /** @private {?string} */
+    this.productID_ = null;
```
dvoytenko: `productId_`. I know that Schema.org has a different styleguide, but JS styleguide in JS code wins.

src/model/page-config-resolver.js
```diff
@@ -283,13 +283,180 @@ class JsonLdParser {
...
+   * - child of an item of type 'NewsArticle'
+   * - not a child of an item of type 'Section'
+   * - not seen before, marked using the alreadySeen tag
+   * @param current the node to be verified
+   * @param root the parent to track up to
+   * @param alreadySeen used to tag already visited nodes
+   * @return {!boolean} valid node
+   * @private
+   */
+  isValidElement_(current, root, alreadySeen) {
```
dvoytenko: IMHO, let's rename `isValidContext_`

src/model/page-config-resolver.js
```diff
@@ -283,13 +283,180 @@ class JsonLdParser {
...
+    for (let node = current;
+        node && !node[alreadySeen]; node = node.parentNode) {
+      node[alreadySeen] = true;
+      if (node.hasAttribute('itemscope')) {
+        const type = node.getAttribute('itemtype');
```
dvoytenko: It could be null. Do `const type = node.getAttribute(...) || ''`. Or change `if` above to include `if (node.hasAttribute('itemscope') || node.hasAttribute('itemtype'))`

src/model/page-config-resolver.js
```diff
@@ -283,13 +283,180 @@ class JsonLdParser {
...
+    }
+    return false;
+  }
+
+  /**
+   * Obtains the product ID that meets the requirements
+   * - child of an item of type 'NewsArticle'
+   * - Not a child of an item of type 'Section'
+   * - child of an item of type 'productID'
+   * @param {Node} root A node that is an item of type 'NewsArticle'
```
dvoytenko: `!Node`

src/model/page-config-resolver.js
```diff
@@ -283,13 +283,180 @@ class JsonLdParser {
...
     this.doc_ = doc;
+    /** @private {?boolean} */
+    this.access_ = null;
+    /** @private {?string} */
+    this.productID_ = null;
+  }
+
+  /**
+   * Returns false if access is restricted, otherwise true
+   * @param {Node} root A node that is an item of type 'NewsArticle'
```
dvoytenko: `!Node`

src/model/page-config-resolver.js
```diff
@@ -283,13 +283,180 @@ class JsonLdParser {
...
+   * @return {?boolean} locked access
+   * @private
+   */
+  discoverAccess_(root) {
+    const ALREADY_SEEN = 'alreadySeenForAccessInfo';
+    const nodeList = root
+        .querySelectorAll("[itemprop='isAccessibleForFree']");
+    for (let i = 0; nodeList[i]; i++) {
+      const element = nodeList[i];
+      const content = element.content || element.textContent;
```
dvoytenko: Replace `element.content` with `element.getAttribute('content')`.

src/model/page-config-resolver.js
```diff
@@ -283,13 +283,180 @@ class JsonLdParser {
...
+   * @return {?string} product ID, if found
+   * @private
+   */
+  discoverProductId_(root) {
+    const ALREADY_SEEN = 'alreadySeenForProductInfo';
+    const nodeList = root
+        .querySelectorAll('[itemprop="productID"]');
+    for (let i = 0; nodeList[i]; i++) {
+      const element = nodeList[i];
+      const content = element.content || element.textContent;
```
dvoytenko: Replace `element.content` with `element.getAttribute('content')`.

src/model/page-config-resolver.js
```diff
@@ -283,13 +283,180 @@ class JsonLdParser {
...
+        if (discoveredProductInfo) {
+          this.productID_ = discoveredProductInfo;
+          props['productID'] = discoveredProductInfo;
+        }
+      } else {
+        props['productID'] = this.productID_;
+      }
+      results.push({id: ++itemId, properties: props});
+    }
+    return new Microdata(results);
```
dvoytenko: So, really lost why we still need this? Shouldn't it just be:

```
if (this.access_ !== null && this.productId_ !== null) {
  return new PageConfig(...);
}
```
?

src/model/page-config-resolver.js
```diff
@@ -283,13 +283,180 @@ class JsonLdParser {
...
+        "[itemscope][itemtype='http://schema.org/NewsArticle']");
+    const domReady = this.doc_.isReady();
+    for (let i = 0; nodeList[i]; i++) {
+      const element = nodeList[i];
+      if (!domReady && !hasNextNodeInDocumentOrder(element)) {
+        continue;
+      }
+      const props = {};
+      if (this.access_ == null) {
+        const discoveredAccess = this.discoverAccess_(element);
```
dvoytenko: Once nuance is that `access`, even if discovered to be a `null` should be defaulted to `true` (unlocked) when `domReady`, since we know we won't find it ever.

src/model/page-config-resolver-test.js
```diff
@@ -268,7 +272,350 @@ describes.realWin('PageConfigResolver', {}, env => {
...
+            <div itemprop="isPartOf" itemscope itemtype="http://schema.org/CreativeWork http://schema.org/Product"> \
+              <meta itemprop="name" content="New York Times"/> \
+              <meta itemprop="productID" content="pub1:premium"/> \
+            </div> \
+            <div itemprop="articleBody" class="paywalled-section"> \
+              Paid content possibly. \
+            </div> \
+          </div>';
+      addMicrodata(divElement);
+      readyState = 'complete';
```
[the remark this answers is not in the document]
sohanirao: added a test like this.

src/model/page-config-resolver-test.js
sohanirao: Done

src/model/page-config-resolver-test.js
sohanirao: Done.

src/model/page-config-resolver.js
sohanirao: Done

src/model/page-config-resolver.js
sohanirao: Done

src/model/page-config-resolver.js
sohanirao: Done

src/model/page-config-resolver-test.js
sohanirao: Sorry which line?

src/model/page-config-resolver.js
sohanirao: Understood.

src/model/page-config-resolver.js
sohanirao: Updated.

PR #327, opened 2018-04-03
Started and canceled callbacks for all flows.

dvoytenko:


src/runtime/link-accounts-flow.js
```diff
@@ -116,6 +117,7 @@ export class LinkbackFlow {
    * @return {!Promise}
    */
   start() {
+    this.deps_.callbacks().triggerFlowStarted('linkAccount');
```
dparikh: Would be nice if we have an enum for all the flow names.

PR #328, opened 2018-04-04
Remove media query styles to be set as important

dparikh:
This is to support MS edge browser, since important styles would not be updated.


src/utils/style.js
```diff
@@ -144,7 +144,6 @@ export const defaultStyles = {
   'visibility': 'visible',
   'white-space': 'normal',
   'widows': '2',
-  'width': 'auto',
```
dvoytenko: Add a test that `width`, `left` and `margin-left` never get into this array.

dparikh: Done

PR #335, opened 2018-04-10
Remove transform for Firefox, Safari

dparikh:


src/ui/ui.css
```diff
@@ -56,7 +56,6 @@ swg-loading {
   position: absolute !important;
   top: 40% !important;
   left: 45% !important;
-  transform: translate(-50%, -50%) !important;
```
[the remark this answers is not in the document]
dvoytenko: But it'd seem the shift/centering is due to `top:40%; left: 45%` above? That seems weird to center with percent. Does it really work correctly?

src/ui/ui.css
```diff
@@ -56,7 +58,6 @@ swg-loading {
   position: absolute !important;
   top: 40% !important;
```
dvoytenko: Why then keep `top`/`left`? I don't think the content will ever be in the middle this way...

src/ui/ui.css
```diff
@@ -45,18 +47,13 @@ swg-loading-container {
 }
 
 swg-container,
```
dvoytenko: Could you move all these `display:block` to the very top?

src/ui/ui.css
```diff
@@ -20,8 +20,10 @@ body {
 }
 
 swg-loading-container {
-  position: absolute !important;
   width: 100% !important;
+  display: flex !important;
+  align-items: center !important;
+  justify-content: center !important;
   height: 148px !important;
```
dvoytenko: Check if this can be made `100%`, coupled with `min-height`.

dparikh: Done

src/ui/ui.css
dparikh: Done. Moved after body {...}

PR #336, opened 2018-04-12
Allow optional close action for Offers View

dparikh:
This change has corresponding CL.

### DIFF
--- a/src/api/subscriptions.js
+++ b/src/api/subscriptions.js
@@ -64,19 +64,19 @@ export class Subscriptions {
 
   /**
    * Starts the Offers flow.
-   * @param {!OptionsRequest=} opt_options
+   * @param {!OffersRequest=} opt_options
    */
   showOffers(opt_options) {}
 
   /**
    * Show subscription option.
-   * @param {!OptionsRequest=} opt_options
+   * @param {!OffersRequest=} opt_options
    */
   showSubscribeOption(opt_options) {}
 
   /**
    * Show abbreviated offers.
-   * @param {!OptionsRequest=} opt_options
+   * @param {!OffersRequest=} opt_options
    */
   showAbbrvOffer(opt_options) {}
 
@@ -167,6 +167,7 @@ export const SubscriptionFlows = {
  *   order is preserved.
  * - list - a predefined list of SKUs. Use of this property is uncommon.
  *   Possible values are "default" and "amp". Default is "default".
+ * - isClosable - a boolean value to determine whether the view is closable.
  *
  * @typedef {{
  *   skus: (!Array<string>|undefined),
@@ -174,7 +175,7 @@ export const SubscriptionFlows = {
  *   isClosable: (boolean|undefined),
  * }}
  */
-export let OptionsRequest;
+export let OffersRequest;
 
 
 /**
--- a/src/api/subscriptions.js
+++ b/src/api/subscriptions.js
@@ -158,7 +158,7 @@ export const SubscriptionFlows = {
   SHOW_ABBRV_OFFER: 'showAbbrvOffer',
   SUBSCRIBE: 'subscribe',
   LINK_ACCOUNT: 'linkAccount',
-}
+};
 
 
 /**
@@ -171,6 +171,7 @@ export const SubscriptionFlows = {
  * @typedef {{
  *   skus: (!Array<string>|undefined),
  *   list: (string|undefined),
+ *   isClosable: (boolean|undefined),
  * }}
  */
 export let OptionsRequest;
@@ -189,4 +190,3 @@ export let LoginRequest;
  * }}
  */
 export let SaveSubscriptionRequest;
-
--- a/src/runtime/offers-flow-test.js
+++ b/src/runtime/offers-flow-test.js
@@ -123,7 +123,7 @@ describes.realWin('OffersFlow', {}, env => {
           showNative: false,
           list: 'other',
           skus: null,
-          isClosable: false,
+          isClosable: true,
         })
         .returns(Promise.resolve(port));
     return offersFlow.start();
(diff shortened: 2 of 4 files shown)


src/api/subscriptions.js
```diff
@@ -171,6 +171,7 @@ export const SubscriptionFlows = {
  * @typedef {{
  *   skus: (!Array<string>|undefined),
  *   list: (string|undefined),
+ *   isClosable: (boolean|undefined),
```
dvoytenko: The `isClosable` field only applies to one of three APIs: `showOffers`. Do you think it's confusing to keep it in this structure? Generally, I like this idea, but still wondering... At the very least we should document that it only applies to `showOffers`. Or alternatively, we could extend `isClosable` to all offer-based APIs. WDYT?

src/runtime/offers-flow-test.js
```diff
@@ -155,6 +159,7 @@ describes.realWin('OffersFlow', {}, env => {
           showNative: true,
           list: 'default',
           skus: null,
+          isClosable: undefined,
```
dvoytenko: I'd rather default explicitly in the over-the-boundary requests. This way the default only rests in one place.

src/runtime/offers-flow.js
```diff
@@ -54,6 +54,7 @@ export class OffersFlow {
           'showNative': deps.callbacks().hasSubscribeRequestCallback(),
           'list': options && options.list || 'default',
           'skus': options && options.skus || null,
+          'isClosable': options && !!options.isClosable,
```
dvoytenko: What do you think the default actually should be?

src/runtime/offers-flow.js
```diff
@@ -238,7 +239,10 @@ export class AbbrvOfferFlow {
     // If result is due to requesting offers, redirect to offers flow
     this.activityIframeView_.acceptResult().then(result => {
       if (result.data['viewOffers']) {
-        new OffersFlow(this.deps_, this.options_).start();
+        const isClosable = result.data && result.data['isClosable'] || false;
+        const options = this.options_ || {};
+        Object.assign(options, {'isClosable': isClosable});
```
dvoytenko: This is not a string property - it's a field. I'm afraid you'll have to programmatically construct options.

src/runtime/offers-flow.js
```diff
@@ -238,7 +239,10 @@ export class AbbrvOfferFlow {
     // If result is due to requesting offers, redirect to offers flow
     this.activityIframeView_.acceptResult().then(result => {
       if (result.data['viewOffers']) {
-        new OffersFlow(this.deps_, this.options_).start();
+        const isClosable = result.data && result.data['isClosable'] || false;
```
dvoytenko: Not sure it's a good idea to pass `isClosable` from the iframe. What does the iframe know that we don't already know?

src/api/subscriptions.js
```diff
@@ -171,6 +171,7 @@ export const SubscriptionFlows = {
...
  * }}
  */
 export let OptionsRequest;
```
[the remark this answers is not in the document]
dparikh: Done

src/api/subscriptions.js
dparikh: I think, keeping `isClosable` for all offersView would be better since, in future if required this feature would be available to make `AbbrvOfferView` and `OptionsView` non-closable.

src/runtime/offers-flow.js
dparikh: Removed this and using the local var instead.

src/runtime/offers-flow.js
dparikh: Done

src/runtime/offers-flow.js
```diff
@@ -120,6 +126,9 @@ export class SubscribeOptionFlow {
     /** @private @const {!../components/dialog-manager.DialogManager} */
     this.dialogManager_ = deps.dialogManager();
 
+    /** @private @const {boolean} */
+    this.isOffersViewClosable_ = false;
```
dvoytenko: Hmm. Not seeing it set anywhere.

src/runtime/offers-flow.js
```diff
@@ -165,7 +175,9 @@ export class SubscribeOptionFlow {
    */
   maybeOpenOffersFlow_(data) {
     if (data && data['subscribe']) {
-      new OffersFlow(this.deps_, this.options_).start();
+      const options = this.options_ || {};
+      options.isClosable = this.isOffersViewClosable_;
```
dvoytenko: Are you sure this is needed? Wouldn't the right value already be in the `this.options_`?

src/runtime/offers-flow.js
```diff
@@ -198,6 +210,9 @@ export class AbbrvOfferFlow {
     /** @private @const {!../components/dialog-manager.DialogManager} */
     this.dialogManager_ = deps.dialogManager();
 
+    /** @private @const {boolean} */
+    this.isOffersViewClosable_ = false;
```
dvoytenko: Ditto

src/runtime/offers-flow.js
dparikh: Used in line #179 as a default value if not available in options.

src/runtime/offers-flow.js
dparikh: Added a check and setting the default value only if `isClosable` is not available in options.

src/runtime/offers-flow.js
dparikh: Same as above.

src/runtime/offers-flow.js
dvoytenko: Hmm. I still don't understand why the private var is needed? Maybe just define a top-level `const DEFAULT_CLOSABLE = false`?

dparikh: Kept private var here so that individual views (AbbrvView, SubscribeOptionsView) can have different values in future (though currently both are same).

Moved to `const` for now.

PR #341, opened 2018-04-18
Return a promise of status or status of link save request

sohanirao:


src/api/subscriptions.js
```diff
@@ -145,7 +145,7 @@ export class Subscriptions {
  /**
   * Starts the save subscriptions flow.
   * @param {!SaveSubscriptionRequest} request
-  * @return {!Promise}
+  * @return {!Promise<!boolean>} status or promise of status of request
```
dvoytenko: No need for `!` in `!boolean` - it's implied.

src/runtime/link-accounts-flow.js
```diff
@@ -194,41 +194,66 @@ export class LinkSaveFlow {
...
+    /** @private @const {!./deps.DepsDef} */
+    this.deps_ = deps;
+
     /** @private @const {!web-activities/activity-ports.ActivityPorts} */
     this.activityPorts_ = deps.activities();
 
     /** @private @const {!../components/dialog-manager.DialogManager} */
     this.dialogManager_ = deps.dialogManager();
 
+    /** {!../api/subscriptions.SaveSubscriptionRequest} */
```
dvoytenko: `@type`. But why do we need a default request? Just for testing? If so, add `TODO` so we don't forget to remove.

src/runtime/link-accounts-flow.js
```diff
@@ -194,41 +194,66 @@ export class LinkSaveFlow {
...
+    /** {!Promise<boolean>} */
+    const returnPromise = new Promise(resolve => {
+      const promise = this.activityIframeView_.port().then(port => {
+        return acceptPortResult(
+            port,
+            feOrigin(),
+            /* requireOriginVerified */ true,
+            /* requireSecureChannel */ true);
+      });
+      promise.then(result => {
```
dvoytenko: You don't need `new Promise` construct here. Just simple return this `promise` here. And move the `catch` and the final `then` close to the top level. It could be all nicely inlined on the `openView` promise below.

src/runtime/link-accounts-flow.js
```diff
@@ -194,41 +194,66 @@ export class LinkSaveFlow {
...
+        return result['linked'];
+      }).catch(() => {
+        return false;
+      }).then(result => {
+        // The flow is complete.
+        this.dialogManager_.completeView(this.activityIframeView_);
+        this.completed_ = true;
```
dvoytenko: What does the `complete_` boolean do?

src/runtime/link-accounts-flow.js
```diff
@@ -194,41 +194,66 @@ export class LinkSaveFlow {
...
+      feUrl('/linksaveiframe'),
+      feArgs({
+        'publicationId': this.deps_.pageConfig().getPublicationId(),
+        'token': this.saveSubscriptionRequest_['token'],
+        'isClosable': true,
+      }),
+      /* shouldFadeBody */ false
+    );
+    /** {!Promise<boolean>} */
+    const returnPromise = new Promise(resolve => {
```
dvoytenko: This change needs a test that ensures no matter what activity returns we always get the promise resolved with the right boolean value.

src/api/subscriptions.js
sohanirao: Done

src/runtime/link-accounts-flow.js
sohanirao: Done

src/runtime/link-accounts-flow.js
sohanirao: Done

src/runtime/link-accounts-flow.js
sohanirao: Done.

src/runtime/link-accounts-flow.js
sohanirao: It was meant as a placeholder till I actually implemented view completion through resolving the result.

PR #346, opened 2018-04-23
Boilerplate and fake data for "deferred account completion" API

dvoytenko:


docs/deferred-account-flow.md
```diff
@@ -0,0 +1,65 @@
...
+
+```
+function onEntitlement(entitlements) {
+  if (!entitlements.enablesThis()) {
+    return;
+  }
+  // The entitlement enables this product.
+  // 1. Check if the publisher site knows this subscription.
+  if (!myKnownSubscription(entitlements)) {
+    // 2. Start deferred account creation.
```
alaskajim: I think we need something here to indicate that they have no identity in particular. That is, they may have received this subscription in the pubsub, but they have no associated user account. Maybe this?...

if (!hasAssociatedUserAccount(entitlements))

dvoytenko: Done

PR #348, opened 2018-04-25
Update saveSubscriptionRequest to include authCode or token

sohanirao:


src/api/subscriptions.js
```diff
@@ -207,7 +207,7 @@ export let LoginRequest;
 
 /**
  * @typedef {{
- *   token: string,
+ *   request: {(token|authCode): string}
```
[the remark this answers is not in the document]
sohanirao: Done.

src/runtime/link-accounts-flow-test.js
```diff
@@ -308,21 +307,38 @@ describes.realWin('LinkSaveFlow', {}, env => {
...
-          token: 'test',
+          request: {token: 'test'},
+          isClosable: true,
+        })
+        .returns(Promise.resolve(port));
+    return linkSaveFlow.start();
+  });
+
+  it('should have valid LinkSaveFlow constructed to use authCode', () => {
+    linkSaveFlow = new LinkSaveFlow(runtime, {request: {authCode: 'testCode'}});
```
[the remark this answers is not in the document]
sohanirao: Done. However, right now the test fails with Error indicating same reason that the expectation verifies.

src/runtime/link-accounts-flow.js
```diff
@@ -205,7 +205,7 @@ export class LinkSaveFlow {
 
     /** TODO(sohanirao): Default request only for test */
     /** @type {!../api/subscriptions.SaveSubscriptionRequest} */
-    const defaultRequest = {token: 'test'};
+    const defaultRequest = {request: {token: 'test'}};
```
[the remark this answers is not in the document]
sohanirao: Done

src/runtime/link-accounts-flow.js
```diff
@@ -225,7 +225,7 @@ export class LinkSaveFlow {
       feUrl('/linksaveiframe'),
       feArgs({
         'publicationId': this.deps_.pageConfig().getPublicationId(),
-        'token': this.saveSubscriptionRequest_['token'],
+        'request': this.saveSubscriptionRequest_['request'],
```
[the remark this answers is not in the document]
sohanirao: Done

src/runtime/link-accounts-flow-test.js
```diff
@@ -322,7 +322,44 @@ describes.realWin('LinkSaveFlow', {}, env => {
...
+  });
+
+  it('should throw error both token and authCode', () => {
+    linkSaveFlow = new LinkSaveFlow(
+      runtime, {token: 'token', authCode: 'testCode'});
+    return linkSaveFlow.start().then(() => {
+      throw new Error('must have failed');
+    }, reason => {
+      expect(() => {
+        throw reason;}).to.throw(/Both authCode and token are available/);
```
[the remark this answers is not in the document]
dvoytenko: This is because the start method itself fails, not the promise produced by this method. In other words, this is a synchronous error. To test it you can do:

```
expect(() => {
  linkSaveFlow.start();
}).to.throw(/.../);
```

src/api/subscriptions.js
```diff
@@ -206,9 +206,10 @@ export let OffersRequest;
 export let LoginRequest;
 
 /**
- * @typedef {{
- *   token: string,
- * }}
+ * Properties:
+ * - one and only one of "token" or "authCode" and
+ * - corresponding value
+ * @const {Object<string, string>}
```
dvoytenko: This should continue to be typedef. It should look like this:

```
@typedef {{
  token: (string|undefined),
  authCode: (string|undefined),
}}
```

src/api/subscriptions.js
```diff
@@ -206,9 +206,10 @@ export let OffersRequest;
 export let LoginRequest;
 
 /**
- * @typedef {{
- *   token: string,
- * }}
+ * Properties:
+ * - one and only one of "token" or "authCode" and
+ * - corresponding value
```
dvoytenko: What does "corresponding value" mean?

src/runtime/link-accounts-flow.js
```diff
@@ -219,15 +215,43 @@ export class LinkSaveFlow {
    * @return {!Promise}
    */
   start() {
+    const iframeArgs = {
+      'publicationId': this.deps_.pageConfig().getPublicationId(),
+      'isClosable': true,
+    };
+    /** string */
+    const token = (() => {
```
dvoytenko: Hmm. This is slightly confusing to me. Why is this block closurized and resolved by function invocation? Also, at this time the `request_` is defined as a non-null value, why check for its existence below? Finally, the request is a typedef, so you should access it's properties directly w/o quotes. 

In other words, isn't this just:
```
const token = this.request_.token;
```

src/api/subscriptions.js
sohanirao: I mean to say the value corresponding to the key , token or authCode. Will remove if it is confusing.

src/runtime/link-accounts-flow.js
sohanirao: Done.

src/api/subscriptions.js
sohanirao: can't have both @const {Object<string, string>} and @typedef right? I will keep the typedef because it's more accurate and comment that only one of they entries can exist.

src/runtime/link-accounts-flow-test.js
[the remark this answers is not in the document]
sohanirao: Done

src/runtime/link-accounts-flow-test.js
```diff
@@ -322,7 +322,44 @@ describes.realWin('LinkSaveFlow', {}, env => {
...
+  it('should throw error no token or authCode', () => {
+    linkSaveFlow = new LinkSaveFlow(runtime);
+    return linkSaveFlow.start().then(() => {
+      throw new Error('must have failed');
+    }, reason => {
+      expect(() => {
+        throw reason;}).to.throw(/Neither token or authCode is available/);
+    });
+  });
+
```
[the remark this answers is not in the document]
sohanirao: Done

src/api/subscriptions.js
dvoytenko: No. But it's definitely not a `@const` and `Object<string, string>` is not a great fit here.

src/api/subscriptions.js
```diff
@@ -221,8 +221,11 @@ export let LoginRequest;
 
 
 /**
+ * Properties:
+ * - one and only one of "token" or "authCode"
```
dvoytenko: Let's document these two links here: for authCode: https://developers.google.com/actions/identity/oauth2-code-flow
and for tokenhttps://developers.google.com/actions/identity/oauth2-implicit-flow

sohanirao: Done.

PR #351, opened 2018-05-04
Support opening a dialog as hidden till ready to show

sohanirao:


src/components/dialog-manager.js
```diff
@@ -58,28 +58,32 @@ export class DialogManager {
   }
 
   /**
+   * @param {boolean=} animated
+   * @param {boolean=} hidden
    * @return {!Promise<!Dialog>}
    */
-  openDialog() {
+  openDialog(animated = true, hidden = false) {
```
dvoytenko: I think it's worth touching base on this a bit: but wouldn't `animated` and `hidden` be mutually exclusive? Did you consider have a separate API, e.g. `openHiddenDialog()` and `openHiddenView()`? Any thoughts on that?

src/components/dialog-test.js
```diff
@@ -99,6 +100,19 @@ describes.realWin('Dialog', {}, env => {
       expect(graypaneStubs.show).to.not.be.called;
     });
 
+    it('should open dialog as hidden', function* () {
+      immediate();
+      dialog.open(ANIMATE, HIDDEN);
+      yield dialog.animating_;
```
dvoytenko: Shouldn't we ensure/test to confirm that it definitely is not animating when hidden?

src/components/dialog-test.js
```diff
@@ -112,6 +126,19 @@ describes.realWin('Dialog', {}, env => {
       expect(graypaneStubs.show).to.be.calledOnce.calledWith(ANIMATE);
     });
 
+    it('should build the view and show hidden iframe', function* () {
+      const openedDialog = yield dialog.open(NO_ANIMATE, HIDDEN);
+      yield openedDialog.openView(view);
+      expect(computedStyle(win, element)['opacity']).to.equal('1');
```
dvoytenko: Let's test that dialog's element is also now visible?

src/components/dialog.js
```diff
@@ -261,17 +264,46 @@ export class Dialog {
...
+
     return view.init(this).then(() => {
       setImportantStyles(view.getElement(), {
         'opacity': 1,
       });
+      if (this.hidden_ && view.shouldFadeBody()) {
+        this.graypane_.show(/* animated */ true);
+      }
+      this.show_();
+      this.hidden_ = false;
```
dvoytenko: `this.hidden_ = false`, imho, should be set inside the `show_()` method.

src/components/dialog.js
```diff
@@ -261,17 +264,46 @@ export class Dialog {
...
       this.setLoading(false);
     });
   }
 
+  /**
+   * Show the iframe
+   * @private
+   * @param {boolean=} showNow
+   */
+  show_(showNow = this.hidden_) {
```
dvoytenko: What does `showNow` really mean? It feels to me that by extension, this method should only have `animated` boolean param.

src/components/dialog.js
```diff
@@ -261,17 +264,46 @@ export class Dialog {
...
+      this.show_();
+      this.hidden_ = false;
       this.setLoading(false);
     });
   }
 
+  /**
+   * Show the iframe
+   * @private
+   * @param {boolean=} showNow
```
dvoytenko: I'm not a big fun of private methods with optional params. There's an extra cost that's not always warranted with only private invocation.

src/components/dialog.js
```diff
@@ -261,17 +264,46 @@ export class Dialog {
...
     }
+
     return view.init(this).then(() => {
       setImportantStyles(view.getElement(), {
         'opacity': 1,
       });
+      if (this.hidden_ && view.shouldFadeBody()) {
+        this.graypane_.show(/* animated */ true);
+      }
+      this.show_();
```
dvoytenko: IMHO, it'd be better to just do something like this here:

```
if (this.hidden_) {
  if (view.shouldFadeBody()) {
   this.graypane_.show(...);
  }
  this.show_();
}
```

Otherwise, in the current state, you split "hidden" conditionals between this part of code and `show_` method.

src/runtime/link-accounts-flow.js
```diff
@@ -240,22 +240,23 @@ export class LinkSaveFlow {
...
-      }).catch(() => {
-        return false;
-      }).then(result => {
-        // The flow is complete.
-        this.dialogManager_.completeView(this.activityIframeView_);
-        return result;
-      });
-    });
+    return this.dialogManager_.openView(this.activityIframeView_,
+        /* animated */ true, /* hidden */ true).then(() => {
```
dvoytenko: This is important enough to test that the dialog was opened as hidden. Not sure how easy it'd be in the current test structure, but this is part of the feature. We should be able to set up spy on `openView` method, I'd think.

src/components/dialog-test.js
sohanirao: start

sohanirao: Done

src/components/dialog.js
sohanirao: done

src/components/dialog.js
sohanirao: done

src/components/dialog.js
sohanirao: done.

src/components/dialog.js
sohanirao: addressed.

src/runtime/link-accounts-flow.js
sohanirao: done

src/components/dialog-manager.js
sohanirao: talked offline. removed animated.

src/components/dialog-test.js
sohanirao: need to address.

src/runtime/link-accounts-flow-test.js
```diff
@@ -337,6 +341,15 @@ describes.realWin('LinkSaveFlow', {}, env => {
     return linkSaveFlow.start();
   });
 
+  it('should open dialog in hidden mode', () => {
+    linkSaveFlow = new LinkSaveFlow(runtime, {token: 'test'});
+    const dialogPromise = Promise.resolve(new Dialog(new GlobalDoc(win)));
+    dialogManagerMock.expects('openDialog')
+        .withExactArgs(/* hidden */true).returns(dialogPromise);
+    const openPromise = Promise.resolve(true);
+    expect(linkSaveFlow.start()).to.eventually.equal(openPromise);
```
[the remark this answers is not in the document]
sohanirao: Done

PR #353, opened 2018-05-10
Removed "not launched" in deferred account creation flow docs

jenniferchui:
Removed "not launched" in deferred account creation flow docs.
@ratana

### DIFF
--- a/docs/deferred-account-flow.md
+++ b/docs/deferred-account-flow.md
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 -->
 
-# Deferred account creation flow (not launched)
+# Deferred account creation flow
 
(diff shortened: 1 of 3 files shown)


docs/deferred-account-flow.md
```diff
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 -->
 
-# Deferred account creation flow (not launched)
+# Deferred account creation flow
```
dvoytenko: Looks great, but could you please also update the related clauses in:

- https://github.com/subscriptions-project/swg-js/blob/master/docs/flows.md
- https://github.com/subscriptions-project/swg-js/blob/master/src/api/subscriptions.js#L107

jenniferchui: Sure.

PR #360, opened 2018-06-16
Add additional URLs to prefetch and preconnect.

cramforce:
As of today this should help a lot with getting things up and running quicker.
The change does add a maintainability burden, so it should be considered whether this is worth it, or if a more sustainable strategy is feasible.


dvoytenko:
CI runs. It just runs very fast :)

PR #363, opened 2018-06-26
Toast css fixes

dparikh:
- Move the JS based styles to media queries
- Remove unused styles
- Ensure that when position is fixed/absolute, have left/right & top/bottom attributes


src/ui/toast.js
```diff
@@ -22,12 +22,8 @@ import {
 
 /** @const {!Object<string, string|number>} */
 export const toastImportantStyles = {
-  'position': 'fixed',
-  'bottom': 0,
   'height': 0,
   'max-height': '46px',
```
dvoytenko: Why not put `max-height` in CSS as well? Are you updating it dynamically as well?

dparikh: Removed from here since already included in CSS

PR #364, opened 2018-07-02
Waiting API

chenshay:


src/api/subscriptions.js
```diff
@@ -134,6 +134,13 @@ export class Subscriptions {
    */
   setOnLinkComplete(callback) {}
 
+  /**
+   * TODO(chenshay): Figure out what type the promise returns.
+   * @param {Promise} accountPromise Publisher's promise to lookup account.
```
dvoytenko: `!Promise`.

src/api/subscriptions.js
```diff
@@ -134,6 +134,13 @@ export class Subscriptions {
    */
   setOnLinkComplete(callback) {}
 
+  /**
+   * TODO(chenshay): Figure out what type the promise returns.
```
dvoytenko: If we return any result in the promise, I think it should simply be the passthrough value from the `accountPromise` for convenience. That way API clients won't need to overnest.

src/api/subscriptions.js
```diff
@@ -208,6 +215,12 @@ export const SubscriptionFlows = {
 };
 
 
+/** @enum {string} */
+export const WaitingViews = {
```
dvoytenko: What is this enum for?

src/runtime/waiting-api.js
```diff
@@ -0,0 +1,97 @@
...
+
+    /** @private @const {!ActivityIframeView} */
+    this.activityIframeView_ = new ActivityIframeView(
+        this.win_,
+        this.activityPorts_,
+        feUrl('/loginwaitingiframe'),
+        feArgs({
+          publicationId: deps.pageConfig().getPublicationId(),
+          productId: deps.pageConfig().getProductId(),
+          accountPromise: this.accountPromise_,
```
dvoytenko: We can't pass a promise here. Actually cannot pass any non-structured-clonable values.

src/runtime/waiting-api.js
```diff
@@ -0,0 +1,97 @@
...
+    this.openViewPromise_ = this.dialogManager_.openView(
+        this.activityIframeView_);
+
+    return this.accountPromise_.then(account => {
+      // Account was found.
+      this.dialogManager_.completeView(this.activityIframeView_);
+      return Promise.resolve(account);
+    }, () => {
+      this.dialogManager_.completeView(this.activityIframeView_);
+      return Promise.reject({
```
dvoytenko: Are we sure we want to automatically start this flow ourselves? There are a lot of configurations options there. I think it'd be best to allow a client code to call `deferredAccountCreation` if it wants to.

src/runtime/waiting-api.js
```diff
@@ -0,0 +1,97 @@
...
+    this.deps_.callbacks().triggerFlowStarted(
+        WaitingViews.LOGIN_WAITING_VIEW);
+
+    this.openViewPromise_ = this.dialogManager_.openView(
+        this.activityIframeView_);
+
+    return this.accountPromise_.then(account => {
+      // Account was found.
+      this.dialogManager_.completeView(this.activityIframeView_);
+      return Promise.resolve(account);
```
dvoytenko: `return account`?

src/api/subscriptions.js
chenshay: done

src/api/subscriptions.js
chenshay: done

src/api/subscriptions.js
chenshay: no longer needed

src/runtime/waiting-api.js
chenshay: done

src/runtime/waiting-api.js
chenshay: done

src/runtime/waiting-api.js
chenshay: done

src/api/subscriptions.js
```diff
@@ -208,6 +214,10 @@ export const SubscriptionFlows = {
 };
 
 
+/** @type {string} */
+export const LOGIN_WAITING_VIEW = 'loginWaitingView';
```
dvoytenko: I'd say we should either put it inside the "SubscriptionFlows" above or remove it completely. I don't think it's a useful analytical signal b/c publishers would always call this function themselves - i.e. they always aware when it's used.

src/runtime/runtime-test.js
```diff
@@ -678,6 +677,15 @@ describes.realWin('Runtime', {}, env => {
       expect(stub).to.be.calledOnce.calledWithExactly(options, callback);
     });
 
+    it('should delegate "showWaitingIndicator"', () => {
+      configuredRuntimeMock.expects('showWaitingIndicator').once()
+          .returns(Promise.resolve());
```
dvoytenko: Like with the previous issue with return values: let's always test that the value returned here is the exact value returned by the configuredRuntime. An easy way is to return some random value here in the mock, e.g. `11` and then test that the result of `showWaitingIndicator.then(value => {...})` below is equal to the same value (`11`).

src/runtime/waiting-api.js
```diff
@@ -0,0 +1,82 @@
...
+    this.openViewPromise_ = this.dialogManager_.openView(
+        this.activityIframeView_);
+
+    return this.accountPromise_.then(account => {
+      // Account was found.
+      this.dialogManager_.completeView(this.activityIframeView_);
+      return account;
+    }, () => {
+      this.dialogManager_.completeView(this.activityIframeView_);
+      return Promise.reject('no account found');
```
dvoytenko: Let's rethrow the reason here - it'll be more obvious to the API user.

chenshay: done

src/runtime/runtime-test.js
chenshay: done

src/api/subscriptions.js
chenshay: done

PR #369, opened 2018-07-12
Toast animation, use transition api

dparikh:
- Use `transition` api for Toast animation.
- Animate Toast in 400ms.
- On close, animate toast again before removing the element from DOM

### DIFF
--- a/src/ui/toast.js
+++ b/src/ui/toast.js
@@ -33,11 +33,14 @@ export const toastImportantStyles = {
  */
 export let ToastSpecDef;
 
+/** @const {string} */
+const swgToastClass = 'swg-toast';
+
 /** @const {!Object<string, string>} */
 const iframeAttributes = {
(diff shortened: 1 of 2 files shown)


src/ui/toast.js
```diff
@@ -106,21 +110,58 @@ export class Toast {
...
+    this.animate_(() => {
+      transition(this.iframe_, {
+        'transform': 'translateY(100%)',
+        'opacity': 1,
+        'visiblity': 'visible',
+      }, 400, 'ease-out');
+    });
+
+    // Remove the toast from the DOM after animation is complete.
+    this.doc_.getWin().setTimeout(() => {
```
dvoytenko: I'd recommend returning a Promise here for ease of testing and/or nesting in code.

dparikh: Done

PR #372, opened 2018-07-16
FIXIT summit - Fix lint warnings

dparikh:
Issue #373

### DIFF
--- a/src/api/user-data.js
+++ b/src/api/user-data.js
@@ -27,7 +27,7 @@ export class UserData {
     /** @const {string} */
     this.idToken = idToken;
     /** @private @const {!Object} */
-    this.data = data;
+    this.data_ = data;
 
     /** @const {string} */
     this.id = data['sub'];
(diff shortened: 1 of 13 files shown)


src/api/user-data.js
```diff
@@ -27,7 +27,7 @@ export class UserData {
     /** @const {string} */
     this.idToken = idToken;
     /** @private @const {!Object} */
-    this.data = data;
+    this.data_ = data;
```
dvoytenko: Please keep the name and make it public instead.

src/utils/jwt.js
```diff
@@ -33,7 +33,9 @@ let JwtTokenInternalDef;
  * Provides helper methods to decode and verify JWT tokens.
  */
 export class JwtHelper {
-
+  /**
```
dvoytenko: Remove this one. Really not a big fun of generic comments.

src/api/user-data.js
dparikh: Done

src/utils/jwt.js
dparikh: Done

PR #376, opened 2018-07-17
Change build script to 1)not minify code 2)remove comments 3)export correctly

chenshay:


kristoferbaxter:
cc @dvoytenko 

This change prompted a few questions:
1. Why do modules in the project use `import` syntax and `module.exports` instead of `exports`?
2. ESLint rules regarding exports seem to be circumvented via `module.exports`, should they be removed or is there a preferred pattern here that could be documented.

**Please note**: This branch is not intended to be merged, but to see what integration with the amp closure compiler rollup plugin would look like in the future.

dvoytenko:
@kristoferbaxter this was my agreement with @erwinmombay initially to import the code into AMP. Would be super glad to follow any other model, especially if we have our own rollup plugin.

PR #380, opened 2018-07-26
Ensure save subscription call returns the result of corresponding runtime API call

sohanirao:


src/runtime/runtime.js
```diff
@@ -349,7 +349,7 @@ export class Runtime {
   saveSubscription(saveSubscriptionRequestCallback) {
     return this.configured_(true)
         .then(runtime => {
-          runtime.saveSubscription(saveSubscriptionRequestCallback);
+          return runtime.saveSubscription(saveSubscriptionRequestCallback);
```
dvoytenko: There should be a test for this.

sohanirao: Dima can you clarify? there are tests that verify delegation to configured run time to call the right flow. I just updated the tests. can you please take a look and let me know if you still think some tests are missing?

sohanirao: Done

PR #383, opened 2018-08-03
login prompt documentation

chenshay:


docs/login-flow.md
```diff
@@ -15,10 +15,55 @@ limitations under the License.
...
 This is the flow in which Google will ask the user for permission to log them in to the publisher's site. It is initiated by the publisher when the publisher doesn't find a user's subscription but Google does find the subscription.
 
-To enable this flow, 
+Here is an example of what this flow can look like:
 
-// TODO(chenshay): finish this part.
\ No newline at end of file
+```
+// Publisher is looking up the user.
+const  accountPromise = new Promise( … );
```
sohanirao: it is unclear what account you are looking for? as per PRD it says that you need to find know if publisher entitlements are false but google entitlements is true, then depending on the consent either call loginNotification or showLoginPrompt. If no user is found, create an account.  What does an account stand for here? the user? or their entitlements? what precursor data structure is required for the accountPromise to be fulfilled?

PR #393, opened 2018-08-20
Loading indicator is only shown if previous view didn't show it

sohanirao:


src/components/dialog.js
```diff
@@ -230,15 +233,45 @@ export class Dialog {
...
+      removeChildren(this.getContainer());
+    }
+    setImportantStyles(view.getElement(), resetViewStyles);
+    if (noloadingIndicatorPresent) {
+      // When loading indicator was not displayed in the previous view,
+      // loading indicator must be displayed while transitioning to new view.      
       this.loadingView_.show();
-    } else {
+      // Keep track of whether or not a view shows loading indicator
+      view.setLoadingIndicator(true);
```
dvoytenko: I'm not understanding this boolean. Why should the unrelated view track whether the previous view is showing loading indicator?

src/components/dialog.js
```diff
@@ -230,15 +233,45 @@ export class Dialog {
...
+      // Keep track of whether or not a view shows loading indicator
+      view.setLoadingIndicator(true);   
+    }
+  }
+
+   /**
+   * Transition out of an old view
+   * @param {boolean} hideLoadingView 
+   */
+  exitTransitionFromOldView(hideLoadingView) {
```
dvoytenko: private?

src/components/dialog.js
```diff
@@ -230,15 +233,45 @@ export class Dialog {
...
+    if (hideLoadingView) {
+      //When a loading indicator is present, hide it.
```
dvoytenko: nit: space

src/components/dialog.js
```diff
@@ -230,15 +233,45 @@ export class Dialog {
...
+      view.setLoadingIndicator(true);   
+    }
+  }
+
+   /**
+   * Transition out of an old view
+   * @param {boolean} hideLoadingView 
+   */
+  exitTransitionFromOldView(hideLoadingView) {
+    if (hideLoadingView) {
```
dvoytenko: Is `if` helpful? Can we just always hide it?

src/components/dialog.js
```diff
@@ -230,15 +233,45 @@ export class Dialog {
...
+      //When a loading indicator is present, hide it.
       this.loadingView_.hide();
     }
+    // If previous view displayed a loading indicator, this view
+    // did not, hence remove previous element instead.
+    if (this.previousView_ && this.previousView_.hasLoadingIndicator()) {
```
dvoytenko: Ditto, if `this.previousView_` exists and attached, should we just always remove it? It doesn't matter whether `hasLoadinfgIndicator` is true, right?

E.g.

if (this.previousView_ && this.previousView_.getElement().parentNode) {
  removeElement(this.previousView_.getElement())
}

src/components/dialog.js
```diff
@@ -230,15 +233,45 @@ export class Dialog {
...
+      this.getContainer().removeChild(this.previousView_.getElement());
```
dvoytenko: Use `removeElement` from dom.js

src/components/dialog.js
```diff
@@ -230,15 +233,45 @@ export class Dialog {
...
-   * Whether to display loading indicator.
-   * @param {boolean} isLoading
+   * Transitions to the next view
+   * @param {!./view.View} view
    */
-  setLoading(isLoading) {
-    if (isLoading) {
+  entryTransitionToNextView(view) {
+    // Temporarily cache the old view
+    this.previousView_ = this.view_;
```
dvoytenko: As a small variation, you could store `this.previousView_` _only_ when it has a progress indicator. You could even call it `this.prevProgressView_` to make it clear. That way some conditions further down would be a bit simpler.

src/ui/activity-iframe-view.js
```diff
@@ -105,6 +110,21 @@ export class ActivityIframeView extends View {
...
+   * @return {boolean}
+   */
+  hasLoadingIndicator() {
+    return this.hasLoadingIndicator_;
+  }
+   /**
+   * Sets the loading indicator for this view
+   * @param {boolean} display
+   */
+  setLoadingIndicator(display) {
```
dvoytenko: So, not at all clear why we need a setter for this. Also, this var is declared above as a `@const`. Could you please explain?

src/components/dialog.js
sohanirao: This is not an unrelated view, this is the current view. Each view keeps track of whether or not it showed the loading indicator.

src/components/dialog.js
```diff
@@ -230,15 +233,45 @@ export class Dialog {
...
+    var noloadingIndicatorPresent = !this.previousView_
+        || (this.previousView_
```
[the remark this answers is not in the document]
sohanirao: ack

src/components/dialog.js
```diff
@@ -230,15 +233,45 @@ export class Dialog {
...
 
   /**
-   * Whether to display loading indicator.
-   * @param {boolean} isLoading
+   * Transitions to the next view
+   * @param {!./view.View} view
    */
-  setLoading(isLoading) {
-    if (isLoading) {
+  entryTransitionToNextView(view) {
```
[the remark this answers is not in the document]
sohanirao: no

src/components/dialog.js
```diff
@@ -230,15 +233,45 @@ export class Dialog {
...
+    // Temporarily cache the old view
+    this.previousView_ = this.view_;
+    var noloadingIndicatorPresent = !this.previousView_
```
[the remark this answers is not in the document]
sohanirao: done

src/components/dialog.js
sohanirao: yes

src/components/dialog.js
sohanirao: yes

src/components/dialog.js
sohanirao: done

src/ui/activity-iframe-view.js
sohanirao: That was a copy paste error. @const is not necessary.

How else can we cache whether or not a view showed the loading indicator? The previousProgressView_ gets nulled out at the completion of view init.

src/components/dialog.js
dvoytenko: I think this should be tracked by the dialog class. But I still not seeing why it helps.

src/components/dialog.js
```diff
@@ -230,17 +234,45 @@ export class Dialog {
...
+      view.setLoadingIndicator(true);   
     }
   }
 
+   /**
+   * Transition out of an old view
+   * @private
+   */
+  exitTransitionFromOldView_() {
+    //When a loading indicator is present, hide it.
```
dvoytenko: nit: space

src/ui/activity-iframe-view.js
dvoytenko: Let's discuss.

src/components/dialog.js
```diff
@@ -230,12 +234,33 @@ export class Dialog {
...
+    }
+  }
+
+   /**
+   * Transition out of an old view
+   * @private
+   */
+  exitTransitionFromOldView_() {
+    //When a loading indicator is present, hide it.
+    this.loadingView_.hide();
```
dvoytenko: Loading indicator is hidden twice: let's leave only one. FWIW, always trying to hide both: loading indicator and previous view seemed like an easier to read option to me.

src/ui/activity-iframe-view.js
```diff
@@ -36,13 +36,15 @@ export class ActivityIframeView extends View {
    * @param {string} src
    * @param {!Object<string, ?>=} args
    * @param {boolean=} shouldFadeBody
+   * @param {boolean=} hasLoadingIndicator
```
dvoytenko: Let's also include the file where you set it to true.

sohanirao: done

sohanirao: done

PR #397, opened 2018-08-30
Loading indicator only shown if previous view didn't show it

sohanirao:


src/components/dialog.js
```diff
@@ -230,12 +234,31 @@ export class Dialog {
   }
 
   /**
-   * Whether to display loading indicator.
-   * @param {boolean} isLoading
+   * Transitions to the next view
```
dvoytenko: nit: all sentences should end with periods. Really saves time on review.

PR #401, opened 2018-09-11
Accept id_token and entitlements from the payment response

dvoytenko:
Partial for #400.

### DIFF
--- a/src/runtime/api-test.js
+++ b/src/runtime/api-test.js
@@ -244,15 +244,15 @@ describe('PurchaseData', () => {
 
 
 describes.sandboxed('SubscribeResponse', {}, () => {
-  let sr, pd, ud, ent, complete, promise;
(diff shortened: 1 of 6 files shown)


src/runtime/api-test.js
```diff
@@ -244,20 +244,35 @@ describe('PurchaseData', () => {
 
 
 describes.sandboxed('SubscribeResponse', {}, () => {
-  let sr, pd, ud, complete, promise;
+  let sr, pd, ud, ent, complete, promise;
```
chenshay: all these variables have very unclear names to people like me who know nothing about this. What is sr, pd, ud, ent?

chenshay: ah now I see. I'd rename these to have the complete name

dvoytenko: Done

PR #402, opened 2018-09-14
Initial commit of uuid utils

dparikh:
Added initial code for generating uuid.


third_party/random_uuid/Random.uuid.js
```diff
@@ -0,0 +1,42 @@
+/** @license
```
dvoytenko: Not quite. Include the original file exactly as is and then add a new file the way you want. So 3 files total. You can call the new file, e.g. "uuid_swg.js" or such to make it clear it's adopted for our needs.

Also, you need `README.md` (so total 4 files) where you reference where the library comes from (URL) and what modifications you made.

dparikh: Done.

PR #411, opened 2018-09-19
Transactionid creation POC - NOT FOR SUBMIT

dparikh:


src/components/dialog.js
```diff
@@ -100,6 +102,11 @@ export class Dialog {
     this.iframe_ = new FriendlyIframe(
         doc.getWin().document, {'class': 'swg-dialog'});
 
+    this.storage_ = new Storage(doc.getWin());
```
dvoytenko: `dialog.js` seems to be a wrong place for runtime-level transaction ID?

PR #417, opened 2018-09-24
Logging Service API

sohanirao:


src/runtime/logging-service-api.js
```diff
@@ -0,0 +1,72 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {ActivityIframeView} from '../ui/activity-iframe-view';
+import {feArgs, feUrl} from './services';
+
+
+export class LoggingServiceApi {
```
dvoytenko: Can we rename to `AnalyticsService`? No need for `Api` suffix.

src/runtime/logging-service-api.js
```diff
@@ -0,0 +1,72 @@
...
+    this.win_ = deps.win();
+
+    /** @private @const {!web-activities/activity-ports.ActivityPorts} */
+    this.activityPorts_ = deps.activities();
+
+    /** @private @const {!../components/dialog-manager.DialogManager} */
+    this.dialogManager_ = deps.dialogManager();
+
+    /** @private @const {!ActivityIframeView} */
+    this.activityIframeView_ = new ActivityIframeView(
```
dvoytenko: This will not work this way - the dialog manager will kill this iframe at an unpredictable time. It's an iframe-based activity, but it's not a "view".

src/runtime/logging-service-api.js
```diff
@@ -0,0 +1,72 @@
...
+        this.win_,
+        this.activityPorts_,
+        feUrl('/serviceiframe'),
+        feArgs({
+          publicationId: deps.pageConfig().getPublicationId(),
+          productId: deps.pageConfig().getProductId(),
+          // No user consent necessary. This is a hidden iframe.
+          userConsent: false,
```
dvoytenko: remove

src/runtime/logging-service-api.js
```diff
@@ -0,0 +1,72 @@
...
+        }),
+        /* shouldFadeBody */ false
+    );
+  }
+
+  /**
+   * @return {!Promise}
+   */
+  start() {
+    this.activityIframeView_.acceptResult().then(() => {
```
dvoytenko: remove

src/runtime/analytics-service.js
```diff
@@ -0,0 +1,201 @@
...
+ */
+
+import {feArgs, feUrl} from './services';
+import {createElement} from '../utils/dom';
+import {isCancelError} from '../utils/errors';
+import {FriendlyIframe} from '../components/friendly-iframe';
+
+/** @const {!Object<string, string>} */
+const iframeAttributes = {
+  'visibility': 'hidden',
```
dvoytenko: `display: none` is also ideal.

src/runtime/analytics-service.js
```diff
@@ -0,0 +1,201 @@
...
+ * limitations under the License.
+ */
+
+import {feArgs, feUrl} from './services';
+import {createElement} from '../utils/dom';
+import {isCancelError} from '../utils/errors';
+import {FriendlyIframe} from '../components/friendly-iframe';
+
+/** @const {!Object<string, string>} */
+const iframeAttributes = {
```
dvoytenko: These are styles, not attributes. Right?

src/runtime/analytics-service.js
```diff
@@ -0,0 +1,201 @@
...
+    /** @private @const {!Doc} */
+    this.doc_ = doc;
+
+    /** @private @const {!web-activities/activity-ports.ActivityPorts} */
+    this.activityPorts_ = activityPorts;
+
+    /** @private @const {!HTMLIFrameElement} */
+    this.iframe_ =
+        /** @type {!HTMLIFrameElement} */ (createElement(
+            this.doc_.getWin().document, 'iframe', iframeAttributes));
```
dvoytenko: Let's clarify attributes vs styles.

src/runtime/analytics-service.js
```diff
@@ -0,0 +1,201 @@
...
+
+    /** @private @const {string} */
+    this.src_ = feUrl('/serviceiframe');
+
+    this.args_ = feArgs({
+      publicationId: args['publicationId'],
```
dvoytenko: Why not just pass pageConfig here and use the `publicationId` from there?

src/runtime/analytics-service.js
```diff
@@ -0,0 +1,201 @@
...
+     */
+    this.portPromise_ = new Promise(resolve => {
+      this.portResolver_ = resolve;
+    });
+
+    /**
+     * @private @constant
+     * {!FriendlyIframe}
+     */
+    this.friendlyIframe_ = new FriendlyIframe(this.doc_.getWin().document,
```
dvoytenko: What is this for?

src/runtime/analytics-service.js
```diff
@@ -0,0 +1,201 @@
...
+  getElement() {
+    return this.iframe_;
+  }
+
+  /**
+   * Build the iframe with the styling after iframe is loaded.
+   * @private
+   */
+  buildIframe_() {
+    const iframeBody = this.friendlyIframe_.getBody();
```
dvoytenko: None of these should be needed: we don't need the container since this iframe is never displayed. We should just simply add our iframe to the document's body.

src/runtime/analytics-service.js
```diff
@@ -0,0 +1,201 @@
...
+    return container;
+  }
+
+  getIframe() {
+    return this.friendlyIframe_;
+  }
+
+  open_() {
+    // Attach.
+    this.doc_.getBody().appendChild(this.friendlyIframe_.getElement());  // Fires onload.
```
dvoytenko: Ditto: we should just add our iframe here and no need for friendly iframe otherwise.

src/runtime/analytics-service.js
```diff
@@ -0,0 +1,201 @@
...
+  }
+
+  /**
+   * @param {!web-activities/activity-ports.ActivityIframePort} port
+   * @return {!Promise}
+   */
+  onOpenIframeResponse_(port) {
+    this.port_ = port;
+    this.portResolver_(port);
+    return this.port_.whenReady();
```
dvoytenko: We don't need to wait for port readiness: it's specifically a UI signal. It basically means "ready to be displayed". Probably a slightly unfortunate name. But otherwise - as soon as we are connected - we are all good to go.

src/runtime/analytics-service.js
```diff
@@ -0,0 +1,201 @@
...
+  }
+
+  /**
+   */
+  close() {
+    this.doc_.getBody().removeChild(this.iframe_.getElement());
```
dvoytenko: It doesn't seem like `this.iframe_` has a `getElement()` method.

src/runtime/analytics-service.js
```diff
@@ -0,0 +1,201 @@
...
+  }
+
+  /**
+   * @return {!Promise}
+   */
+  start() {
+    /* TODO(sohanirao): Determine if completion must be handled and how.
```
dvoytenko: I don't think we need to handle completion.

src/runtime/analytics-service.js
```diff
@@ -0,0 +1,201 @@
...
+    this.port().then(port => {
+      port.onMessage(callback);
+    });
+  }
+
+  /**
+   * Accepts results from the caller.
+   * @return {!Promise<!web-activities/activity-ports.ActivityResult>}
+   */
+  acceptResult() {
```
dvoytenko: Similarly: no completion -> no result.

src/runtime/analytics-service.js
```diff
@@ -0,0 +1,201 @@
...
+    return this.port().then(port => port.acceptResult());
+  }
+
+  /**
+   * @param {function()} callback
+   */
+  onCancel(callback) {
```
dvoytenko: Ditto: this iframe will not be cancelable.

src/runtime/analytics-service.js
```diff
@@ -0,0 +1,121 @@
...
+    /** @private @const {!Doc} */
+    this.doc_ = doc;
+
+    /** @private @const {!web-activities/activity-ports.ActivityPorts} */
+    this.activityPorts_ = activityPorts;
+
+    /** @private @const {!HTMLIFrameElement} */
+    this.iframe_ =
+        /** @type {!HTMLIFrameElement} */ (createElement(
+            this.doc_.getWin().document, 'iframe', iframeStyles));
```
dvoytenko: I double-checked the `createElement` function and the 3rd argument there is element attributes, _not_ styles. So, I think this is incorrect. Instead, probably we need this:

```
this.iframe_ = createElement(...document, 'iframe');
setImportantStyles(iframeStyles);
```

src/runtime/analytics-service.js
```diff
@@ -0,0 +1,121 @@
...
+
+    /**
+     * @private @const
+     * {!Promise<!web-activities/activity-ports.ActivityIframePort>}
+     */
+    this.portPromise_ = new Promise(resolve => {
+      this.portResolver_ = resolve;
+    });
+
+    this.container_ = null;
```
dvoytenko: Remove?

src/runtime/analytics-service.js
```diff
@@ -0,0 +1,121 @@
...
+  /**
+   */
+  close() {
+    this.doc_.getBody().removeChild(this.getElement());
+  }
+
+  /**
+   * @return {!Promise<!web-activities/activity-ports.ActivityIframePort>}
+   */
+  port() {
```
dvoytenko: I guess the `port()`, `message`, and `onMessage` will become `@private` eventually?

src/runtime/analytics-service.js
```diff
@@ -0,0 +1,127 @@
...
+    return this.portPromise_;
+  }
+
+  /**
+   * TODO(sohanirao): Replace with AnalyticsEvent
+   * @param {!Object} data
+   */
+  logEvent(event) {
+    this.port_().then(port => {
+      return port.whenReady().then(() => {
```
dvoytenko: Same thing here: no need to wait for `whenReady()` - as soon as you're connected - you should be able to use it. However, feel free to use ready signal if you see a compelling reason. My point simply that it's not necessary if all you need is a non-displayed host.

PR #423, opened 2018-10-04
Implementing IsReadyToPay Caching

chenshay:


src/runtime/entitlements-manager.js
```diff
@@ -76,6 +77,7 @@ export class EntitlementsManager {
         this.positiveRetries_, opt_expectPositive ? 3 : 0);
     if (opt_expectPositive) {
       this.storage_.remove(ENTS_STORAGE_KEY);
+      this.storage_.remove(IS_READY_TO_PAY_STORAGE_KEY);
```
dvoytenko: Please also clear it in the `clear()` method. (You may need to sync with master).

src/runtime/entitlements-manager.js
```diff
@@ -121,23 +123,25 @@ export class EntitlementsManager {
...
-            raw, /* requireNonExpired */ true);
-        if (cached && cached.enablesThis()) {
-          // Already have a positive response.
-          this.positiveRetries_ = 0;
-          return cached;
+      return this.storage_.get(IS_READY_TO_PAY_STORAGE_KEY).then(irtp => {
+        // Try cache first.
+        if (raw) {
+          const cached = this.getValidJwtEntitlements_(
+              raw, /* requireNonExpired */ true, !!irtp);
```
dvoytenko: Will `!!` actually give us what we want? What types can storage save/return? Aren't they all converted to strings?

src/runtime/entitlements-manager.js
```diff
@@ -192,7 +196,11 @@ export class EntitlementsManager {
    * @return {!Entitlements}
    */
   parseEntitlements(json) {
-    const isReadyToPay = json['isReadyToPay'];
+    const isReadyToPay = /** @type {boolean} */ (json['isReadyToPay']);
+    if (json['isReadyToPay']) {
```
dvoytenko: `if (isReadyToPay) {}`. But probably need to always set? Otherwise if RTP goes from true to false, you will never overwrite false? Or we could test that its `!= null`?

src/runtime/entitlements-manager.js
```diff
@@ -121,23 +123,25 @@ export class EntitlementsManager {
...
     return this.storage_.get(ENTS_STORAGE_KEY).then(raw => {
-      // Try cache first.
-      if (raw) {
-        const cached = this.getValidJwtEntitlements_(
-            raw, /* requireNonExpired */ true);
-        if (cached && cached.enablesThis()) {
-          // Already have a positive response.
-          this.positiveRetries_ = 0;
-          return cached;
+      return this.storage_.get(IS_READY_TO_PAY_STORAGE_KEY).then(irtp => {
```
dvoytenko: Can we use `Promise.all` for this?

src/runtime/entitlements-manager.js
chenshay: done

src/runtime/entitlements-manager.js
chenshay: fixed

src/runtime/entitlements-manager.js
chenshay: done

src/runtime/entitlements-manager.js
chenshay: fixed

src/runtime/entitlements-manager.js
```diff
@@ -131,11 +134,17 @@ export class EntitlementsManager {
...
+      this.storage_.get(IS_READY_TO_PAY_STORAGE_KEY),
+    ]).then(cachedValues => {
+      const raw = cachedValues[0];
+      const irtp = cachedValues[1];
       // Try cache first.
       if (raw) {
         const cached = this.getValidJwtEntitlements_(
-            raw, /* requireNonExpired */ true);
+            raw, /* requireNonExpired */ true,
+            /** @type {boolean|undefined} */ (irtp));
```
dvoytenko: I don't believe this is correct. Doesn't storage only operate with strings?

src/runtime/entitlements-manager.js
```diff
@@ -203,7 +212,9 @@ export class EntitlementsManager {
    * @return {!Entitlements}
    */
   parseEntitlements(json) {
-    const isReadyToPay = json['isReadyToPay'];
+    this.storage_.set(IS_READY_TO_PAY_STORAGE_KEY, json['isReadyToPay']);
+    const isReadyToPay =
```
dvoytenko: Why not just reverse these two lines and do `this.storage_.set(IS_READY_TO_PAY_STORAGE_KEY, isReadyToPay);`.

But also two issues here:
 1. I believe storage only accepts strings.
 2. That means that null/undefined will be stored as "null"/"undefined" which is not what you want. You can always remove from storage on null/undefined.

src/runtime/entitlements-manager-test.js
```diff
@@ -88,7 +88,7 @@ describes.realWin('EntitlementsManager', {}, env => {
     };
     return {
       'signedEntitlements': enc(header) + '.' + enc(payload) + '.SIG',
-      'isReadyToPay': !!opt_isReadyToPay,
+      'isReadyToPay': String(opt_isReadyToPay),
```
dvoytenko: Does this emulate a JSON response? B/c in JSON we actually do use booleans.

src/runtime/entitlements-manager-test.js
```diff
@@ -131,6 +131,10 @@ describes.realWin('EntitlementsManager', {}, env => {
           .withExactArgs('toast')
           .returns(Promise.resolve(null))
           .atLeast(0);
+      storageMock.expects('get')
+          .withExactArgs('isreadytopay')
+          .returns(Promise.resolve('undefined'))
```
dvoytenko: Why string here? I'd think `undefined` as a string is a mistake resulting from incorrect write. Otherwise, the storage would return `null` for a missing value.

src/runtime/entitlements-manager.js
```diff
@@ -203,7 +212,12 @@ export class EntitlementsManager {
    * @return {!Entitlements}
    */
   parseEntitlements(json) {
-    const isReadyToPay = json['isReadyToPay'];
+    const isReadyToPay = this.irtpStringToBoolean_(json['isReadyToPay']);
```
dvoytenko: JSON arrives with the actual boolean - that's what servers return.

src/runtime/entitlements-manager.js
```diff
@@ -203,7 +212,12 @@ export class EntitlementsManager {
...
+    if (isReadyToPay == undefined) {
```
dvoytenko: This should be `x == null` - comparisons against `null` are more canonical for this.

src/runtime/entitlements-manager.js
```diff
@@ -343,4 +357,22 @@ export class EntitlementsManager {
...
   }
+
+  /**
+   * Convert String value of isReadyToPay
+   * (from JSON or Cache) to a boolean value.
+   * @param {string} value
+   * @return {boolean|undefined}
+   * @private
+   */
+  irtpStringToBoolean_(value) {
```
dvoytenko: Can we make it a non-exported module function instead?

PR #426, opened 2018-10-07
Provide a clear API to clear the state and cache

dvoytenko:
/cc @emestud

### DIFF
--- a/src/runtime/entitlements-manager.js
+++ b/src/runtime/entitlements-manager.js
@@ -80,6 +80,17 @@ export class EntitlementsManager {
   }
 
   /**
+   * Clears all of the entitlements state and cache.
+   */
+  clear() {
+    this.responsePromise_ = null;
+    this.positiveRetries_ = 0;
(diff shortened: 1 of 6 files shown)


src/runtime/entitlements-manager.js
```diff
@@ -79,6 +79,17 @@ export class EntitlementsManager {
...
 
+  /**
+   * Clears all of the entitlements state and cache.
+   */
+  clear() {
+    this.responsePromise_ = null;
+    this.positiveRetries_ = 0;
+    this.unblockNextNotification();
+    this.storage_.remove(ENTS_STORAGE_KEY);
+    this.storage_.remove(TOAST_STORAGE_KEY);
```
dparikh: Do we need to include ReadyToPay signal in the future too?

May be we can hav a const defined for all the storage keys to clear?

dvoytenko: Yes. Will need RTP as well. But I think we'll just add it here as an extra item.

PR #431, opened 2018-10-09
update analytics context in analytics service when started

sohanirao:


src/runtime/analytics-service.js
```diff
@@ -84,21 +97,63 @@ export class AnalyticsService {
     return this.iframe_;
   }
 
+  /**
+   * @return {!string}
+   * @private
+   */
+  getQueryString_() {
+    return this.doc_.getWin().location.search.split('?')[1];
```
dvoytenko: IIRC, you don't need to remove `?`.

src/runtime/analytics-service.js
```diff
@@ -84,21 +97,63 @@ export class AnalyticsService {
     return this.iframe_;
   }
 
+  /**
+   * @return {!string}
```
dvoytenko: `!string` -> `string`. True for all primitives in JavaScript.

src/runtime/analytics-service.js
```diff
@@ -84,21 +97,63 @@ export class AnalyticsService {
...
+    return this.doc_.getWin().document.referrer;
+  }
+
+  /**
+   * @return {!Promise}
+   */
+  setContext_() {
+    const utmParams = parseQueryString(this.getQueryString_());
+    this.context_.setTransactionId(this.xid_.get());
+    this.context_.setReferringOrigin(this.getReferrer_());
```
dvoytenko: Wrap it in `parseUrl().origin` or `getOrigin()`

src/runtime/analytics-service.js
```diff
@@ -84,21 +97,63 @@ export class AnalyticsService {
...
           this.portResolver_(port);
+          return this.setContext_();
         });
   }
 
   /**
    */
   close() {
+    // Clean up context
+    this.context_ = new AnalyticsContext();
```
dvoytenko: It's defined as `@const` up top.

src/runtime/transaction-id.js
```diff
@@ -0,0 +1,51 @@
...
+    /** @private @const {!Storage} */
+    this.storage_ = deps.storage();
+  }
+
+  /**
+   * Returns the current transaction id
+   *  @return {!Promise<string>}
+   */
+  get() {
+    return this.storage_.get('transaction_id').then(id => {
```
dvoytenko: I believe we decided to drop storage already? Why do we still need it?

sohanirao: Based on offline discussion, we will keep this for now.

src/runtime/analytics-service.js
sohanirao: will replace with todo for now.

src/runtime/analytics-service.js
sohanirao: Done

src/runtime/analytics-service.js
```diff
@@ -19,25 +19,27 @@ import {createElement} from '../utils/dom';
...
   display: 'none',
 };
 
+
 export class AnalyticsService {
   /**
-   * @param {!../model/doc.Doc} doc
-   * @param {!web-activities/activity-ports.ActivityPorts} activityPorts
-   * @param {!../model/page-config.PageConfig} config
+   * @param {!./deps.DepsDef} deps
```
chenshay: The naming is confusing for me. deps isn't very descriptive. When instantiating this you're passing a ConfiguredRuntime type.

sohanirao: Runtime implements those APIs, and this is the pattern we follow for all flows.

src/runtime/analytics-service.js
sohanirao: done

src/runtime/analytics-service.js
dvoytenko: `ConfiguredRuntime` implements `Deps`, which is short for dependencies. But it's not configured runtime itself.

src/runtime/analytics-service.js
```diff
@@ -84,21 +96,62 @@ export class AnalyticsService {
...
+  getQueryString_() {
+    return this.doc_.getWin().location.search.slice(1);
+  }
+
+  /**
+   * @return {string}
+   * @private
+   */
+  getReferrer_() {
+    return parseUrl(this.doc_.getWin().document.referrer).origin;
```
dvoytenko: I'd rather have `parse().origin` to be done outside of this class. This is something that we critically need to test and `getReferrer_` is here to mock out this part of code.

src/runtime/analytics-service.js
```diff
@@ -84,21 +96,62 @@ export class AnalyticsService {
     return this.iframe_;
   }
 
+  /**
+   * @return {string}
+   * @private
+   */
+  getQueryString_() {
+    return this.doc_.getWin().location.search.slice(1);
```
dvoytenko: No need for `slice(1)`

src/runtime/analytics-service.js
```diff
@@ -84,21 +96,62 @@ export class AnalyticsService {
...
         this.args_).then(port => {
           this.portResolver_(port);
+          return this.setContext_();
         });
   }
 
   /**
    */
   close() {
+    // TODO(sohanirao): Clean up context?
```
dvoytenko: No need in this `TODO`. I think we can keep it safely dirty.

src/runtime/analytics-service.js
sohanirao: done

src/runtime/analytics-service.js
sohanirao: done

src/runtime/analytics-service.js
sohanirao: done

PR #445, opened 2018-10-12
Client side config changes to enable analytics for impression events

sohanirao:


src/runtime/entitlements-manager.js
```diff
@@ -43,10 +44,10 @@ export class EntitlementsManager {
     this.win_ = win;
 
     /** @private @const {!../model/page-config.PageConfig} */
-    this.config_ = config;
+    this.pageConfig_ = config;
```
dvoytenko: In this case, let's rename `config` argument as well to `pageConfig`.

src/api/subscriptions.js
```diff
@@ -256,6 +264,7 @@ export const WindowOpenMode = {
 export function defaultConfig() {
   return {
     windowOpenMode: WindowOpenMode.AUTO,
+    analyticsMode: AnalyticsMode.IMPRESSIONS,
```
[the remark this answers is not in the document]
sohanirao: this is already addressed in the latest commit on this PR. Let me know if it still doesn't show up.

src/runtime/entitlements-manager.js
sohanirao: will do

PR #446, opened 2018-10-12
Provide getContext() API for other flows to use

sohanirao:


src/runtime/analytics-service.js
```diff
@@ -105,25 +111,45 @@ export class AnalyticsService {
   }
 
   /**
-   * @return {!Promise}
+   * @private
    */
   setContext_() {
+    if (!this.context_) {
```
dvoytenko: There's a race condition here.

PR #451, opened 2018-10-16
Trigger analytics event logging from pay flow start

sohanirao:


src/runtime/pay-flow.js
```diff
@@ -50,6 +51,9 @@ export class PayStartFlow {
 
     /** @private @const {string} */
     this.sku_ = sku;
+
+    /** @private @const {!AnalyticsService} */
+    this.analyticsService_ = new AnalyticsService(deps);
```
dvoytenko: We shouldn't be creating the analytics service for each use. We should use a singleton. How about creating it in runtime and including it in deps?

sohanirao: Done and a bunch of related to chagnes to that.

src/runtime/entitlements-manager.js
```diff
@@ -118,7 +117,7 @@ export class EntitlementsManager {
       this.responsePromise_ = this.getEntitlementsFlow_();
     }
     return this.responsePromise_.then(response => {
-      if (response.isReadyToPay != null) {
+      if (!(response.isReadyToPay == null)) {
```
dvoytenko: What's wrong with `isReadyToPlay != null`?

sohanirao: while doing tests, I found that even though readyToPay expectation is not specified (null or undefined), the entitlements promise was always false. I wanted to create this check (!(response.isReadyToPay == null || response.isReadyToPay == undefined) and since they are equivalent, I changed it to !(response.isReadyToPay == null) . If this is equivalent to isReadyToPay != null then I can revert change to this line.

PR #454, opened 2018-10-17
Add transactionId to analytics service

dparikh:


src/runtime/analytics-service.js
```diff
@@ -64,6 +68,7 @@ export class AnalyticsService {
 
     /**
      * @private @const {!TransactionId}
+     * TODO(diparikh): Remove once replaced by transactionId().
```
dvoytenko: It's already replaced. No?

src/runtime/analytics-service.js
```diff
@@ -74,6 +79,25 @@ export class AnalyticsService {
     this.lastAction_ = null;
   }
 
+  /**
+   * @param {boolean=} override
+   */
+  setTransactionId(override = false) {
```
dvoytenko: We will be setting it using the actual value that comes from the payments, right? In that case, we should pass it as an argument?

src/runtime/analytics-service.js
```diff
@@ -53,6 +54,9 @@ export class AnalyticsService {
     /** @private @const {string} */
     this.publicationId_ = deps.pageConfig().getPublicationId();
 
+    /** @private {?string} */
+    this.transactionId_ = uuidFast();
```
dvoytenko: According to this, it's never `?`

src/runtime/analytics-service-test.js
```diff
@@ -69,6 +69,7 @@ describes.realWin('AnalyticsService', {}, env => {
   describe('AnalyticsService', () => {
     it('should have analyticsService constructed', () => {
       const activityIframe = analyticsService.getElement();
+      expect(analyticsService.getTransactionId()).to.not.undefined;
```
dvoytenko: But it's never `undefined` right now...

src/runtime/analytics-service.js
```diff
@@ -74,6 +79,25 @@ export class AnalyticsService {
...
+    if (override) {
+      this.transactionId_ = uuidFast();
+    }
+  }
+
+  /**
+   * return {string}
+   */
+  getTransactionId() {
+    if (!this.transactionId_) {
```
dvoytenko: But it's always defined and non-null. When is this possible then?

src/runtime/analytics-service.js
```diff
@@ -74,6 +79,25 @@ export class AnalyticsService {
...
+   * @param {boolean=} override
+   */
+  setTransactionId(override = false) {
+    if (override) {
+      this.transactionId_ = uuidFast();
+    }
+  }
+
+  /**
+   * return {string}
```
dvoytenko: `@`

src/runtime/analytics-service.js
dparikh: Fixed. Originally, I had set the initial value to `null`;

src/runtime/analytics-service.js
dparikh: I have `setTransactionId` only in case we want to reset for some reason.
When we receive the value from the payments, we could just check if it matches existing `transactionId`

WDYT?

src/runtime/analytics-service.js
dparikh: I mean to remove this and associated tests once this PR looks good.
Removed now.

src/runtime/analytics-service.js
dparikh: Removed this check. This was again left-over from my original code where in the constructor, I would not set the value.

src/runtime/analytics-service.js
dparikh: Done

src/runtime/analytics-service-test.js
dparikh: yes, just to ensure that it is NOT `undefined`

src/runtime/analytics-service.js
```diff
@@ -62,18 +65,30 @@ export class AnalyticsService {
...
 
     /** @private {?Promise} */
     this.lastAction_ = null;
   }
 
+  /**
+   * @param {string} transactionId
+   * @param {boolean=} override
+   */
+  setTransactionId(transactionId, override = false) {
```
dvoytenko: Still not quire sure about this. Why would one ever call `setTransactionId` w/o `override = true`? Alternatively, we could also drop `this.transactionId_` and always just use `this.context_.get/setTransactionId()`

src/runtime/analytics-service.js
```diff
@@ -104,9 +119,6 @@ export class AnalyticsService {
     return this.doc_.getWin().document.referrer;
   }
 
-  /**
-   * @return {!Promise}
-   */
   setContext_() {
```
dvoytenko: Please add `@private`

src/runtime/analytics-service.js
```diff
@@ -62,18 +65,30 @@ export class AnalyticsService {
...
     this.lastAction_ = null;
   }
 
+  /**
+   * @param {string} transactionId
+   * @param {boolean=} override
+   */
+  setTransactionId(transactionId, override = false) {
+    if (override) {
+      this.transactionId_ = transactionId;
```
dvoytenko: It also needs to set `this.context_.setTransactionId(...)` to the new value. This would need a test.

src/runtime/analytics-service.js
dparikh: oops! Done

src/runtime/analytics-service.js
dparikh: Done.

src/runtime/analytics-service-test.js
dvoytenko: Maybe let's replace this with a match of some sort? E.g. `expect(getTransactionId()).to.match(/ddd-ddd-ddd..../)`?

src/runtime/analytics-service-test.js
```diff
@@ -167,8 +167,7 @@ describes.realWin('AnalyticsService', {}, env => {
         expect(request.getContext().getUtmMedium()).to.equal('email');
         expect(request.getContext().getUtmSource()).to.equal('scenic');
         expect(request.getContext().getUtmName()).to.equal('campaign');
-        expect(request.getContext().getTransactionId()).to.equal(
-            'google-transaction-0');
+        expect(request.getContext().getTransactionId()).to.be.not.undefined;
```
dvoytenko: Ditto.

src/runtime/analytics-service.js
```diff
@@ -122,9 +134,7 @@ export class AnalyticsService {
     if (source) {
       this.context_.setUtmSource(source);
     }
-    return this.xid_.get().then(id => {
-      this.context_.setTransactionId(id);
-    });
+    this.context_.setTransactionId(uuidFast());
```
dvoytenko: That will reset it in an expected way if it was already initialized. Let's do 

```
// Make sure the transactionId is initialized.
this.context_.setTransactionId(this.getTransactionId());
```

And this situation needs a test. Again, the repro scenario:
1. Call `const value = getTransactionId()`
2. Then get `this.setContext_()` to be called.
3. As the result, this should be true: `value == this.getTransactionId()`. Or in other words, transaction id has not changed.

dparikh: Done

src/runtime/analytics-service-test.js
dparikh: Done.

PR #468, opened 2018-10-25
Replace Subscription implementation - part 1

chenshay:


src/runtime/pay-flow.js
```diff
@@ -82,6 +94,8 @@ export class PayStartFlow {
       'swg': {
         'publicationId': this.pageConfig_.getPublicationId(),
         'skuId': this.sku_,
+        'oldSkuId': this.oldSku_,
```
dvoytenko: Ideally we wouldn't send these keys at all if they are not relevant.

chenshay: done

src/runtime/pay-flow.js
```diff
@@ -71,21 +80,37 @@ export class PayStartFlow {
...
    */
   start() {
-    // Start/cancel events.
-    this.deps_.callbacks().triggerFlowStarted(SubscriptionFlows.SUBSCRIBE, {
+    const triggerFlowKeys = {
       'sku': this.sku_,
-    });
+    };
+
+    const swgKeys = {
```
dvoytenko: I'd call it `swgPayload` or `swgPaymentRequest`

src/runtime/pay-flow.js
```diff
@@ -71,21 +80,37 @@ export class PayStartFlow {
...
+      'publicationId': this.pageConfig_.getPublicationId(),
+      'skuId': this.sku_,
+    };
+
+    if (!!this.oldSku_) {
```
dvoytenko: Just `if (this.oldSku_) {...}`

src/runtime/pay-flow.js
```diff
@@ -71,21 +80,37 @@ export class PayStartFlow {
...
+      triggerFlowKeys['oldSku'] = this.oldSku_;
```
dvoytenko: This is good, but to clarify the priority: payment payload and trigger params are very different - the payment payload travels across several layers to in a set of separate system stacks. That's why I prefer to keep that payload as clean as possible. Trigger parameters are different from this - it's all ours.

chenshay: ack

src/runtime/pay-flow.js
chenshay: done

src/runtime/pay-flow.js
chenshay: done

src/runtime/pay-flow.js
```diff
@@ -45,8 +45,11 @@ export class PayStartFlow {
   /**
    * @param {!./deps.DepsDef} deps
    * @param {string} sku
+   * @param {?string|undefined} opt_oldSkuId
+   * @param {?string|undefined} opt_replaceSkuProrationMode
```
dvoytenko: This appears to be an enum?

chenshay: changed to enum

src/runtime/runtime.js
```diff
@@ -291,6 +291,23 @@ export class Runtime {
         .then(runtime => runtime.showOffers(opt_options));
   }
 
+  /**
+   * Replaces an existing subscription with a different one.
+   *
+   * @param {string} newSku
+   * @param {string} oldSku
+   * @param {?../api/subscriptions.ReplaceSkuProrationMode|undefined} opt_prorationMode
+   * @return {Promise}
```
dvoytenko: What's the response though?

chenshay: fixed

src/api/subscriptions.js
```diff
@@ -257,6 +257,13 @@ export const WindowOpenMode = {
   REDIRECT: 'redirect',
 };
 
+/**
+ * @enum {string}
+ */
+export const ReplaceSkuProrationMode = {
+  UNKNOWN: '0',
```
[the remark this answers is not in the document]
chenshay: fixed

src/api/subscriptions.js
```diff
@@ -257,6 +257,16 @@ export const WindowOpenMode = {
...
 };
 
+/**
+ * @enum {string}
+ */
+export const ReplaceSkuProrationMode = {
+  // The replacement takes effect immediately, and the remaining time will
+  // be prorated and credited to the user. This is the current default
+  // behavior.
+  IMMEDIATE_WITH_TIME_PRORATION: '1',
```
dvoytenko: The values should be strings as well - if we want to map them to numbers - we should do it internally. Realistically API users cannot use enums - they are here for our compilation needs and documentation. And we don't necessary want them to be passing us cryptic numbers.

src/api/subscriptions.js
```diff
@@ -323,3 +333,20 @@ export let SaveSubscriptionRequestCallback;
...
+ *  specific proration mode to charge the user.
+ *  The default is IMMEDIATE_WITH_TIME_PRORATION.
+ * 
+ *  @typedef {{
+ *    skuId: string,
+ *    oldSkuId: (string|undefined),
+ *    replaceSkuProrationMode: (ReplaceSkuProrationMode|undefined),
+ * }}
+ */
+export let SubscriptionRequest;
```
dvoytenko: nit: +1 new line

src/runtime/pay-flow.js
```diff
@@ -59,8 +62,10 @@ export class PayStartFlow {
     /** @private @const {!../components/dialog-manager.DialogManager} */
     this.dialogManager_ = deps.dialogManager();
 
-    /** @private @const {string} */
-    this.sku_ = sku;
+    /** @private @const {!../api/subscriptions.SubscriptionRequest} */
+    this.subscriptionRequestKeys_ =
```
dvoytenko: Let's name `subscriptionRequest_`

src/runtime/pay-flow.js
```diff
@@ -71,21 +76,23 @@ export class PayStartFlow {
    * @return {!Promise}
    */
   start() {
+    // Add the 'publicationId' key to the subscriptionRequestKeys_ object.
+    const swgPaymentRequest =
+        Object.assign({}, this.subscriptionRequestKeys_, {
```
dvoytenko: We also will need to handle that enum question I mentioned above.

src/api/subscriptions.js
chenshay: done

src/runtime/pay-flow.js
chenshay: done

src/api/subscriptions.js
```diff
@@ -257,6 +257,27 @@ export const WindowOpenMode = {
...
+  // be prorated and credited to the user. This is the current default
+  // behavior.
+  IMMEDIATE_WITH_TIME_PRORATION: 'IMMEDIATE_WITH_TIME_PRORATION',
+};
+
+/**
+ * String values input by the publisher are mapped to the number values.
+ * @enum {number}
+ */
+export const ReplaceSkuProrationModeMapping = {
```
dvoytenko: This is an internal mapping detail, so we should move closer to where it's used within the runtime.

src/runtime/pay-flow.js
```diff
@@ -71,21 +77,30 @@ export class PayStartFlow {
    * @return {!Promise}
    */
   start() {
+    // Map the proration mode to the enum value (if proration exists).
+    const prorationMode = this.subscriptionRequest_.replaceSkuProrationMode;
+    if (prorationMode) {
+      this.subscriptionRequest_.replaceSkuProrationMode =
```
dvoytenko: We need to update this directly in the payment request, not the input structure. Because we use it for analytics, etc - it's not good if we update it inline and give API users back a different object than the one they provided.

src/api/subscriptions.js
chenshay: done

src/runtime/pay-flow.js
chenshay: done

src/runtime/pay-flow.js
```diff
@@ -71,21 +89,30 @@ export class PayStartFlow {
...
+          ReplaceSkuProrationModeMapping[prorationMode];
+    }
+
     // Start/cancel events.
-    this.deps_.callbacks().triggerFlowStarted(SubscriptionFlows.SUBSCRIBE, {
-      'sku': this.sku_,
-    });
-    this.analyticsService_.setSku(this.sku_);
+    this.deps_.callbacks().triggerFlowStarted(
+        SubscriptionFlows.SUBSCRIBE, swgPaymentRequest);
```
dvoytenko: This should be `this.subscriptionRequest_` right?

src/api/subscriptions.js
dvoytenko: This should be removed now?

src/runtime/pay-flow.js
```diff
@@ -38,15 +41,28 @@ import {
 } from '../utils/json';
 import {AnalyticsEvent} from '../proto/api_messages';
 
+
+/**
+ * String values input by the publisher are mapped to the number values.
+ * @enum {number}
```
dvoytenko: This is pretty critical, but this cannot be an enum. Enum's keys (but not values) are obfuscated. Thus, if you try to do `ReplaceSkuProrationModeMapping['THE_ACTUAL_VALUE']` - it won't have the right mapping. Thus, this has to be an object, and the keys must be in quotes.

chenshay: done

PR #479, opened 2018-11-09
JsError service to report client-side errors

dvoytenko:


src/runtime/jserror-test.js
```diff
@@ -0,0 +1,96 @@
...
+      expect(src.pathname)
+          .to.equal('/$frontend$/_/SubscribewithgoogleClientUi/jserror');
+      expect(params['script']).to.equal('$frontend$/swg/js/v1/swg.js');
+      expect(params['line']).to.equal('1');
+      expect(params['error']).to.equal('Error: A B: broken');
+      expect(params['trace']).to.match(/Error: A B: broken/);
+    });
+  });
+
+  it('should create an error if not provided', () => {
```
chenshay: If not provided what?

dvoytenko: Would this be better? "Should create an error if one not provided"

PR #484, opened 2018-11-19
Start LinkComplete flow after SaveSubscription flow

sohanirao:


src/runtime/link-accounts-flow.js
```diff
@@ -271,24 +271,38 @@ export class LinkSaveFlow {
...
+        return this.activityIframeView_.port().then(port => {
+          return acceptPortResultData(
+              port,
+              feOrigin(),
+              /* requireOriginVerified */ true,
+              /* requireSecureChannel */ true);
+        }).then(result => {
+          // The flow is complete.
+          if (!result['linked']) {
+            throw new Error('save subscription failed');
```
dvoytenko: This shouldn't be an error. User refusing to link should be transparent to the publisher.

src/runtime/link-accounts-flow.js
```diff
@@ -271,24 +271,38 @@ export class LinkSaveFlow {
...
+          }
+          return this.dialogManager_.completeView(this.activityIframeView_);
+        }).then(() => {
+          this.deps_.dialogManager().popupClosed();
+          linkConfirm = new LinkCompleteFlow(this.deps_);
+          this.deps_.callbacks().triggerFlowStarted(SubscriptionFlows.LINK_ACCOUNT);
+          return linkConfirm.start();
```
dvoytenko: So, is `index` passed anywhere?

src/runtime/link-accounts-flow.js
```diff
@@ -271,24 +271,38 @@ export class LinkSaveFlow {
...
+          this.deps_.callbacks().triggerFlowStarted(SubscriptionFlows.LINK_ACCOUNT);
+          return linkConfirm.start(); 
+        }).then(() => {
+          this.deps_.callbacks().triggerLinkProgress();
+          return linkConfirm.whenComplete().then(() => {
+            this.deps_.callbacks().triggerLinkComplete();
+            this.deps_.callbacks().resetLinkProgress();
+            return true;
+          }, reason => {
+            this.deps_.callbacks().triggerFlowCanceled(SubscriptionFows.LINK_ACCOUNT);
```
dvoytenko: The linking itself is not canceled here, is it? It's already linked, I believe.

src/runtime/link-accounts-flow.js
sohanirao: yes I'll upload a new version. this was before i made the change to boq to send the user linked.

src/runtime/link-accounts-flow.js
sohanirao: Regardless of how we are arriving at the link confirmation flow, if it is cancelled we trigger the callback. I expect the same callback will remove the link like it would be expected to for the linkAccount() flow, no?

src/runtime/link-accounts-flow.js
sohanirao: You are right, we should just return here.

src/runtime/link-accounts-flow.js
```diff
@@ -218,6 +218,15 @@ export class LinkSaveFlow {
...
     this.requestPromise_ = null;
 
+    /** @private {?Promise} */
+    this.linkedPromise_ = null;
+
+    /** @private {?Promise} */
+    this.confirmPromise_ = null;
+
+    /** @private {?Promise} */
+    this.completePromise_ = null;
```
dvoytenko: Is this promise actually used anywhere?

sohanirao: actually we want this to be called only once
this.dialogManager_.completeView(this.activityIframeView_);
 I thought a completePromise looks nice and can be used for testing to ensure a dialog was completed. I ended not needing in the tests though.

dvoytenko: FWIW, I think calling `completeView` numerous times is fine. Thought it's ok to have a check here as well. I was just reacting to a null promise which is generally atypical.

sohanirao: Addressed in the next patch.

PR #489, opened 2018-11-28
Support node 8.x and 10.x

dparikh:


package.json
```diff
@@ -5,7 +5,7 @@
   "description": "Subscribe with Google",
   "main": "index.js",
   "engines": {
-    "node": ">=8",
+    "node": "^8.0.0 || ^10.0.0",
```
dvoytenko: What's wrong with v 9 and 11?

dparikh: There was not LTS for 9.x branch and 10.x is the latest LTS.
(I got this from AMP)

dvoytenko: Ok

PR #493, opened 2018-12-06
Fix pay.js to avoid calls to endsWith. Upstream fix is comming in parallel

dvoytenko:


third_party/gpay/src/validator.js
```diff
@@ -111,10 +111,6 @@ function apiV2DoesMerchantSupportSpecifiedCardType(
  * message.
  */
 function validateSecureContext() {
-  if (window.location.hostname.endsWith(Constants.TRUSTED_DOMAIN)) {
-    // This is for local development.
```
cramforce: LOL

PR #495, opened 2018-12-07
Make CSS injection use Doc to correctly handle shadow dom case.

jpettitt:
Make CSS injection use `Doc.getHead()` and `Doc.getDocumentRoot()` to correctly handle AMP document mode when in shadow dom mode (IE inject the CSS inside the shadow root if needed).

### DIFF
--- a/src/utils/dom.js
+++ b/src/utils/dom.js
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+import {Doc} from '../model/doc';
 import {assert} from './log';
 import {setStyles} from './style';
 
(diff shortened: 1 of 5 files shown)


src/utils/dom.js
```diff
@@ -102,16 +103,16 @@ export function removeChildren(parent) {
...
 /**
  * Injects the provided styles in the HEAD section of the document.
- * @param {!Document} doc The document object.
+ * @param {!Doc} doc The document object.
  * @param {string} styleText The style string.
  * @return {!Element}
  */
 export function injectStyleSheet(doc, styleText) {
-  const styleElement = createElement(doc, 'style', {
+  const styleElement = createElement(doc.getRootNode(), 'style', {
```
dvoytenko: This should be `createElement(doc.getWin().document)`. Unfortunately, `ShadowRoot` interface does not provide `createElement` API. A bit surprised the compiler is not complaining about this :(

src/utils/dom.js
```diff
@@ -102,16 +103,16 @@ export function removeChildren(parent) {
 
 /**
  * Injects the provided styles in the HEAD section of the document.
- * @param {!Document} doc The document object.
+ * @param {!Doc} doc The document object.
```
dvoytenko: I also see `loading-view-test.js` and `dom-test.js` using `injectStyleSheet`. Do they need to be also updated?

PR #496, opened 2018-12-07
Initial Draft at a client interface for proto buf lite

sohanirao:


src/components/activity-client-interface.js
```diff
@@ -0,0 +1,127 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {AnalyticsContext} from '../proto/api_messages';
+import {ActivityIframePort} from 'web-activities/activity-ports';
+import {ActivityMode, ActivityResult} from 'web-activities/activities';
+
+/**
+ * @typedef {!Array<(string|boolean|number|null|!Array<(string|boolean|number|null)>)>} SerializedBuffer
```
dvoytenko: Is this just a bad copy-paste? Speaking of which - I still think we should move all of our protolite `!Array<(string|.......)>` to just `!Array`. Such a big union of types is somewhat confusing. It's also incomplete.

src/components/activity-client-interface.js
```diff
@@ -0,0 +1,127 @@
...
+ */
+export class ActivityClientInterface{
```
dvoytenko: We need to see what's the best way to wrap around these APIs. Activities has two main constructs: ActivityPorts and ActivityPort. ActivityPort could represent an iframe or a popup. They are similar in how they start and end, but not in other details.

In SwG we have 3-4 notions (or maybe even more):
1. Popup
2. Dialog View iframe.
3. Toast iframe.
4. Service iframe.

Intuitively I feel like we'd be better of if we followed the initial API when wrapping. We could even call it `SwgActivityPorts` and `SwgActivityPort`. But let's discuss in more detail.

src/components/activity-client-interface.js
```diff
@@ -0,0 +1,127 @@
...
+    this.callbackMap_ = {};
+  }
+
+  /**
+   *
+   * @param {string} pub_id
+   * @param {?AnalyticsContext} context
+   */
+  start(pub_id, context) {
+    this.message(
```
dvoytenko: This should be implicit as part of the `ActivityPorts.openIframe` API. It's critical that we supply `publicationId` as part of connection mechanism. Otherwise we take a risk of allowing messages before security parameters are confirmed. And it'd be nice to pass the rest of the context there as well.

src/components/activity-client-interface.js
```diff
@@ -0,0 +1,127 @@
...
+  getMode() {
+    return this.activityIframePort_.getMode();
+  }
+
+  /**
+   * Waits until the activity port is connected to the host.
+   * @return {!Promise}
+   */
+  connect() {
+    return this.activityIframePort_.connect();
```
dvoytenko: Is there a `connect` API? I believe it's also part of `openIframe`

src/components/activity-client-interface.js
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+  disconnect() {
```
dvoytenko: Ditto?

src/components/activity-client-interface.js
```diff
@@ -0,0 +1,127 @@
...
+  resized() {
+    this.activityIframePort_.resized();
+  }
+
+  /**
+   *
+   * @param {!SerializedBuffer} request
+   */
+  execute(request) {
+    this.message({req: request});
```
dvoytenko: `req` should be in quotes. Also thinking we should make new fields more distinct - it'll be easier to tell them apart until the full transition. WDYT if we use all caps for them?

src/components/activity-client-interface.js
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+  /**
+   * @param {string} responseType
```
dvoytenko: The idea is not to use string but the actual type which would give us proper typing. It'd look something like this:

```
/**
 * @param {function(new:T)} type
 * @param {function(T)} callback
 * @template T
 */
```

src/components/activity-client-interface.js
```diff
@@ -0,0 +1,127 @@
...
+   * @param {function(!{!SerializedBuffer})} callback
+   */
+  on(responseType, callback) {
+    if (!this.callbackMap_[responseType]) {
```
dvoytenko: If we use a single handler system (which is great), we should throw an error for duplicate handlers.

src/components/activity-client-interface.js
```diff
@@ -0,0 +1,127 @@
...
+      this.callbackMap_[responseType] = callback;
+    }
+    this.onMessage(data => {
```
dvoytenko: We probably need only one message handler?

src/components/activity-client-interface.js
```diff
@@ -0,0 +1,127 @@
...
+      if (data == null || data[0] == null) {
+        return;
+      }
+      const dataType = data[0];
+      const cb = this.callbackMap_[dataType];
```
dvoytenko: As it is, this will be causing lots of `undefined is not a function` errors.

PR #502, opened 2018-12-26
replaceSku - compute the old sku for the publisher

chenshay:


src/runtime/runtime.js
```diff
@@ -715,9 +716,22 @@ export class ConfiguredRuntime {
...
   /** @override */
   subscribe(skuOrSubscriptionRequest) {
-    if (typeof skuOrSubscriptionRequest != 'string' &&
-        !isExperimentOn(this.win_, ExperimentFlags.REPLACE_SUBSCRIPTION)) {
-      throw new Error('Not yet launched!');
+    if (typeof skuOrSubscriptionRequest != 'string') {
+      if (!isExperimentOn(this.win_, ExperimentFlags.REPLACE_SUBSCRIPTION)) {
+        throw new Error('Not yet launched!');
+      }
+      const currentSkuId = this.pageConfig_.getProductId();
```
dvoytenko: productId and sku are not the same thing.

PR #508, opened 2019-01-10
Save Subscription API returns a promise of a boolean status

sohanirao:


src/runtime/link-accounts-flow.js
```diff
@@ -237,26 +234,41 @@ export class LinkSaveFlow {
...
+      startPromise = Promise.reject(
+          createCancelError(this.win_, 'not linked'));
+    }
+    const completePromise = startPromise.then(() => {
+      this.deps_.callbacks().triggerLinkProgress();
+      return linkConfirm.whenComplete();
+    });
+
+    return completePromise.then(() => {
+      return true;
```
chenshay: why does this need to return true? why not just completePromise?

src/runtime/link-accounts-flow.js
```diff
@@ -306,44 +317,27 @@ export class LinkSaveFlow {
...
+      this.complete_();
+      // Handle cancellation from user, link confirm start or completion here
       if (isCancelError(reason)) {
         this.deps_.callbacks().triggerFlowCanceled(
             SubscriptionFlows.LINK_ACCOUNT);
         return false;
       }
-      // In case this flow wasn't complete, complete it here
-      this.complete_();
       throw reason;
```
chenshay: if the error is not a cancel error - then wouldn't you want to return false after throwing a reason?

src/runtime/link-accounts-flow.js
sohanirao: when complete Promise is done, we know that the process succeeded. When something fails, we check if its due to cancellation from the user at any point in the process and we return false. otherwise we let the error be thrown to the caller.

The expected outcome is that upon calling this API, the caller gets a true or false upon completion or error.

src/runtime/link-accounts-flow.js
sohanirao: We still have to complete the existing flow with the dialog manager. I actually moved the complete above isCancelError..mabye need to push the new patch.

PR #509, opened 2019-01-29
Do not expose port out of ActivityIframeView

sohanirao:
Activity Port doesn't need to be exposed from ActivityIframeView. It is already passed into the constructor. Handle the only case where the port is used from this Object by providing another API. This change enables ActivityIframeView to potentially use another ActivityPort internally that can communicate differently (use Protos) than existing ActivityPort.


src/ui/activity-iframe-view.js
```diff
@@ -124,28 +125,29 @@ export class ActivityIframeView extends View {
    * @return {!Promise}
    */
   onOpenIframeResponse_(port, dialog) {
-    this.port_ = port;
+    this.portResolved_ = port;
```
chenshay: I'd rather call this port_ and the method portPromise_()

src/ui/activity-iframe-view.js
```diff
@@ -124,28 +125,29 @@ export class ActivityIframeView extends View {
...
-    return this.port_.whenReady();
+    return this.portResolved_.whenReady();
   }
 
   /**
    * @return {!Promise<!web-activities/activity-ports.ActivityIframePort>}
+   * @private
    */
-  port() {
+  port_() {
```
chenshay: change to portPromise_

src/ui/activity-iframe-view.js
sohanirao: there's already a variable called portPromise_ on Line 91

src/ui/activity-iframe-view.js
chenshay: then change to getPortPromise_

sohanirao: done

PR #512, opened 2019-02-14
Expand allowed @type values

jpettitt:
Expand allowed types
 
fixes https://github.com/ampproject/amphtml/issues/20747

### DIFF
--- a/src/model/page-config-resolver-test.js
+++ b/src/model/page-config-resolver-test.js
@@ -177,6 +177,22 @@ describes.realWin('PageConfigResolver', {}, env => {
           .to.equal('pub1:basic');
     });
 
+    it('should allow alternate LD Types', () => {
+      schema['@type'] = ['CreativeWork'];
+      addJsonLd(schema);
+      readyState = 'complete';
+      expect(new PageConfigResolver(gd).check().getProductId())
+          .to.equal('pub1:basic');
+    });
+
+    it('should allow work with multiple LD types', () => {
+      schema['@type'] = ['NewsArticle', 'Person'];
+      addJsonLd(schema);
+      readyState = 'complete';
+      expect(new PageConfigResolver(gd).check().getProductId())
+          .to.equal('pub1:basic');
+    });
+
     it('should allow full namespace for LD type', () => {
       schema['@type'] = ['http://schema.org/NewsArticle'];
       addJsonLd(schema);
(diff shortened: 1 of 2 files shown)


src/model/page-config-resolver.js
```diff
@@ -268,16 +283,22 @@ class JsonLdParser {
 
   /**
    * @param {!Object} json
-   * @param {string} expectedType
+   * @param {Array<string>} expectedTypes
    * @return {boolean}
    */
-  checkType_(json, expectedType) {
+  checkType_(json, expectedTypes) {
```
sohanirao: could we export checkType() out of this and MicrodataParser method? Seems like duplicate functionality except for the argument.
In this class, isValidType() which takes two arrays. Here it gets invoked with this.valueArray_(json, '@type') as argument and in MicrodataParser, it gets itemtype.split(/\s+/)

It seems like types allowed and its management happens outside both JsonLdParser and MicrodataParser

Seems like this guidance on what's okay and not okay is coming outside of these Parsers, which decides what is considered valid configuration. Might be an overkill at this point to have a class that dictates our standard clearly (that would include type checking as well as allowed types). But exporting checkType() out of these classes will atleast consolidate common behavior into one place and be clear that these parsers take guidance from outside. Thoughts?

src/model/page-config-resolver-test.js
```diff
@@ -177,6 +177,22 @@ describes.realWin('PageConfigResolver', {}, env => {
...
+    });
+
+    it('should allow work with multiple LD types', () => {
+      schema['@type'] = ['NewsArticle', 'Person'];
+      addJsonLd(schema);
+      readyState = 'complete';
+      expect(new PageConfigResolver(gd).check().getProductId())
+          .to.equal('pub1:basic');
+    });
+
```
sohanirao: could we have a negative test case for a number of types but none of them feature in the Allowed type list.

src/model/page-config-resolver-test.js
```diff
@@ -300,6 +316,27 @@ describes.realWin('PageConfigResolver', {}, env => {
...
+            </div> \
+          </div>';
+      addMicrodata(divElement);
+      const resolver = new PageConfigResolver(gd);
+      return resolver.resolveConfig().then(config => {
+        expect(config.isLocked()).to.be.false;
+        expect(config.getProductId()).to.equal('pub1:premium');
+      });
+    });
+
```
sohanirao: could we have a negative test case for a number of types but none of them feature in the Allowed type list.

src/model/page-config-resolver.js
```diff
@@ -157,7 +172,7 @@ class JsonLdParser {
         continue;
       }
       element[ALREADY_SEEN] = true;
-      if (element.textContent.indexOf('NewsArticle') == -1) {
+      if (!RE_ALLOWED_TYPES.test(element.textContent)) {
```
sohanirao: is there a reason this cannot use checkType() as well?

src/model/page-config-resolver.js
```diff
@@ -338,25 +359,22 @@ class MicrodataParser {
     for (let node = current;
         node && !node[alreadySeen]; node = node.parentNode) {
       node[alreadySeen] = true;
-      if (node.hasAttribute('itemscope')) {
+      // document notdes don't have hasAttribute
```
sohanirao: nodes

src/model/page-config-resolver.js
jpettitt: done

src/model/page-config-resolver-test.js
jpettitt: done

src/model/page-config-resolver-test.js
jpettitt: done

src/model/page-config-resolver.js
jpettitt: using `checkType` will require parsing json which is expensive (publisher can include entire article text in ld+json markup so it can be quite long).  This is an optimization to see if our allowed types are even present before we go in ad to a real parse.  It only matters if swg.js is included on pages that are not our type so it may be unneeded. WDYT?

src/model/page-config-resolver.js
jpettitt: fixed

PR #516, opened 2019-02-27
Contribution UI

dparikh:
Added contribution API (Experimental)

```
subscriptions.contribute(sku);
```

TODO:
Need to update the Payment confirmation views on gPay complete.

### DIFF
--- a/src/api/subscriptions.js
+++ b/src/api/subscriptions.js
@@ -126,6 +126,12 @@ export class Subscriptions {
    */
   subscribe(skuOrSubscriptionRequest) {}
 
+   /**
+   * Starts contributions purchase flow.
+   * @param {string|SubscriptionRequest} skuOrSubscriptionRequest
+   */
+  contribute(skuOrSubscriptionRequest) {}
+
   /**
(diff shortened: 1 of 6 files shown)


src/api/subscriptions.js
```diff
@@ -279,6 +285,16 @@ export const ReplaceSkuProrationMode = {
...
 
+/**
+ * The product type. The Offers UI is rendered differently based on the
+ * ProductType and 'product_type' parameter is passed to Payments flow to render
+ * the GPay content accordingly. ProductType Defaults to 'SUBSCRIPTION'.
+ * @enum {string}
+ */
+export const ProductType = {
+  SUBSCRIPTION: 'SUBSCRIPTION',
+  UI_CONTRIBUTION: 'UI_CONTRIBUTION',
```
chenshay: nit: why is this UI_CONTRIBUTION? Why not just CONTRIBUTION?

dparikh: This type is defined by Play. The main reason is that the impact is only in the UI, the internal details remain same.
The enum is defined in the Play.

PR #517, opened 2019-02-28
Add debugLog()

jpettitt:
Adds a `debugLog()` utility function that is a gated by a `location.hash` value containing `swg.dbg=1`. Messages look like:

`[Subscriptions] message text`

Adds a `debugLog(config)` and `debugLog(mode)` to dump the config and mode on startup.

image

Fixes b/122649408

### DIFF
--- a/src/utils/log.js
+++ b/src/utils/log.js
@@ -14,6 +14,20 @@
  * limitations under the License.
  */
 
+ /**
+  * Debug logger, only log message if #swg.log=1
+  * @param {...*} var_args [decription]
+  */
+export function debugLog(var_args) {
+  if (/swg.dbg=1/.test(self.location.hash)) {
+    const logArgs = ['[Subscriptions]'];
+    for (let i = 0; i < arguments.length; i++) {
+      logArgs.push(arguments[i]);
+    }
+    log.apply(log, logArgs);
+  }
+}
+
 /**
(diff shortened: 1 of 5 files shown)


src/utils/log.js
```diff
@@ -14,6 +14,20 @@
  * limitations under the License.
  */
 
+ /**
+  * Debug logger, only log message if #swg.log=1
+  * @param {...*} var_args [decription]
+  */
+export function debugLog(var_args) {
+  if (/swg.dbg=1/.test(self.location.hash)) {
```
dvoytenko: I'd rather call it `swg.debug`. Also, do you think we should calculate it once and cache?

src/utils/log.js
```diff
@@ -14,6 +14,20 @@
...
+    const logArgs = ['[Subscriptions]'];
```
dvoytenko: Nit: 
```
const logArgs = Array.prototype.slice.call(arguments, 0);
logArgs.unshift('[Subscriptions]');
```

src/utils/log.js
jpettitt: will change the dbg -> debug.

Having it be a cache value makes testing harder and since it's not in any loops I left it as the simple test..

src/utils/log.js
jpettitt: fixed

PR #520, opened 2019-03-12
Contributions: preserving  product_type

dparikh:
Store and pass`product_type` in the `subscriptionResponse` for `payconfirmiframe` to use.

This allows changes to the payconfirmiframe page to render the different text/label.

### DIFF
--- a/src/runtime/pay-flow-test.js
+++ b/src/runtime/pay-flow-test.js
@@ -285,7 +285,7 @@ describes.realWin('PayCompleteFlow', {}, env => {
     jserrorMock.verify();
   });
 
-  it('should have valid flow constructed ******', () => {
+  it('should have valid flow constructed', () => {
     const purchaseData = new PurchaseData();
     const userData = new UserData('ID_TOK', {
       'email': 'test@example.org',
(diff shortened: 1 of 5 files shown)


src/runtime/pay-flow-test.js
```diff
@@ -281,14 +285,15 @@ describes.realWin('PayCompleteFlow', {}, env => {
     jserrorMock.verify();
   });
 
-  it('should have valid flow constructed', () => {
+  it('should have valid flow constructed ******', () => {
```
chenshay: what are all these stars for?

src/runtime/pay-flow-test.js
```diff
@@ -281,14 +285,15 @@ describes.realWin('PayCompleteFlow', {}, env => {
...
     const purchaseData = new PurchaseData();
     const userData = new UserData('ID_TOK', {
       'email': 'test@example.org',
     });
     const entitlements = new Entitlements('service1', 'RaW', [], null);
     const response = new SubscribeResponse(
-        'RaW', purchaseData, userData, entitlements);
+        'RaW', purchaseData, userData, entitlements,
+        ProductType.SUBSCRIPTION, entitlements);
```
chenshay: entitlements is here twice. You need to remove one of them

src/runtime/pay-flow-test.js
```diff
@@ -332,6 +338,7 @@ describes.realWin('PayCompleteFlow', {}, env => {
           _client: 'SwG $internalRuntimeVersion$',
           publicationId: 'pub1',
           loginHint: 'test@example.org',
+          productType: undefined,
```
chenshay: is it supposed to be null or undefined or doesn't matter?

src/runtime/pay-flow-test.js
dparikh: leftover from debugging...Removed.

src/runtime/pay-flow-test.js
dparikh: Removed.

src/runtime/pay-flow-test.js
dparikh: Here `productType: undefined` is correctly matched.

PR #522, opened 2019-03-15
button-api: fix Spanish button titles

plr108:
Change es-latam and es-latn TITLE_LANG_MAP values to match button text


src/runtime/button-api.js
```diff
@@ -23,8 +23,8 @@ const TITLE_LANG_MAP = {
   'ar': 'الاشتراك عبر Google',
   'de': 'Abonnieren mit Google',
   'es': 'Suscríbete con Google',
-  'es-latam': 'Suscribirse con Google',
-  'es-latn': 'Suscribirse con Google',
+  'es-latam': 'Suscríbete con Google',
```
jpettitt: This will only change the button title, it won't change the text on the button which comes from the SVG.

dparikh: The SVG contains correct text. The title was not matching with the text.

PR #533, opened 2019-04-29
Smartbox implementation

dparikh:
Initial implementation.

### DIFF
--- a/assets/swg-button.css
+++ b/assets/swg-button.css
@@ -257,3 +257,17 @@
 .swg-button-dark:lang(zh-tw):after {
   background-image: url("i18n/b-zh-tw-dk.svg");
 }
+
+.swg-smart-button {
+  border: none;
+  padding: 0;
+  background: transparent;
+  min-height: 126px;
+  position: relative;
+  width: 300px;
+  border-radius: 0;
+}
+
+.swg-smart-button:focus {
+  outline: none;
+}
--- a/assets/swg-button.css
+++ b/assets/swg-button.css
@@ -91,6 +91,25 @@
   background-color: #5f6368;
 }
 
+/*
+  This class defines style only for the container button.
+  All button styles are rendered inside the nested iframe.
+*/
+.swg-smart-button {
+  border: none;
+  padding: 0;
+  background: transparent;
+  min-height: 126px;
+  position: relative;
+  min-width: 300px;
+  width: 300px;
+  border-radius: 0;
+  overflow: hidden;
+}
+
+.swg-smart-button:focus {
+  outline: none;
+}
 
 /**
  * i18n bindings
@@ -256,18 +275,4 @@
 }
 .swg-button-dark:lang(zh-tw):after {
   background-image: url("i18n/b-zh-tw-dk.svg");
-}
-
-.swg-smart-button {
-  border: none;
-  padding: 0;
-  background: transparent;
-  min-height: 126px;
-  position: relative;
-  width: 300px;
-  border-radius: 0;
-}
-
-.swg-smart-button:focus {
-  outline: none;
 }
--- a/examples/sample-pub/publisher.js
+++ b/examples/sample-pub/publisher.js
@@ -107,10 +107,15 @@ function startFlowAuto() {
 
   if (flow == 'smartbutton') {
     whenReady(function(subsciptions) {
+      var subs = subsciptions;
       whenDemoReady(function() {
         var smartButtonContainerDiv = document.querySelector('#swgId');
-        subsciptions.createSubscriptionButton(
-            smartButtonContainerDiv, {theme: 'light', lang: 'en'});
+        subs.renderSmartButton(
+            smartButtonContainerDiv,
+            {theme: 'light', lang: 'en'},
+            function() {
+              subs.showOffers({isClosable: true});
+            });
       });
     });
     return;
--- a/examples/sample-pub/publisher.js
+++ b/examples/sample-pub/publisher.js
@@ -109,8 +109,7 @@ function startFlowAuto() {
     whenReady(function(subsciptions) {
       var subs = subsciptions;
       whenDemoReady(function() {
-        var smartButton =
-            document.querySelector('.swg-smart-button');
+        var smartButton = document.querySelector('.swg-smart-button');
         subs.attachSmartButton(
             smartButton,
             {theme: 'light', lang: 'en'},
--- a/examples/sample-pub/publisher.js
+++ b/examples/sample-pub/publisher.js
@@ -109,8 +109,9 @@ function startFlowAuto() {
     whenReady(function(subsciptions) {
       var subs = subsciptions;
       whenDemoReady(function() {
-        var smartButtonContainerDiv = document.querySelector('#swgId');
-        subs.renderSmartButton(
+        var smartButtonContainerDiv =
+            document.querySelector('.swg-smart-button');
+        subs.attachSmartButton(
             smartButtonContainerDiv,
             {theme: 'light', lang: 'en'},
             function() {
--- a/examples/sample-pub/publisher.js
+++ b/examples/sample-pub/publisher.js
@@ -109,10 +109,10 @@ function startFlowAuto() {
     whenReady(function(subsciptions) {
       var subs = subsciptions;
       whenDemoReady(function() {
-        var smartButtonContainerDiv =
+        var smartButton =
             document.querySelector('.swg-smart-button');
         subs.attachSmartButton(
-            smartButtonContainerDiv,
+            smartButton,
             {theme: 'light', lang: 'en'},
             function() {
               subs.showOffers({isClosable: true});
--- a/examples/sample-pub/publisher.js
+++ b/examples/sample-pub/publisher.js
@@ -109,12 +109,13 @@ function startFlowAuto() {
     whenReady(function(subsciptions) {
       var subs = subsciptions;
       whenDemoReady(function() {
-        var smartButton = document.querySelector('button.swg-smart-button');
+        var smartButton = document.querySelector('button#mySmartButton');
         if (!smartButton) {
           console.log(
-              '***Add <button class="swg-smart-button"></button> to this page');
+              'A <button id="smartButton"> is required to render smartbox');
           return;
         }
+
         subs.attachSmartButton(
             smartButton,
             {theme: 'light', lang: 'en'},
--- a/examples/sample-pub/publisher.js
+++ b/examples/sample-pub/publisher.js
@@ -104,6 +104,18 @@ function startFlowAuto() {
     });
     return;
   }
+
+  if (flow == 'smartbutton') {
+    whenReady(function(subsciptions) {
+      whenDemoReady(function() {
+        var smartButtonContainerDiv = document.querySelector('#swgId');
+        subsciptions.createSubscriptionButton(
+            smartButtonContainerDiv, {theme: 'light', lang: 'en'});
+      });
+    });
+    return;
+  }
+
   if (flow == 'button') {
     whenReady(function(subscriptions) {
       whenDemoReady(function() {
--- a/examples/sample-pub/publisher.js
+++ b/examples/sample-pub/publisher.js
@@ -109,7 +109,7 @@ function startFlowAuto() {
     whenReady(function(subsciptions) {
       var subs = subsciptions;
       whenDemoReady(function() {
-        var smartButton = document.querySelector('.swg-smart-button');
+        var smartButton = document.querySelector('button.swg-smart-button');
         subs.attachSmartButton(
             smartButton,
             {theme: 'light', lang: 'en'},
--- a/examples/sample-pub/publisher.js
+++ b/examples/sample-pub/publisher.js
@@ -110,6 +110,11 @@ function startFlowAuto() {
       var subs = subsciptions;
       whenDemoReady(function() {
         var smartButton = document.querySelector('button.swg-smart-button');
+        if (!smartButton) {
+          console.log(
+              '***Add <button class="swg-smart-button"></button> to this page');
+          return;
+        }
         subs.attachSmartButton(
             smartButton,
             {theme: 'light', lang: 'en'},
--- a/examples/sample-pub/publisher.js
+++ b/examples/sample-pub/publisher.js
@@ -109,7 +109,7 @@ function startFlowAuto() {
     whenReady(function(subsciptions) {
       var subs = subsciptions;
       whenDemoReady(function() {
-        var smartButton = document.querySelector('button#mySmartButton');
+        var smartButton = document.querySelector('button#smartButton');
         if (!smartButton) {
           console.log(
               'A <button id="smartButton"> is required to render smartbox');
--- a/examples/sample-pub/views/article.html
+++ b/examples/sample-pub/views/article.html
@@ -110,7 +110,7 @@
           <p class="byline">By <% article.author %> on <% article.date %></p>
           <div class="hr"></div>
 
-          <button class="swg-smart-button"></button>
+          <button class="swg-smart-button" lang="fr"></button>
           <p class="text">
             Jelly jujubes marshmallow donut. Wafer cheesecake lollipop fruitcake danish powder. Pudding croissant biscuit icing. Gummi bears fruitcake cake wafer cake cookie caramels marshmallow bonbon.
           </p>
--- a/examples/sample-pub/views/article.html
+++ b/examples/sample-pub/views/article.html
@@ -110,7 +110,7 @@
           <p class="byline">By <% article.author %> on <% article.date %></p>
           <div class="hr"></div>
 
-          <div id="swgId"></div>
+          <button class="swg-smart-button"></button>
           <p class="text">
             Jelly jujubes marshmallow donut. Wafer cheesecake lollipop fruitcake danish powder. Pudding croissant biscuit icing. Gummi bears fruitcake cake wafer cake cookie caramels marshmallow bonbon.
           </p>
--- a/examples/sample-pub/views/article.html
+++ b/examples/sample-pub/views/article.html
@@ -110,7 +110,7 @@
           <p class="byline">By <% article.author %> on <% article.date %></p>
           <div class="hr"></div>
 
-          <div id="swgId" style="min-height:126px;min-width: 280px;"></div>
+          <div id="swgId"></div>
           <p class="text">
             Jelly jujubes marshmallow donut. Wafer cheesecake lollipop fruitcake danish powder. Pudding croissant biscuit icing. Gummi bears fruitcake cake wafer cake cookie caramels marshmallow bonbon.
           </p>
--- a/examples/sample-pub/views/article.html
+++ b/examples/sample-pub/views/article.html
@@ -110,7 +110,7 @@
           <p class="byline">By <% article.author %> on <% article.date %></p>
           <div class="hr"></div>
 
-          <button class="swg-smart-button" lang="fr"></button>
+          <button class="swg-smart-button"></button>
           <p class="text">
             Jelly jujubes marshmallow donut. Wafer cheesecake lollipop fruitcake danish powder. Pudding croissant biscuit icing. Gummi bears fruitcake cake wafer cake cookie caramels marshmallow bonbon.
           </p>
--- a/examples/sample-pub/views/article.html
+++ b/examples/sample-pub/views/article.html
@@ -110,7 +110,6 @@
           <p class="byline">By <% article.author %> on <% article.date %></p>
           <div class="hr"></div>
 
-          <button class="swg-smart-button"></button>
           <p class="text">
             Jelly jujubes marshmallow donut. Wafer cheesecake lollipop fruitcake danish powder. Pudding croissant biscuit icing. Gummi bears fruitcake cake wafer cake cookie caramels marshmallow bonbon.
           </p>
--- a/examples/sample-pub/views/article.html
+++ b/examples/sample-pub/views/article.html
@@ -110,6 +110,7 @@
           <p class="byline">By <% article.author %> on <% article.date %></p>
           <div class="hr"></div>
 
+          <div id="swgId" style="display:none;height:10px;min-width: 20px;background-color: #ff1"></div>
           <p class="text">
             Jelly jujubes marshmallow donut. Wafer cheesecake lollipop fruitcake danish powder. Pudding croissant biscuit icing. Gummi bears fruitcake cake wafer cake cookie caramels marshmallow bonbon.
           </p>
--- a/examples/sample-pub/views/article.html
+++ b/examples/sample-pub/views/article.html
@@ -110,7 +110,7 @@
           <p class="byline">By <% article.author %> on <% article.date %></p>
           <div class="hr"></div>
 
-          <div id="swgId" style="display:none;height:10px;min-width: 20px;background-color: #ff1"></div>
+          <div id="swgId" style="min-height:126px;min-width: 280px;"></div>
           <p class="text">
             Jelly jujubes marshmallow donut. Wafer cheesecake lollipop fruitcake danish powder. Pudding croissant biscuit icing. Gummi bears fruitcake cake wafer cake cookie caramels marshmallow bonbon.
           </p>
--- a/src/api/subscriptions.js
+++ b/src/api/subscriptions.js
@@ -242,9 +242,10 @@ export class Subscriptions {
    *
    * @param {!Element} container
    * @param {!ButtonOptions=} opt_options
-   * @return {!Element}
+   * @param {function()=} opt_callback
+   * @return {!Promise}
    */
-  createSubscriptionButton(container, opt_options) {}
+  renderSmartButton(container, opt_options, opt_callback) {}
 
   /**
    * Retrieves the propensity module that provides APIs to
@@ -387,7 +388,6 @@ export let SaveSubscriptionRequestCallback;
  * @typedef {{
  *   theme: string,
  *   lang: (string|undefined),
- *   isClosable: (boolean|undefined),
  * }}
  */
 export let ButtonOptions;
--- a/src/api/subscriptions.js
+++ b/src/api/subscriptions.js
@@ -243,7 +243,6 @@ export class Subscriptions {
    * @param {!Element} container
    * @param {!ButtonOptions=} opt_options
    * @param {function()=} opt_callback
-   * @return {!Promise}
    */
   renderSmartButton(container, opt_options, opt_callback) {}
 
@@ -268,7 +267,6 @@ export const SubscriptionFlows = {
   LINK_ACCOUNT: 'linkAccount',
   SHOW_LOGIN_PROMPT: 'showLoginPrompt',
   SHOW_LOGIN_NOTIFICATION: 'showLoginNotification',
-  SMART_SUBSCRIPTION_BUTTON: 'showSmartSubscriptionButton',
 };
 
 
--- a/src/api/subscriptions.js
+++ b/src/api/subscriptions.js
@@ -237,7 +237,7 @@ export class Subscriptions {
   attachButton(button, optionsOrCallback, opt_callback) {}
 
   /**
-   * Creates smartButton element and the provided callback.
+   * Attaches smartButton element and the provided callback.
    * The default theme is "light".
    *
    * @param {!Element} container
--- a/src/api/subscriptions.js
+++ b/src/api/subscriptions.js
@@ -241,10 +241,10 @@ export class Subscriptions {
    * The default theme is "light".
    *
    * @param {!Element} container
-   * @param {!ButtonOptions=} opt_options
+   * @param {!ButtonOptions|function()} optionsOrCallback
    * @param {function()=} opt_callback
    */
-  renderSmartButton(container, opt_options, opt_callback) {}
+  attachSmartButton(container, optionsOrCallback, opt_callback) {}
 
   /**
    * Retrieves the propensity module that provides APIs to
--- a/src/api/subscriptions.js
+++ b/src/api/subscriptions.js
@@ -237,6 +237,16 @@ export class Subscriptions {
   attachButton(button, optionsOrCallback, opt_callback) {}
 
   /**
+   * Creates smartButton element and the provided callback.
+   * The default theme is "light".
+   *
+   * @param {!Element} container
+   * @param {!ButtonOptions=} opt_options
+   * @return {!Element}
+   */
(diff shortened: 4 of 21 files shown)


examples/sample-pub/views/article.html
```diff
@@ -110,6 +110,7 @@ <h1 class="heading"><% article.title %></h1>
           <p class="byline">By <% article.author %> on <% article.date %></p>
           <div class="hr"></div>
 
+          <div id="swgId" style="display:none;height:10px;min-width: 20px;background-color: #ff1"></div>
```
dparikh: Publisher would require to specify the container `DIV`, where `smartbutton` would be injected by the API call. 

Need to finalize, what `css` `style` attributes are allowed/accepted by SwG.

src/api/subscriptions.js
```diff
@@ -375,6 +386,8 @@ export let SaveSubscriptionRequestCallback;
  *
  * @typedef {{
  *   theme: string,
+ *   lang: (string|undefined),
+ *   isClosable: (boolean|undefined),
```
dparikh: `isClosable` will be passed to the `showOffers` API call when `dialog` is opened.

src/components/dialog-manager.js
```diff
@@ -72,17 +72,19 @@ export class DialogManager {
   /**
    * @param {!./view.View} view
    * @param {boolean=} hidden
+   * @param {?Element=} elem
    * @return {!Promise}
    */
-  openView(view, hidden = false) {
+  openView(view, hidden = false, elem) {
```
dparikh: For Testing purpose only.

Dialog will not be used for `smartbutton`

src/components/dialog.js
```diff
@@ -273,14 +273,20 @@ export class Dialog {
   /**
    * Opens the given view and removes existing view from the DOM if any.
    * @param {!./view.View} view
+   * @param {?Element=} elem
    * @return {!Promise}
    */
-  openView(view) {
+  openView(view, elem) {
```
dparikh: For testing purpose only.

src/runtime/runtime.js
```diff
@@ -816,6 +824,31 @@ export class ConfiguredRuntime {
...
+    const opts = /** @type {!ButtonOptions} */
+        (Object.assign(options || {}, {theme: Theme.LIGHT}));
+    if (!isExperimentOn(this.win_, ExperimentFlags.SMARTBOX)) {
+      throw new Error('Not yet launched!');
+    }
+    if (!container) {
+      throw new Error('No element found to render Smart button!');
+    }
+    return this.documentParsed_.then(() => {
+      return this.getEntitlements().then(entitlements => {
```
dparikh: `getEntitlements` API call is required to find if there are `subscriptions` and if any, do not display `smartbutton.
Also, `isReadyToPay` flag is passed to the `SmartSubscriptionButtonFlow` to decide what `message` is rendered underneath the button.

src/api/subscriptions.js
```diff
@@ -236,6 +236,16 @@ export class Subscriptions {
...
 
+  /**
+   * Creates smartButton element and the provided callback.
+   * The default theme is "light".
+   *
+   * @param {!Element} container
+   * @param {!ButtonOptions=} opt_options
+   * @return {!Element}
+   */
+  createSubscriptionButton(container, opt_options) {}
```
dparikh: rename to `renderSmartButton(container, options, callback)`

src/api/subscriptions.js
```diff
@@ -236,6 +236,17 @@ export class Subscriptions {
...
+  /**
+   * Creates smartButton element and the provided callback.
+   * The default theme is "light".
+   *
+   * @param {!Element} container
+   * @param {!ButtonOptions=} opt_options
+   * @param {function()=} opt_callback
+   * @return {!Promise}
+   */
+  renderSmartButton(container, opt_options, opt_callback) {}
```
dvoytenko: Let's also discuss the naming of `attachSmartButton`. The example above is pretty similar from many points of view. We might have a more obvious API if we move in that path.

src/api/subscriptions.js
```diff
@@ -257,6 +268,7 @@ export const SubscriptionFlows = {
   LINK_ACCOUNT: 'linkAccount',
   SHOW_LOGIN_PROMPT: 'showLoginPrompt',
   SHOW_LOGIN_NOTIFICATION: 'showLoginNotification',
+  SMART_SUBSCRIPTION_BUTTON: 'showSmartSubscriptionButton',
```
dvoytenko: I'm not convinced we need this.

src/components/view.js
```diff
@@ -41,6 +41,13 @@ export class View {
    */
   init(unusedDialog) {}
 
+
```
dvoytenko: I'd hope no changes to the `View` interface would be needed.

examples/sample-pub/views/article.html
```diff
@@ -110,6 +110,7 @@ <h1 class="heading"><% article.title %></h1>
           <p class="byline">By <% article.author %> on <% article.date %></p>
           <div class="hr"></div>
 
+          <div id="swgId" style="min-height:126px;min-width: 280px;"></div>
```
dparikh: To revert this change before merge.

src/runtime/button-api.js
```diff
@@ -17,6 +17,14 @@
 import {createElement} from '../utils/dom';
 import {msg} from '../utils/i18n';
 
+/**
+ * @enum {string}
+ */
+export const Theme = {
+  'LIGHT': 'light',
```
dvoytenko: Enum names should not have quotes.

src/runtime/runtime.js
```diff
@@ -816,6 +824,23 @@ export class ConfiguredRuntime {
     this.buttonApi_.attach(button, optionsOrCallback, opt_callback);
   }
 
+  /** @override */
+  renderSmartButton(container, options, callback) {
+    if (!options) {
+      Object.assign({}, {theme: Theme.LIGHT, lang: 'en'});
```
dvoytenko: Hmm. I don't think this is doing anything. Plus the `if (!options)` is also wrong I believe. Per method's signature, the you should be expecting one one of these shapes: 

```
renderSmartButton(container, callback)
```
or
```
renderSmartButton(container, options, callback)
```

Thus, the second argument is not guaranteed to be an options map. And it's never not specified.

src/api/subscriptions.js
```diff
@@ -236,6 +236,17 @@ export class Subscriptions {
...
   attachButton(button, optionsOrCallback, opt_callback) {}
 
+  /**
+   * Creates smartButton element and the provided callback.
+   * The default theme is "light".
+   *
+   * @param {!Element} container
+   * @param {!ButtonOptions=} opt_options
+   * @param {function()=} opt_callback
+   * @return {!Promise}
```
dvoytenko: What does this "promise" mean at this time?

src/runtime/runtime.js
```diff
@@ -816,6 +824,23 @@ export class ConfiguredRuntime {
...
+    }
+    if (!isExperimentOn(this.win_, ExperimentFlags.SMARTBOX)) {
+      throw new Error('Not yet launched!');
+    }
+    if (!container) {
+      throw new Error('No element found to render Smart button!');
+    }
+    return this.documentParsed_.then(() => {
+      return new SmartSubscriptionButtonApi(this, container, options, callback)
```
dvoytenko: Is this a good idea to return in an API?

src/runtime/smart-button-api.js
```diff
@@ -0,0 +1,113 @@
...
+ * distributed under the License is distributed on an "AS-IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+
+import {ActivityIframeView} from '../ui/activity-iframe-view';
+import {Theme} from './button-api';
+//import {SubscriptionFlows} from '../api/subscriptions';
```
dvoytenko: remove

src/runtime/smart-button-api.js
```diff
@@ -0,0 +1,113 @@
...
+    this.container_ = container;
+
+    /** @private const {?function()=} */
+    this.callback_ = callback;
+
+    /** @private @const {string} */
+    this.theme_ = options && options.theme || Theme.LIGHT;
+
+    /** @private {boolean} */
+    this.isClosable_ = options && options.isClosable || false;
```
dvoytenko: Remove?

src/runtime/smart-button-api.js
```diff
@@ -0,0 +1,113 @@
...
+    if (this.isClosable_ == undefined) {
+      this.isClosable_ = false;  // Default is to hide Close button.
+    }
+
+    /** @private @const {!ActivityIframeView} */
+    this.activityIframeView_ = new ActivityIframeView(
```
dvoytenko: Can we not have it to be a view right away?

src/runtime/smart-button-api.js
```diff
@@ -0,0 +1,113 @@
...
+          'productId': deps.pageConfig().getProductId(),
+          'publicationId': deps.pageConfig().getPublicationId(),
+          'theme': this.theme_,
+          'lang': options && options.lang || 'en',
+        }),
+        /* shouldFadeBody */ false);
+  }
+
+  /**
+   * Starts the smart button subscription button flow.
```
dvoytenko: Nit: not a flow.

src/runtime/smart-button-api.js
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  /**
+   * @private
+   */
+  buildContent_() {
+    this.container_.appendChild(this.activityIframeView_.getElement());
+
+    setImportantStyles(this.container_, {
+      'height': '126px',
```
dvoytenko: This seems odd to set it on the container. Why not just set it on the element itself? We are doing it early enough so that the developer can reflect the changes early in the development cycle.

src/runtime/smart-button-api.js
```diff
@@ -0,0 +1,113 @@
...
+    });
+
+    this.activityIframeView_.initContainer().then(() => {
+      setImportantStyles(this.activityIframeView_.getElement(), {
+        'opacity': 1,
+        'height': '100%',
```
dvoytenko: Ditto: this is rather dengerous.

src/runtime/button-api.js
dparikh: Done

src/api/subscriptions.js
dparikh: This was left over from previous implementation. Was not being used anywhere, removed.

src/runtime/smart-button-api.js
dparikh: Removed

src/runtime/smart-button-api.js
dparikh: Done

src/runtime/smart-button-api.js
dparikh: Removed

src/api/subscriptions.js
dparikh: Not required, removed.

src/runtime/runtime.js
dparikh: Since now it is an API, return is not required.  Removed.

src/runtime/smart-button-api.js
dparikh: Removed from here. 
We might still want to override any styles, publishers might have set, to ensure that the height is at least (`min-height: 126px;` )

src/components/view.js
dparikh: `view` interface has signature, `init(unusedDialog)` and for `smartButton` we do not have `dialog` param to pass.

src/runtime/smart-button-api.js
dparikh: Removed.

src/components/view.js
dvoytenko: `View` is still defined as a "view" inside a dialog. So using it w/o a dialog does not seem to be like a right use.

src/runtime/smart-button-api.js
dvoytenko: Setting it on the element itself is totally ok, i believe. Our main question - how early are we doing this. Here . you do it pretty early, so there's no reason to just set it on your button element.

dvoytenko: Still a lot of conflation between a container and a button concepts. I think we should review the `attachSmartButton` path - it could clarify these things. Did we ever try to place our `<iframe>` inside a `<button>`?

src/runtime/runtime.js
dparikh: Changed the supported signature for `renderSmartButton`

src/runtime/smart-button-api.js
dparikh: Moved out from constructor.

src/components/view.js
dparikh: OK. Reverting this change.
Making direct call to `activityPort.openIframe` now.

src/runtime/smart-button-api.js
dparikh: Will discuss offline...

src/api/subscriptions.js
dparikh: Done as discussed offline

examples/sample-pub/views/article.html
```diff
@@ -110,6 +110,7 @@ <h1 class="heading"><% article.title %></h1>
           <p class="byline">By <% article.author %> on <% article.date %></p>
           <div class="hr"></div>
 
+          <button class="swg-smart-button"></button>
```
dparikh: Remove before merge.

assets/swg-button.css
```diff
@@ -257,3 +257,17 @@
 .swg-button-dark:lang(zh-tw):after {
   background-image: url("i18n/b-zh-tw-dk.svg");
 }
+
+.swg-smart-button {
```
dvoytenko: Please add a comment here that all "button" styles will be defined inside the iframe.

assets/swg-button.css
```diff
@@ -257,3 +257,17 @@
...
+  border: none;
+  padding: 0;
+  background: transparent;
+  min-height: 126px;
+  position: relative;
+  width: 300px;
```
dvoytenko: Let's also define a min-width.

assets/swg-button.css
```diff
@@ -257,3 +257,17 @@
 .swg-button-dark:lang(zh-tw):after {
   background-image: url("i18n/b-zh-tw-dk.svg");
 }
+
+.swg-smart-button {
+  border: none;
+  padding: 0;
+  background: transparent;
+  min-height: 126px;
+  position: relative;
```
dvoytenko: Should we also do `overflow: hidden`?

assets/swg-button.css
```diff
@@ -257,3 +257,17 @@
 .swg-button-dark:lang(zh-tw):after {
   background-image: url("i18n/b-zh-tw-dk.svg");
 }
+
```
dvoytenko: Please move it up before all the language re-definitions.

examples/sample-pub/publisher.js
```diff
@@ -104,6 +104,24 @@ function startFlowAuto() {
     });
     return;
   }
+
+  if (flow == 'smartbutton') {
+    whenReady(function(subsciptions) {
+      var subs = subsciptions;
+      whenDemoReady(function() {
+        var smartButtonContainerDiv =
```
dvoytenko: It's no longer a container, so let's also remove "container" from the name. And also "div".

src/api/subscriptions.js
```diff
@@ -236,6 +236,16 @@ export class Subscriptions {
    */
   attachButton(button, optionsOrCallback, opt_callback) {}
 
+  /**
+   * Creates smartButton element and the provided callback.
```
dvoytenko: s/Creates/Attaches

src/runtime/button-api.js
```diff
@@ -98,20 +100,17 @@ export class ButtonApi {
...
-    const options =
-        typeof optionsOrCallback != 'function' ?
-        optionsOrCallback : null;
-    const callback = /** @type {function()} */ (
-        (typeof optionsOrCallback == 'function' ? optionsOrCallback : null) ||
-            opt_callback);
+    const options = this.getOptions_(optionsOrCallback);
+    const callback = /** @type {function(Event):boolean} */
+        (this.getCallback_(optionsOrCallback, opt_callback));
     let theme = options && options['theme'];
```
dvoytenko: So it looks like now options are never null/undefined here?

src/runtime/button-api.js
```diff
@@ -98,20 +100,17 @@ export class ButtonApi {
...
    */
   attach(button, optionsOrCallback, opt_callback) {
-    const options =
-        typeof optionsOrCallback != 'function' ?
-        optionsOrCallback : null;
-    const callback = /** @type {function()} */ (
-        (typeof optionsOrCallback == 'function' ? optionsOrCallback : null) ||
-            opt_callback);
+    const options = this.getOptions_(optionsOrCallback);
+    const callback = /** @type {function(Event):boolean} */
```
dvoytenko: Why does it return boolean? Why is typecast needed here if we have a function to return this function to us?

src/runtime/button-api.js
```diff
@@ -122,4 +121,51 @@ export class ButtonApi {
...
+
+  /**
+   *
+   * @param {?../api/subscriptions.ButtonOptions|function():null} optionsOrCallback
+   * @param {function():null=} opt_callback
+   * @return {function():null|function(Event):boolean}
+   * @private
+   */
+  getCallback_(optionsOrCallback, opt_callback) {
+    const callback = /** @type {function():null|function(Event):boolean} */ (
```
dvoytenko: Why the variance with return type? Also return type of `null` is not a clear concept.

src/runtime/button-api.js
```diff
@@ -122,4 +121,51 @@ export class ButtonApi {
...
+   * @param {!Element} button
+   * @param {!../api/subscriptions.ButtonOptions|function()} optionsOrCallback
+   * @param {function():null=} opt_callback
+   * @return {!Element}
+   */
+  attachSmartButton(deps, button, optionsOrCallback, opt_callback) {
+    const options = this.getOptions_(optionsOrCallback);
+    const callback = /** @type {function():null} */
+        (this.getCallback_(optionsOrCallback, opt_callback));
+    let theme = options && options['theme'];
```
dvoytenko: The `theme` never used?

src/runtime/button-api.js
```diff
@@ -98,20 +100,17 @@ export class ButtonApi {
...
-    const callback = /** @type {function()} */ (
-        (typeof optionsOrCallback == 'function' ? optionsOrCallback : null) ||
-            opt_callback);
+    const options = this.getOptions_(optionsOrCallback);
+    const callback = /** @type {function(Event):boolean} */
+        (this.getCallback_(optionsOrCallback, opt_callback));
     let theme = options && options['theme'];
-    if (theme !== 'light' && theme !== 'dark') {
-      theme = 'light';
+    if (theme !== Theme.LIGHT && theme !== Theme.DARK) {
```
dvoytenko: I think we should move this code to `getOptions_`

src/runtime/runtime.js
```diff
@@ -420,6 +420,13 @@ export class Runtime {
     return this.buttonApi_.create(optionsOrCallback, opt_callback);
   }
 
+  /** @override */
+  attachSmartButton(container, opt_options, callback) {
```
dvoytenko: Use `optionsOrCallback, opt_callback` naming here as well.

src/runtime/runtime.js
```diff
@@ -816,6 +823,20 @@ export class ConfiguredRuntime {
     this.buttonApi_.attach(button, optionsOrCallback, opt_callback);
   }
 
+  /** @override */
+  attachSmartButton(container, optionsOrCallback, opt_callback) {
```
dvoytenko: s/container/button

src/runtime/runtime.js
```diff
@@ -816,6 +823,20 @@ export class ConfiguredRuntime {
...
+    if (!isExperimentOn(this.win_, ExperimentFlags.SMARTBOX)) {
+      throw new Error('Not yet launched!');
+    }
+    if (!container) {
```
dvoytenko: not needed.

src/runtime/runtime.js
```diff
@@ -816,6 +823,20 @@ export class ConfiguredRuntime {
...
+      throw new Error('No container element found to render Smart button!');
+    }
+    this.documentParsed_.then(() => {
```
dvoytenko: I'm actually not sure why we need to wait for `documentParsed` here. Can you explain?

src/runtime/smart-button-api.js
```diff
@@ -0,0 +1,123 @@
...
+};
+
+
+/**
+ * The class for Smart button Api.
+ */
+export class SmartSubscriptionButtonApi {
+  /**
+   * @param {!./deps.DepsDef} deps
+   * @param {!Element} container
```
dvoytenko: s/container/button

src/runtime/smart-button-api.js
```diff
@@ -0,0 +1,123 @@
...
+      'publicationId': this.deps_.pageConfig().getPublicationId(),
+      'theme': options && options.theme || 'light',
+      'lang': this.options_ && this.options_.lang || 'en',
+    });
+  }
+
+  /**
+   * @return {!HTMLIFrameElement}
+   */
+  getElement() {
```
dvoytenko: Do we need this method? If we do, imo it should return `button`, not iframe.

src/runtime/smart-button-api.js
```diff
@@ -0,0 +1,123 @@
...
+        createElement(this.doc_, 'iframe', iframeAttributes));
+
+    /** @private @const {!Element} */
+    this.container_ = container;
+
+    /** @private {!../api/subscriptions.ButtonOptions|undefined} */
+    this.options_ = options;
+
+    /** @private const {?function()=} */
+    this.callback_ = callback;
```
dvoytenko: Should we add a callback on the button itself right away as well to handle the keyboard tab/enter case?

assets/swg-button.css
dparikh: Yes, Added.

assets/swg-button.css
dparikh: Done

assets/swg-button.css
dparikh: Done

assets/swg-button.css
dparikh: Moved up.

examples/sample-pub/publisher.js
dparikh: Done

src/runtime/runtime.js
dparikh: Was just extra check. Removed.

src/runtime/runtime.js
dparikh: That was earlier implementation, when this was defined as a `flow`. Removed.

src/runtime/smart-button-api.js
dparikh: Done

src/runtime/smart-button-api.js
dparikh: Not really required, since used only within the class. Removed.

src/runtime/smart-button-api.js
dparikh: Added. Though, we need a way to `focus` the `iframe->button` when user `tabs` to the outer `button`.

src/api/subscriptions.js
dparikh: Done

src/runtime/runtime.js
dparikh: Done.

src/runtime/button-api.js
dparikh: This is due to that fact that the returned `callback` function is passed as a param to `addEventListener` and if not typcasted, throws following error:

```
src/runtime/button-api.js:120: ERROR - Invalid type for parameter 2 of function button.addEventListener.
Expected : EventListener|function(Event):boolean|undefined|null
Found    : function(Event|null):boolean|null

                                     ^^^^^^^^
1 error(s), 0 warning(s)```

Same function is used for `attachSmartButton` but not as an argument to `addEventListener` so needed this typecase.

File: src/runtime/button-api.js (hunk 30 above)
Comment (dparikh) [reply src/runtime/button-api.js]: This is to support two different function `attach` (having `addEventListener`) and 'attachSmartButton` (callback).
Without this, linter fails.

File: src/runtime/button-api.js (hunk 32 above)
Comment (dparikh) [reply src/runtime/button-api.js]: Re-factored this code to `getOptions_`

File: src/runtime/runtime.js (hunk 34 above)
Comment (dparikh) [reply src/runtime/runtime.js]: Done

File: src/runtime/button-api.js (hunk 31 above)
Comment (dparikh) [reply src/runtime/button-api.js]: Re-factored this, moved to `getOptions_`

File: src/runtime/button-api.js (hunk 28 above)
Comment (dparikh) [reply src/runtime/button-api.js]: Yes, updated.

Comment (dparikh) [reply src/runtime/button-api.js]: Yes. Remove that code updated for `options[lang]` as well.

File: examples/sample-pub/publisher.js [hunk 40]
```diff
@@ -104,6 +104,23 @@ function startFlowAuto() {
     });
     return;
   }
+
+  if (flow == 'smartbutton') {
+    whenReady(function(subsciptions) {
+      var subs = subsciptions;
+      whenDemoReady(function() {
+        var smartButton = document.querySelector('button.swg-smart-button');
```
dparikh: Added stricts CSS selector to ensure that publisher is using `button` component.

src/runtime/smart-button-api.js
```diff
@@ -0,0 +1,122 @@
...
+  /**
+   * Make a call to build button content and listens for the 'click' message.
+   * @return {!Element}
+   */
+  start() {
+    /**
+     * Add a callback to the button itself to fire the iframe click action when
+     * user tabs to the button and hits enter.
+     */
+    this.button_.addEventListener('click', this.callback_);
```
dvoytenko: To come back to this point. It'd be better to indirect/wrap this callback than forcing the types all the way up. E.g.

```
const callback = this.callback;
button.addEventListener('click', () => {
  callback();
});
```

src/runtime/smart-button-api.js
```diff
@@ -0,0 +1,122 @@
...
+    /** @private @const {!web-activities/activity-ports.ActivityPorts} */
+    this.activityPorts_ = deps.activities();
+
+    /** @private @const {!HTMLIFrameElement} */
+    this.iframe_ =
+    /** @type {!HTMLIFrameElement} */ (
+        createElement(this.doc_, 'iframe', iframeAttributes));
+
+    /** @private @const {!Element} */
+    this.button_ = button;
```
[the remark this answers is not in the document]
dparikh: `assets/swg-button.css`

src/api/subscriptions.js
```diff
@@ -236,6 +236,16 @@ export class Subscriptions {
    */
   attachButton(button, optionsOrCallback, opt_callback) {}
 
+  /**
+   * Attaches smartButton element and the provided callback.
+   * The default theme is "light".
+   *
+   * @param {!Element} container
```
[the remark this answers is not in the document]
dparikh: Done.

src/runtime/button-api.js
```diff
@@ -98,28 +100,74 @@ export class ButtonApi {
...
+   * @param {!../api/subscriptions.ButtonOptions|function()} optionsOrCallback
+   * @return {!../api/subscriptions.ButtonOptions}
+   * @private
+   */
+  getOptions_(optionsOrCallback) {
+    const options = /** @type {!../api/subscriptions.ButtonOptions} */
+        (typeof optionsOrCallback != 'function' ?
+        optionsOrCallback : {'theme': Theme.LIGHT});
+
+    const theme = options && options['theme'];
```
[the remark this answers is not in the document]
dparikh: Yes. Removed additional check.

src/runtime/button-api.js
dparikh: removed now.

src/runtime/smart-button-api.js
```diff
@@ -0,0 +1,122 @@
...
+
+    /** @private {!../api/subscriptions.ButtonOptions|undefined} */
```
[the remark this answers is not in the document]
dparikh: Removed.

src/runtime/smart-button-api.js
dparikh: Done

dparikh: done

examples/sample-pub/views/article.html
dparikh: Done

src/runtime/button-api.js
dparikh: Done.

src/runtime/button-api.js
```diff
@@ -98,28 +100,75 @@ export class ButtonApi {
...
+   * @param {function()=} opt_callback
+   * @return {!Element}
+   */
+  attachSmartButton(deps, button, optionsOrCallback, opt_callback) {
+    const options = this.getOptions_(optionsOrCallback);
+    const callback = /** @type {function()} */
+        (this.getCallback_(optionsOrCallback, opt_callback));
+
+    // Add required CSS class, if missing.
+    if (!button.classList.contains('swg-smart-button')) {
```
dvoytenko: No need for `if`. It's implemented internally by `classList.add(...)`

src/runtime/smart-button-api.js
```diff
@@ -0,0 +1,124 @@
...
+    this.callback_ = callback;
+
+    /** @private @const {string} */
+    this.src_ = feUrl('/smartboxiframe');
+
+    /** @private @const {!Object} */
+    this.args_ = feArgs({
+      'productId': this.deps_.pageConfig().getProductId(),
+      'publicationId': this.deps_.pageConfig().getPublicationId(),
+      'theme': options && options.theme || 'light',
```
dvoytenko: Some `options` are resolved via arg and some via `this.options_`. Please make this consistent.

src/runtime/button-api.js
dparikh: Removed.

src/runtime/smart-button-api.js
dparikh: Done.

PR #535, opened 2019-04-30
Propensity API update and experiment changes

sohanirao:
- sendEvent() API now requires a PropensityEvent type as documented
- getPropensityModule() now provides the Propensity Module without experiment enabled
- getPropensity() API is now hidden behind an experiment.

### DIFF
--- a/src/runtime/propensity.js
+++ b/src/runtime/propensity.js
@@ -15,6 +15,10 @@
  */
 import * as PropensityApi from '../api/propensity-api';
 import {PropensityServer} from './propensity-server';
+import {
+  isExperimentOn,
+} from './experiments';
+import {ExperimentFlags} from './experiment-flags';
 
 /**
  * @implements {PropensityApi.PropensityApi}
@@ -50,6 +54,9 @@ export class Propensity {
 
   /** @override */
   getPropensity(type) {
+    if (!isExperimentOn(this.win_, ExperimentFlags.PROPENSITY)) {
+      throw new Error('Not yet launched!');
+    }
     if (type && !Object.values(PropensityApi.PropensityType).includes(type)) {
       throw new Error('Invalid propensity type requested');
     }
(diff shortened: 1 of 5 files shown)


src/runtime/propensity.js
```diff
@@ -61,13 +68,13 @@ export class Propensity {
...
+    if (!Object.values(PropensityApi.Event).includes(userEvent.name)) {
       throw new Error('Invalid user event provided');
     }
-    // TODO(sohanirao): drop the params for some events?
-    // TODO(sohanirao) : verify parameters for some events
-    const paramString = jsonParams && JSON.stringify(jsonParams);
-    this.propensityServer_.sendEvent(userEvent, paramString);
+    // TODO(sohanirao, mborof): Idenfity the new interface with event
+    // manager and update the lines below to adhere to that interface
+    const paramString = userEvent.data && JSON.stringify(userEvent.data);
```
jpettitt: If `userEvent.data` is not an object but is present you could end up sending something that isn't json - is that your intent?  eg data: "foobar" will send a string not a depth 1 json object.

sohanirao: This is a good question. I guess by not checking it's a string, I have left the API loose. Ideally, it should be a JSON object as guided by the API documentation. Not sure if should check the type of the event.data field and ensure its a JSON object. Although, that was the idea here, the publisher should provide a key indicating what the value corresponds to.

IanUllery: Hi, my name is Ian. I know this project is probably none of my business, but I'm learning to code and would love to know what you guys are doing. Could you please clarify?

Also, I thought you should know that this branch is out-of-date with the master branch.

jpettitt: I think if you are expecting an object you should enforce that and throw if it isn't one.

@IanUllery re out of date, no problem, if needed we'll merge master back to this branch.   As to what we're doing.  We're defining an API for a module that predicts how likely a user is to subscribe to a publication.  API definitions are really important as they represent a contract between two different codebases.  With that in mind we're being picky about getting it right because it's hard to change later.  In this instance the comment was about argument validation - we expect the `userEvent.data` to be an object we serialize to JSON, I was suggesting that we validate that it actually is what we expect.  Defensive programming.

IanUllery: Thanks for the reply. Why is this code hard to change any more than other projects?

jpettitt: Once deployed this code will live on many 3rd party publisher sites (swg.js in integrated in amp-subscriptions-google and is deployed on major news sites across the world).  Once developers deploy against an API any change we make will require them to change their code, if we get it right now we can avoid a lot of work for others later.

sohanirao: I have addressed the concern of validating argument to be json format where expected. PTAL

src/runtime/propensity.js
```diff
@@ -61,13 +68,19 @@ export class Propensity {
...
+  sendEvent(userEvent) {
+    if (!Object.values(PropensityApi.Event).includes(userEvent.name)) {
       throw new Error('Invalid user event provided');
     }
-    // TODO(sohanirao): drop the params for some events?
-    // TODO(sohanirao) : verify parameters for some events
-    const paramString = jsonParams && JSON.stringify(jsonParams);
-    this.propensityServer_.sendEvent(userEvent, paramString);
+    if (userEvent.data && !isObject(userEvent.data)) {
+      throw new Error('Event param should be a JSON');
```
jpettitt: nit: change "should be a JSON" to "must be an Object"

PR #543, opened 2019-05-09
Event manager

mborof:
Changes to proto plus the event manager and its test code.  Removes a build rule forbidding the word AMP in our codebase.


src/proto/api_messages_test.js
```diff
@@ -15,90 +15,110 @@
...
-        return false;
-      }
-      for (let j = 0; j < arr.length; j++) {
-        if (arr[j] != otherArr[j]) {
-          return false;
-        }
-      }
-    } else {
-      if (thisArray[i] != otherArray[i]) {
+if (!otherArray || !thisArray) {
```
sohanirao: indented unnecessarily

sohanirao:
Too many indents are making it hard to review this code. Can you please remove unnecessary indents and upload again?

src/proto/api_messages_test.js
mborof: I seemed to see not enough indenting.  I've fixed that

src/proto/api_messages_test.js
```diff
@@ -15,17 +15,19 @@
...
+  EventOriginator,
+  deserialize} from './api_messages';
 
 /**
- * Compare two protos
- * @param {!Array} thisArray
- * @param {!Array} otherArray
- * @return {boolean}
- */
+* Compare two protos
```
sohanirao: unnecessary indent

src/proto/api_messages_test.js
```diff
@@ -15,17 +15,19 @@
  */
 
 import {
-    AnalyticsContext,
-    AnalyticsRequest,
-    AnalyticsEvent,
-    deserialize} from './api_messages';
+  AnalyticsContext,
```
sohanirao: unnecessary indent

src/proto/api_messages_test.js
```diff
@@ -93,6 +107,10 @@ describe('api_messages', () => {
       analyticscontext.setLabelList([]);
       analyticsrequest.setContext(analyticscontext);
       analyticsrequest.setEvent(AnalyticsEvent.UNKNOWN);
+      const /** !AnalyticsEventMeta  */ analyticseventmeta = new AnalyticsEventMeta();
+      analyticseventmeta.setEventOriginator(EventOriginator.UNKNOWN_CLIENT);
```
sohanirao: is the indent right for this? looks wierd here but may be different in your IDE etc

src/runtime/swg-client-event-manager.js
```diff
@@ -0,0 +1,312 @@
...
+  addFilterer(callback) {
+    assert(SwgClientEventManager.isFunction_(callback),
+        'Invalid callback in SwgClientEventManager.addFilterer');
+    this.filterers_.push(callback);
+  }
+
+  /**Call this function to inform all listeners that a client event has
+   * occurred.
+   * @param {!SwgClientEvent} event
+   * @public
```
sohanirao: no need to say @public. It's implied.

src/runtime/swg-client-event-manager.js
```diff
@@ -0,0 +1,312 @@
...
+   * Throws an error if it isn't.
+   * @param {!Object} enumType
+   * @param {*} enumValue
+   * @private
+   */
+  static isValidEnumValue_(enumType, enumValue) {
+    return Object.values(enumType).includes(enumValue);
+  }
+
+  /**Returns true if the passed value is a JSON object
```
sohanirao: There's already an isObject as a utility in this code base

src/runtime/swg-client-event-manager.js
```diff
@@ -0,0 +1,312 @@
...
+ * @typedef {Array<CallBack>} CallBackArray
+ */
+export let CallBackArray;
+
+export const ShouldFilter = {
+  CONTINUE_EXECUTING: 0,
+  STOP_EXECUTING: 1,
+};
+
+export class SwgClientEvent {
```
sohanirao: Since this object gets passed around a lot, I'd prefer to have a seperate helper class and create SwgClientEvent as a typedef
{
  type: <enum from api_messages.proto>
  isFromUserAction: <boolean>
  params: <Object>
}

src/runtime/swg-client-event-manager.js
```diff
@@ -0,0 +1,312 @@
...
+    }
+    return true;
+  }
+}
+
+export class SwgClientEventManager {
+
+  /**True if the value is a function
+   * @param {*} value
+   * @private
```
sohanirao: Since this is not specific to this class,maybe add this if not already present into utils

src/runtime/swg-client-event-manager.js
```diff
@@ -0,0 +1,312 @@
...
+ */
+export let SwgEventManagerFilterer;
+
+/** An array of callback functions
+ * @typedef {SwgEventManagerListener|SwgEventManagerFilterer} CallBack
+ */
+export let CallBack;
+
+/** An array of callback functions
+ * @typedef {Array<CallBack>} CallBackArray
```
sohanirao: Can we remove this. Array<!Callback> callbacks is good enough

src/runtime/swg-client-event-manager.js
```diff
@@ -0,0 +1,312 @@
...
+/**
+ * Return false to cancel the event.  You can also modify the event object
+ * which is passed by reference.
+ * @callback SwgEventManagerFilterer
+ * @param {SwgClientEvent} event
+ * @return {ShouldFilter}
+ */
+export let SwgEventManagerFilterer;
+
+/** An array of callback functions
```
sohanirao: typo?

src/runtime/swg-client-event-manager.js
```diff
@@ -0,0 +1,312 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {AnalyticsEvent,EventOriginator} from '../proto/api_messages';
+
+/**
+ * The callback used to notify registered listeners about new client events.
+ * @callback SwgEventManagerListener
```
sohanirao: can we use a definition like we do in rest of the code for a callback

{function(!SwgClientEvent)} SwgEventListener

Also do we need to call it SwgEventManagerListener? SwgEventListener not enough?

src/runtime/swg-client-event-manager.js
```diff
@@ -0,0 +1,312 @@
...
+ * @typedef {SwgEventManagerListener|SwgEventManagerFilterer} CallBack
+ */
+export let CallBack;
+
+/** An array of callback functions
+ * @typedef {Array<CallBack>} CallBackArray
+ */
+export let CallBackArray;
+
+export const ShouldFilter = {
```
sohanirao: Can we call this FilterResult
and use @enum to specify this as enum? instead of a map?

src/runtime/swg-client-event-manager.js
```diff
@@ -0,0 +1,312 @@
...
+ * The callback used to notify registered listeners about new client events.
+ * @callback SwgEventManagerListener
+ * @param {SwgClientEvent} event
+ */
+export let SwgEventManagerListener;
+
+/**
+ * Return false to cancel the event.  You can also modify the event object
+ * which is passed by reference.
+ * @callback SwgEventManagerFilterer
```
sohanirao: use function() JSDOC to specify this as well instead of @callback, because that is the pattern in the rest of the code.

src/runtime/swg-client-event-manager.js
```diff
@@ -0,0 +1,312 @@
...
+  /**True if the value is a function
+   * @param {*} value
+   * @private
+   */
+  static isFunction_(value) {
+    return typeof value === 'function';
+  }
+
+  constructor() {
+    /** @type {!CallBackArray} @private */
```
sohanirao: /** @private {!Array<!SwgEventListener>} */

sohanirao: private comes first and make it an Array object after removing CallBackArray

src/runtime/swg-client-event-manager.js
```diff
@@ -0,0 +1,312 @@
...
+    this.listeners_ = [];
+
+    /** @type {!CallBackArray} @private */
```
sohanirao: same as comment above

src/runtime/swg-client-event-manager.js
```diff
@@ -0,0 +1,312 @@
...
+    this.filterers_ = [];
+  }
+
+  /**Remove all existing filterers and listeners
+   */
+  clear() {
```
sohanirao: any use case for this API? Please feel free to refer me to to some documentation/design doc if you think I need more context on this.

src/runtime/swg-client-event-manager.js
```diff
@@ -0,0 +1,312 @@
...
+    this.listeners_ = [];
+    this.filterers_ = [];
+  }
+
+  /**
+   * Ensures the callback function is notified anytime one of the passed
+   * events occurs unless a filterer returns false.
+   * @param {!SwgEventManagerListener} callback
+   * @public
```
sohanirao: remove

src/runtime/swg-client-event-manager.js
```diff
@@ -0,0 +1,312 @@
...
+   */
+  addListener(callback) {
```
sohanirao: I like registerEventListerner better.

Are we no longer providing a list of events of interest? so by default they listen to all events? In that case lets make it specific what the callback is listening to
/**
 * Registers a callback to listen to all events
 * @param {SwgClientEventListener} callback
 */
registerEventListener(callback)

or 
/**
 * Registers a listener for events. If no event(s) is specified, the callback
 * is invoked for all events.
 * @param {(?Array<SwgClientEvent>|?SwgClientEvent)} events
 * @param {!SwgEventListener} callback
 */
registerListener(events, callback)

src/runtime/swg-client-event-manager.js
```diff
@@ -0,0 +1,312 @@
...
+    assert(SwgClientEventManager.isFunction_(callback),
+        'Invalid callback in SwgClientEventManager.addListener');
```
sohanirao: callback must be a function

sohanirao: if (! utils.isFunction(callback)) {
  throw new Error('Listener must be a function');
}

src/runtime/swg-client-event-manager.js
```diff
@@ -0,0 +1,312 @@
...
+
+  /**
+   * Filterers are called before listeners, if it returns
+   * ShouldFilter.STOP_EXECUTING then the event is canceled.
+   * @param {!SwgEventManagerFilterer} callback
+   * @public
+   */
+  addFilterer(callback) {
+    assert(SwgClientEventManager.isFunction_(callback),
+        'Invalid callback in SwgClientEventManager.addFilterer');
```
sohanirao: similar to above

src/runtime/swg-client-event-manager.js
```diff
@@ -0,0 +1,312 @@
...
+   * @public
+   */
+  addListener(callback) {
+    assert(SwgClientEventManager.isFunction_(callback),
+        'Invalid callback in SwgClientEventManager.addListener');
+    this.listeners_.push(callback);
+  }
+
+  /**
+   * Filterers are called before listeners, if it returns
```
sohanirao: Register a filterer for events so that depending on the
result of the filtering operation, the event may be dropped
before calling the listeners registered.
@param {!SwGEventFilterer} callback

src/runtime/swg-client-event-manager.js
```diff
@@ -0,0 +1,312 @@
...
+   * @param {!SwgEventManagerFilterer} callback
+   * @public
+   */
+  addFilterer(callback) {
+    assert(SwgClientEventManager.isFunction_(callback),
+        'Invalid callback in SwgClientEventManager.addFilterer');
+    this.filterers_.push(callback);
+  }
+
+  /**Call this function to inform all listeners that a client event has
```
sohanirao: Log an event.
The registered listeners will be invoked unless filtered.

src/runtime/swg-client-event-manager.js
```diff
@@ -0,0 +1,312 @@
...
+   * occurred.
+   * @param {!SwgClientEvent} event
+   * @public
+   */
+  logEvent(event) {
+    event.isValid(true);
+    let callbacks = this.filterers_;
```
sohanirao: any reason to have an extra variable here?

src/runtime/swg-client-event-manager.js
```diff
@@ -0,0 +1,312 @@
...
+    for (let callbackNum = 0; callbackNum < callbacks.length; callbackNum++) {
+      if (callbacks[callbackNum](event) === ShouldFilter.STOP_EXECUTING) {
+        return false;
```
sohanirao: does LogEvent has a return value? What is the caller expected to do with the return value?

sohanirao: I think what you want to do is not log the event here? just return; should do

src/runtime/swg-client-event-manager.js
```diff
@@ -0,0 +1,312 @@
...
+      }
+    }
+    callbacks = this.listeners_;
```
sohanirao: any reason we can't work with this.listeners_ directly?

src/runtime/swg-client-event-manager.js
```diff
@@ -0,0 +1,312 @@
...
+
+export const ShouldFilter = {
+  CONTINUE_EXECUTING: 0,
+  STOP_EXECUTING: 1,
+};
+
+export class SwgClientEvent {
+  /**
+   * Ensures the passed enum value is a member of the passed type or '*'.
+   * Throws an error if it isn't.
```
sohanirao: I'd have just one method 
/**
 * @param {!SwgClientEvent} event
 * @return {boolean}
 */
 isValidEvent(event) {
   if (event.type == 
 }

The caller of this event can decide to throw an error or not.

src/runtime/swg-client-event-manager.js
sohanirao: Also lets use SwgEventFilterer

src/runtime/swg-client-event-manager-test.js
```diff
@@ -0,0 +1,161 @@
...
+const DEFAULT_TYPE = AnalyticsEvent.IMPRESSION_AD;
+const DEFAULT_ORIGIN = EventOriginator.SWG_CLIENT;
+const DEFAULT_EVENT = new SwgClientEvent(DEFAULT_TYPE, DEFAULT_ORIGIN);
+const OTHER_TYPE = AnalyticsEvent.ACTION_PAYMENT_COMPLETE;
+const OTHER_ORIGIN = EventOriginator.AMP_CLIENT;
+const BAD_VALUE = 'I should throw an error';
+const EventManager = new SwgClientEventManager();
+
+describes.sandboxed('SwgClientEvent', {}, () => {
+  it('should respect properties that have been set', () => {
```
sohanirao: replace this with tests to just validate a set of valid and invalid events.

src/runtime/swg-client-event-manager-test.js
```diff
@@ -0,0 +1,161 @@
...
+    expect(event.getEventType()).to.equal(OTHER_TYPE);
+    expect(event.getEventOriginator()).to.equal(OTHER_ORIGIN);
+    expect(event.getIsFromUserAction()).to.be.true;
+    expect(event.getAdditionalParameters().aValue).to.equal(45);
+
+    event.setIsFromUserAction(null);
+    expect(event.getIsFromUserAction()).to.be.null;
+  });
+
+  it('should have a working copy constructor', () => {
```
sohanirao: Is this necessary? A copy constructor, given we are going to use a typedef for SwgClientEvent?

src/runtime/swg-client-event-manager.js
```diff
@@ -0,0 +1,312 @@
...
+        'Invalid callback in SwgClientEventManager.addFilterer');
+    this.filterers_.push(callback);
+  }
+
+  /**Call this function to inform all listeners that a client event has
+   * occurred.
+   * @param {!SwgClientEvent} event
+   * @public
+   */
+  logEvent(event) {
```
sohanirao: Since we are directly invoking callbacks within this method, we need to make sure the call to logEvent is non blocking. Can we think of ways to do so? Again please feel to refer me to this design consideration in a doc if you have already addressed it.

src/proto/api_messages_test.js
mborof: I think the issue here is that it was over indented before and I have unindented it to respect the 2 spaces rule

src/proto/api_messages_test.js
mborof: I added a space before the * so it is more standard.  I'm not sure why I have to manually format this file when I copy it into visual studio.

src/proto/api_messages_test.js
mborof: I think I've fixed the indenting

src/runtime/swg-client-event-manager-test.js
mborof: I thought it would be useful for the AMP client to be able to intercept the SwG event, suppress it, copy it and modify it by just setting the originator to itself and submitting a new event.  This means that if we add properties later AMP *might* be able to avoid some updates to set the properties because it can just take the ones set by SwG.

Copy might not be necessary thought - it could suppress the event and modify it without copying it.  I'm amenable to that.

sohanirao: thanks I'll resolve the conversation when i see a patch that addresses this.

src/runtime/swg-client-event-manager.js
sohanirao: Instead of clear() you might be wanting to support a "unregister" operation?

Supporting this though is not that straightforward. You might need to return a subscriberId when a subscriber registers a callback and associate the callback with that ID. So that when someone unregisters their callback, they invoke that ID. Unless we see a strong use case for some subscriber no longer wanting events, we may be over-engineering this.

src/runtime/swg-client-event-manager-test.js
mborof: The other thing that had occurred to me is that someone could declare a const at the top of the file like

const MY_EVENT = new SwgClientEvent(...);

And pass that to event manager.  We do not want people modifying these events.  So if they want to reuse it - they should copy it and make it their own.

src/runtime/swg-client-event-manager.js
mborof: That didn't work in that the exported variable was marked as "any" when I hovered over it in visual studio.  I modified this to use that syntax though without the labeling.  (its now directly in the function signatures and applied to the new const arrays.

src/runtime/swg-client-event-manager.js
mborof: Not sure specifically what you were referring to but I think this is deleted now anyway.

src/runtime/swg-client-event-manager.js
mborof: I think I did that properly.  At least hover is indicating the callback returns a number in Visual Studio Code

src/runtime/swg-client-event-manager.js
mborof: I've taken some of your suggestion, let me know if you want further refinement.

src/runtime/swg-client-event-manager.js
mborof: I've modified your suggestion

src/runtime/swg-client-event-manager.js
mborof: I don't really know what the caller would do with this information but its very easy to inform them whether or not the event is filtered so why not?

src/runtime/swg-client-event-manager.js
mborof: I believe this is done

src/runtime/swg-client-event-manager.js
mborof: I'm potentially amenable adding event but I intentionally took out "Event" from it because you are already accessing it through .getEventManager().addListener() so it seems like having event in there again is duplicative.

I am amenable to making it "register" instead of add but I see little difference between the meanings and add is shorter.

src/api/swg-client-event-manager-api.js
```diff
@@ -0,0 +1,69 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS-IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {AnalyticsEvent,EventOriginator} from '../proto/api_messages';
+
+/** @enum {number}  */
```
sohanirao: Can you add some comments to what the result means?

src/api/swg-client-event-manager-api.js
```diff
@@ -0,0 +1,69 @@
...
+ * - eventType: Required. The AnalyticsEvent type that occurred.
+ * - eventOriginator: Required.  The codebase that initiated the event.
+ * - additionalParameters: Optional.  A JSON object to store generic data.
+ * - isFromUserAction: Optional.  True if the user took an action to generate
+ *   the event.
+ *
+ *  @typedef {{
+ *    eventType: (!AnalyticsEvent),
+ *    eventOriginator: (!EventOriginator),
+ *    additionalParameters: (?Object),
```
sohanirao: can we reverse the order line 36/37

src/api/swg-client-event-manager-api.js
```diff
@@ -0,0 +1,69 @@
...
+ */
+export let SwgClientEvent;
+
+/**
+ * @interface
+ */
+export class SwgClientEventManagerApi {
+  /**
+   * Ensures the callback function is notified anytime one of the passed
+   * events occurs unless a filterer returns false.
```
sohanirao: one of the passed events implies we are passing events in the argument. It's confusing. Maybe I suggested this , sorry for asking you to change it again. Now that I read it, it does sound confusing. 
"Call this function to log an event.  The registered listeners will be
  invoked unless the event is filtered."

src/api/swg-client-event-manager-api.js
```diff
@@ -0,0 +1,69 @@
...
+
+  /**
+   * Register a filterer for events if you need to potentially cancel an event
+   * before the listeners are called.  A filterer should return
+   * FilterResultApi.STOP_EXECUTING to cancel an event.
+   * @param {!function(!SwgClientEvent):FilterResult} callback
+   */
+  addFilterer(callback) { }
+
+  /**Call this function to log an event.  The registered listeners will be
```
sohanirao: New line after /**

src/api/swg-client-event-manager-api.js
```diff
@@ -0,0 +1,69 @@
...
+export class SwgClientEventManagerApi {
+  /**
+   * Ensures the callback function is notified anytime one of the passed
+   * events occurs unless a filterer returns false.
+   * @param {!function(!SwgClientEvent)} callback
+   */
+  addListener(callback) { }
+
+  /**
+   * Register a filterer for events if you need to potentially cancel an event
```
sohanirao: what does it mean to cancel an event? Register a filterer for events so that depending on the result of the filtering operation, the event may be dropped before calling the listeners registered.

src/api/swg-client-event-manager-api.js
```diff
@@ -0,0 +1,69 @@
...
+   * before the listeners are called.  A filterer should return
+   * FilterResultApi.STOP_EXECUTING to cancel an event.
+   * @param {!function(!SwgClientEvent):FilterResult} callback
+   */
+  addFilterer(callback) { }
+
+  /**Call this function to log an event.  The registered listeners will be
+   * invoked unless the event is filtered.  Returns false if the event was
```
sohanirao: there is no return value right. Need to update this.

src/runtime/swg-client-event-manager.js
sohanirao: addEventListener would have made more sense if you were going to take some listeners for some events and other listeners for other events. It just seems like we are registering a callback for all events. I prefer "register" if both are okay with you.

src/runtime/swg-client-event-manager.js
```diff
@@ -0,0 +1,121 @@
...
+   * @overrides
+   */
+  addFilterer(callback) {
+    if (!isFunction(callback)) {
+      throw new Error('Event manager filterers must be a function');
+    }
+    this.filterers_.push(callback);
+  }
+
+  /**Call this function to log an event.  The registered listeners will be
```
sohanirao: new line after /**

src/runtime/swg-client-event-manager.js
```diff
@@ -0,0 +1,121 @@
...
+import {isObject, isFunction, isEnumValue} from '../utils/types';
+import * as EventManagerApi from '../api/swg-client-event-manager-api';
+
+/**
+ * Returns a standard string used to describe an issue with an event object
+ * @param {!string} valueName
+ * @param {?*} value
+ * @returns {!string}
+ */
+function getEventError(valueName, value) {
```
sohanirao: can you rename this to be createEventErrorMessage()?
In the comments on line 22, can you annotate this to be a helper function?

src/runtime/swg-client-event-manager.js
```diff
@@ -0,0 +1,121 @@
...
+  return 'An SwgClientEvent has an invalid ' + valueName + '(' + value + ')';
+}
+
+/**Throws an error if the event is invalid.
```
sohanirao: /** new line

src/api/subscriptions.js
```diff
@@ -242,6 +243,11 @@ export class Subscriptions {
    * @return {!Promise<PropensityApi>}
    */
   getPropensityModule() {}
+
+  /**Returns the SwgClientEventManager used for logging client events.
```
sohanirao: new line after /**

src/runtime/swg-client-event-manager-test.js
```diff
@@ -0,0 +1,155 @@
...
+      additionalParameters: {},
+    };
+    const eventMan = new SwgClientEventManager();
+
+    let errorCount = 0;
+    const tryIt = () => {
+      try {
+        eventMan.logEvent(event);
+      } catch (e) {
+        errorCount++;
```
sohanirao: do you not intend to validate the error message as well?

src/runtime/swg-client-event-manager-test.js
```diff
@@ -0,0 +1,155 @@
...
+
+  it('should be able to listen for events', function*() {
+    const eventMan = new SwgClientEventManager();
+    let receivedEventsCount = 0;
+    const callback = () => receivedEventsCount++;
+
+    //verify it can listen to 1
+    eventMan.addListener(callback);
+    yield eventMan.logEvent(DEFAULT_EVENT);
+
```
sohanirao: group lines 121, 122, 125 instead

src/runtime/swg-client-event-manager-test.js
```diff
@@ -0,0 +1,155 @@
...
+    expect(receivedEventsCount).to.equal(1);
+    //verify it can listen to 2 at the same time
+    eventMan.addListener(callback);
+    yield eventMan.logEvent(DEFAULT_EVENT);
+    expect(receivedEventsCount).to.equal(3);
+  });
+
+  it('should be able to filter out some events', function*() {
```
sohanirao: This is a good test. This test however does not verify if the right events were filtered. If the implementation did the opposite behavior (when stop executing is the return value, it actually continues and otherwise returns), this test will still pass without actually performing the task.

src/api/swg-client-event-manager-api.js
mborof: It's returning the promise of calling the filterers and callbacks.

src/runtime/swg-client-event-manager-test.js
mborof: Ok, done

src/runtime/swg-client-event-manager-test.js
mborof: I think it does because it log's an event that it expects to be filtered and then immediately checks the count to verify it is 0.  Then it logs an event that shouldn't be filtered and immediately checks to see the count is 1.  If it were doing the opposite of what was expected the first check (for 0) would fail.

src/runtime/swg-client-event-manager.js
mborof: did you just mean mention helper in the description?

src/runtime/swg-client-event-manager.js
mborof: I don't see any difference between add and register but I don't mind changing it.  For consistency I feel like filterers should have the same naming convention though.

src/api/swg-client-event-manager-api.js
mborof: (So someone could wait if they wanted or do something after the logging)

src/api/swg-client-event-manager-api.js
sohanirao: based on your comments it seems like the name:
DROP_EVENT
PROCESS_EVENT 
might be more descriptive enums? Wdyt?

src/api/swg-client-event-manager-api.js
sohanirao: Sorry i meant the additional parameters should be the last in the list of fields

src/api/swg-client-event-manager-api.js
sohanirao: Why not use the lastAction_ private variable for testing. The API caller itself doesn't seem to care about a returned promise.

src/runtime/runtime.js
```diff
@@ -876,9 +891,9 @@ function createPublicRuntime(runtime) {
     saveSubscription: runtime.saveSubscription.bind(runtime),
     createButton: runtime.createButton.bind(runtime),
     attachButton: runtime.attachButton.bind(runtime),
+    getPropensityModule: runtime.getPropensityModule.bind(runtime),
     attachSmartButton: runtime.attachSmartButton.bind(runtime),
```
sohanirao: any reason to change the order?

src/runtime/swg-client-event-manager-test.js
sohanirao: you are correct. it wont pass the first check. Still do you think there is value in testing that the filterer is filtering out the events that must be dropped or keeping those that should be processed?

src/runtime/swg-client-event-manager.js
```diff
@@ -0,0 +1,125 @@
...
+  return 'An SwgClientEvent has an invalid ' + valueName + '(' + value + ')';
+}
+
+/**
+ * Throws an error if the event is invalid.
+ * @param {!EventManagerApi.SwgClientEvent} event
+ * @returns {!Promise}
+ */
+function validateEvent(event) {
+  if (!event) {
```
sohanirao: by virtue of adding  a ! to the argument in line 33, this check will never happen. you can remove.

src/runtime/swg-client-event-manager.js
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  if (!isEnumValue(AnalyticsEvent, event.eventType)) {
+    throw new Error(createEventErrorMessage('eventType', event.eventType));
+  }
+  if (!isEnumValue(EventOriginator, event.eventOriginator)) {
+    throw new Error(createEventErrorMessage('eventOriginator',
+        event.eventOriginator));
+  }
+  if (!event.additionalParameters) {
```
sohanirao: any reason additionalParameters can't be null?

src/runtime/swg-client-event-manager.js
```diff
@@ -0,0 +1,125 @@
...
+    event.additionalParameters = {};
+  }
+  if (!isObject(event.additionalParameters)) {
+    throw new Error(createEventErrorMessage('additionalParameters',
+        event.additionalParameters));
+  }
+  if (event.isFromUserAction !== null
```
sohanirao: this lines of code is confusing. if you are simply trying to ensure that if this value is not null, its a bool, why not add to utils and use it here?

src/runtime/swg-client-event-manager.js
```diff
@@ -0,0 +1,125 @@
...
+    /** @private {!Array<function(!EventManagerApi.SwgClientEvent)>} */
+    this.listeners_ = [];
+
+    /** @private {!Array<function(!EventManagerApi.SwgClientEvent):!EventManagerApi.FilterResult>} */
+    this.filterers_ = [];
+  }
+
+  /**
+   * Ensures the callback function is notified anytime one of the passed
+   * events occurs unless a filterer returns false.
```
sohanirao: filterer does not return a boolean value.

src/runtime/swg-client-event-manager.js
sohanirao: sure

src/runtime/swg-client-event-manager.js
sohanirao: the callers are propensity APIs and analyticsService OSS logEvent API. These APIs don't expect a return value.

src/api/swg-client-event-manager-api.js
```diff
@@ -0,0 +1,73 @@
...
+  /** the event is canceled and the listeners are not informed about it */
+  STOP_EXECUTING: 1,
+};
+
+/**
+ * Defines a client event in SwG
+ * Properties:
+ * - eventType: Required. The AnalyticsEvent type that occurred.
+ * - eventOriginator: Required.  The codebase that initiated the event.
+ * - additionalParameters: Optional.  A JSON object to store generic data.
```
sohanirao: Can additional parameters be the last field in this typedef?

src/api/swg-client-event-manager-api.js
```diff
@@ -0,0 +1,73 @@
...
+   * @param {!function(!SwgClientEvent):FilterResult} callback
+   */
+  registerEventFilterer(callback) { }
+
+  /**
+   * Call this function to log an event.  The registered listeners will be
+   * invoked unless the event is filtered.  Returns false if the event was
+   * filtered and throws an error if the event is invalid.
+   * @param {!SwgClientEvent} event
+   * @returns {!Promise}
```
sohanirao: no return

src/runtime/swg-client-event-manager-test.js
```diff
@@ -0,0 +1,195 @@
...
+};
+
+/**
+ * Returns a standard string used to describe an issue with an event object
+ * @param {!string} valueName
+ * @param {?*} value
+ * @returns {!string}
+ */
+function createEventErrorMessage(valueName, value) {
+  return 'An SwgClientEvent has an invalid ' + valueName + '(' + value + ')';
```
sohanirao: Event has an invalid ....
is good enough. No need to specify SwgClientEvent. Unless you want them to refer to it. You can see, please SwGClientEvent definition.

src/api/swg-client-event-manager-api.js
mborof: That applies to the analytics service testing code.  The testing code I am referring to is the event manager testing code:

    eventMan.registerEventListener(callback);
    yield eventMan.logEvent(DEFAULT_EVENT);
    expect(receivedEventsCount).to.equal(3);

It needs to know the promise is fulfilled before it can test the expectations.

mborof: And there is nothing else to key on as far as I am aware.

src/api/swg-client-event-manager-api.js
mborof: If you have a better way of testing event manager I see no reason to have it

src/runtime/runtime.js
mborof: Not really, it just happened during merge.  Plus it seemed like the smart button came in after it so it should be last.  Doesn't really matter though

src/runtime/swg-client-event-manager-test.js
mborof: Yes, that's why I wrote this test.

src/runtime/swg-client-event-manager.js
mborof: My test code passes a null in and generates the error I created.  I replaced it with an isObject check.

src/runtime/swg-client-event-manager.js
mborof: I guess I lean towards it being a well formatted standard object.  I decided to use an isObject check and ensure it is null if it isn't an object.

mborof: I just want to ensure it always has a standard structure.  I changed it to required it to be an object or null.

src/runtime/swg-client-event-manager.js
mborof: Ok.  I did decide to keep the null check outside of the isBoolean check though.

src/runtime/swg-client-event-manager.js
mborof: I'm not sure what you're referring to

mborof:
FYI that we broke this up into smaller bits which will be done separately.

PR #553, opened 2019-05-16
Event man

mborof:
Functioning code for event manager EXCEPT I am getting a very illogical gulp presubmit error that Sohani and I can't figure out.

### DIFF
--- a/src/runtime/client-event-manager-test.js
+++ b/src/runtime/client-event-manager-test.js
@@ -178,17 +178,19 @@ describes.sandboxed('EventManager', {}, () => {
     //filter out the default origin
     eventMan.registerEventFilterer(event =>
       event.eventOriginator === DEFAULT_ORIGIN ?
-          EventManagerApi.FilterResult.STOP_EXECUTING :
-          EventManagerApi.FilterResult.CONTINUE_EXECUTING
+          EventManagerApi.FilterResult.CANCEL_EVENT :
+          EventManagerApi.FilterResult.PROCESS_EVENT
     );
 
     //ensure the default origin is filtered out
-    yield eventMan.logEvent(DEFAULT_EVENT);
+    eventMan.logEvent(DEFAULT_EVENT);
+    yield eventMan.lastAction_;
     expect(receivedEventsCount).to.equal(0);
 
     //ensure the other origin is not filtered out
     DEFAULT_EVENT.eventOriginator = OTHER_ORIGIN;
-    yield eventMan.logEvent(DEFAULT_EVENT);
+    eventMan.logEvent(DEFAULT_EVENT);
+    yield eventMan.lastAction_;
     expect(receivedEventsCount).to.equal(1);
     eventMan.eventOriginator = DEFAULT_ORIGIN;
   });
--- a/src/runtime/client-event-manager-test.js
+++ b/src/runtime/client-event-manager-test.js
@@ -32,16 +32,6 @@ const DEFAULT_EVENT = {
   additionalParameters: {},
 };
 
-/**
- * Returns a standard string used to describe an issue with an event object
- * @param {!string} valueName
- * @param {?*} value
- * @returns {!string}
- */
-function createEventErrorMessage(valueName, value) {
-  return 'Event has an invalid ' + valueName + '(' + value + ')';
-}
-
 describes.sandboxed('EventManager', {}, () => {
   it('should throw an error for invalid events', () => {
     /** @type {!EventManagerApi.ClientEvent} */
@@ -73,12 +63,12 @@ describes.sandboxed('EventManager', {}, () => {
 
     //validate event type
     event.eventType = BAD_VALUE;
-    expected = createEventErrorMessage('eventType', BAD_VALUE);
+    expected = 'Event has an invalid eventType (' + BAD_VALUE + ')';
     tryIt();
     expect(errorCount).to.equal(1);
     expect(matchedExpected).to.equal(1);
     event.eventType = null;
-    expected = createEventErrorMessage('eventType', null);
+    expected = 'Event has an invalid eventType (' + null + ')';
     tryIt();
     expect(errorCount).to.equal(2);
     expect(matchedExpected).to.equal(2);
@@ -91,12 +81,12 @@ describes.sandboxed('EventManager', {}, () => {
     errorCount = 0;
     matchedExpected = 0;
     event.eventOriginator = BAD_VALUE;
-    expected = createEventErrorMessage('eventOriginator', BAD_VALUE);
+    expected = 'Event has an invalid eventOriginator (' + BAD_VALUE + ')';
     tryIt();
     expect(errorCount).to.equal(1);
     expect(matchedExpected).to.equal(1);
     event.eventOriginator = null;
-    expected = createEventErrorMessage('eventOriginator', null);
+    expected = 'Event has an invalid eventOriginator (' + null + ')';
     tryIt();
     expect(errorCount).to.equal(2);
     expect(matchedExpected).to.equal(2);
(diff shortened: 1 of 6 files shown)


src/runtime/client-event-manager.js
```diff
@@ -15,33 +15,133 @@
...
+    if (!isFunction(filterer)) {
+      throw new Error('Event manager filterers must be a function');
+    }
+    this.filterers_.push(filterer);
   }
 
   /**
    * @overrides
    */
   logEvent(event) {
```
sohanirao: This is still as asynchronous we'd like it to be. The correct implementation is as what you had it before.

this.lastAction_ =  validateEvent(event).then(() => {
 // filter event
 // invoke callbacks
});

Let's revisit why you want lastAction to track the filtering. Once again, as a principle, tests should not dictate the implementation.

src/runtime/client-event-manager.js
```diff
@@ -15,33 +15,133 @@
...
+      for (let filterer = 0; filterer < this.filterers_.length; filterer++) {
+        if (this.filterers_[filterer](event) === FilterResult.CANCEL_EVENT) {
+          resolve();
+          return;
+        }
+      }
+
+      const promises = [];
+      let listenerNum = 0;
+      //builds an array of 1 promise per listener
```
sohanirao: since this whole part is running asynchornously after validate event returns to the caller, any reason you need to chain promises?

sounds like what you want to do is dispatch events to its listeners in an event queue pattern.

I'd like to understand the need for doing this though. If the entire filtering and invoking listeners is happening asynchronously as long as the event is validated, nothing is blocking on it. 

Another way to make it asynchronous is create a private method that returns the result of filtering as a promise. Since processing is blocked on filtering, you can make dispatching an event block on filter result. Example
/**
 * @param {!ClientEvent} event
 * @return {!Promise<boolean>}
 */
filterEvent(event) {
  let shouldProcess = true;
  this.filterers_.foreach(filterer => {
    // If any fliterer wants to cancel this event, cancel it
    if (filterer(event) == FilterResult.CANCEL_EVENT) {
      shouldProcess = false;
      break;
    }
  });
  return Promise.resolve(shouldProcess);
}
/**
 * @param {ClientEvent} event
 */
dispatchEvent(event) {
  this.listeners_.foreach(listener => {
    setTimeout(() => listener(event));
  });
}

logEvent(event) {
  return validateEvent(event).then(() => {
    // Filtering and dispatching the event are asynchronous
    return filterEvent(event);
  }).then(process => {
    if (process) {
      dispatchEvent(event);
    }
  });

mborof: We discussed and decided that 1 promise would be fine.  There is no need to asynchronously call each individual listener and then know that each of those listeners was called.

src/runtime/client-event-manager.js
mborof: We discussed my modified approach.

src/runtime/client-event-manager.js
```diff
@@ -15,33 +15,117 @@
...
+  if (!isEnumValue(AnalyticsEvent, event.eventType)) {
+    throw new Error(createEventErrorMessage('eventType', event.eventType));
+  }
+
+  if (!isEnumValue(EventOriginator, event.eventOriginator)) {
+    throw new Error(createEventErrorMessage('eventOriginator',
+        event.eventOriginator));
+  }
+
+  if (!isObject(event.additionalParameters)
```
sohanirao: Should have caught this earlier, my bad. Would like to understand why the code below won't suffice? If it's not null, it should be an object. Otherwise, it can be null or undefined. 

if (event.additionalParameters != null &&
     !isObject(event.additionalParameters)) {
  throw error
}

src/runtime/client-event-manager-test.js
```diff
@@ -0,0 +1,199 @@
...
+  additionalParameters: {},
+};
+
+/**
+ * Returns a standard string used to describe an issue with an event object
+ * @param {!string} valueName
+ * @param {?*} value
+ * @returns {!string}
+ */
+function createEventErrorMessage(valueName, value) {
```
sohanirao: The catch here is that the same code is being used in test and in prod. So we won't ever know what to expect.  Wherever there's expected message, why not just define a mannual string. You are making a call to mannually create the expected message anyway.

src/runtime/client-event-manager.js
mborof: It seems like a property most workflows shouldn't really need to worry about.  I don't see any reason to make them fill in an empty {}.

src/runtime/client-event-manager-test.js
mborof: For the same reason you functionalize things for the production code: so if you change it you only need to change it once. It's now twice (once for the actual code and once for the test code) - but at least it isn't 8 times that then grows linearly every time you add another property.

mborof: Though truthfully I don't understand your comment "we won't ever know what to expect".

sohanirao: when test code duplicates production code, it's not clear what to expect. i'd recommend removing expected message helper in the test and actually spell out the entire string. This makes it clear what is expected and how the implementation accomplishes that is not necessary for the test to know.

src/runtime/client-event-manager.js
sohanirao: it is still unclear what your code is trying to do. if its not null and not an object throw an error. there seems to be no harm in sending {}.

src/runtime/client-event-manager-test.js
```diff
@@ -0,0 +1,199 @@
...
+    let receivedEventsCount = 0;
+    const callback = () => receivedEventsCount++;
+
+    //verify it can listen to 1
+    eventMan.registerEventListener(callback);
+    eventMan.logEvent(DEFAULT_EVENT);
+    yield eventMan.lastAction_;
+    expect(receivedEventsCount).to.equal(1);
+
+    //verify it can listen to 2 at the same time
```
sohanirao: do you mean 2 of the same events? we are not sending two events at the same time. This is the second event yes?

mborof: I mean verify EventManager supports multiple listeners.  It is sending 1 event and there are 2 registered listeners.  Granted its the same listener twice but that shouldn't matter.  It then verifies its listener got called twice after a single event is sent.

src/runtime/client-event-manager.js
mborof: I don't see code that gives something a default value if its undefined as particularly confusing.  Seems like a fairly common operation.  I did remove it anyway because whether its there or not

src/runtime/client-event-manager-test.js
mborof: It just means that if the error message is changed it needs to be updated in multiple locations but OK

src/runtime/client-event-manager-test.js
mborof: I did clarify the comment

src/runtime/client-event-manager.js
sohanirao: Thanks it's a lot more readable. 
nit: you might want to reverse the order. the behavior of toString on a null value is [Object null] in JavaScript and the isObject() method invokes this method internally. 

if (!!event.additionalParameters && !isObject(event.additionalParameters)
This way, if the event.additionalParameters is null, the check can exit immediately. If event.additionalParameters is not null, then actually check if it's an object.

Again don't have to do it this way. Just a nit if you agree that it looks simpler or easier to read.

src/runtime/client-event-manager.js
sohanirao: To conclude, since the listeners are all internal callbacks (propensity service, analytics service and AMP), we can assure the quality of these listeners and invoke them synchronously.

PR #554, opened 2019-05-21
Event Manager / Propensity Client

mborof:
This is the bulk of changes required by Propensity to start using the events manager.  Still to do would be actually sending the new data to the server (and logging it).  There will be a future branch focusing on just that.


mborof:
Hold off until the other event manager branch is merged

mborof:
This is ready for review.

src/runtime/propensity.js
```diff
@@ -15,7 +15,23 @@
...
 import * as PropensityApi from '../api/propensity-api';
 import {PropensityServer} from './propensity-server';
-import {isObject} from '../utils/types';
+import {isObject,isEnumValue} from '../utils/types';
+import {AnalyticsEvent,EventOriginator} from '../proto/api_messages';
+import {Event} from '../api/propensity-api';
+
+/** @private @const {!Object<string,AnalyticsEvent>} */
+const PropensityEventToAnalyticsEvent = {
+  [Event.IMPRESSION_PAYWALL]: AnalyticsEvent.IMPRESSION_PAYWALL,
```
sohanirao: why square brackets?

src/runtime/propensity-server.js
```diff
@@ -14,7 +14,34 @@
...
+
+/** @private @const {!Object<number,string>} */
+const AnalyticsEventToPropensityEvent = {
+  [AnalyticsEvent.UNKNOWN]: null,
+  [AnalyticsEvent.IMPRESSION_PAYWALL]: Event.IMPRESSION_PAYWALL,
+  [AnalyticsEvent.IMPRESSION_AD]: Event.IMPRESSION_AD,
+  [AnalyticsEvent.IMPRESSION_OFFERS]: Event.IMPRESSION_OFFERS,
+  [AnalyticsEvent.IMPRESSION_SUBSCRIBE_BUTTON]: null,
+  [AnalyticsEvent.IMPRESSION_SMARTBOX]: null,
+  [AnalyticsEvent.ACTION_SUBSCRIBE]: null,
```
sohanirao: AnalytisEvent.ACTION_SUBSCRIBE maps to Event.ACTION_PAYMENT_FLOW_STARTED

src/runtime/propensity.js
mborof: If you take out the square brackets the syntax doesn't compile.  I wasn't sure how to actually make the keys defined from an enum value until I saw it done with the [...] in another file.

src/runtime/propensity-server.js
mborof: There is also an AnalyticsEvent.ACTION_PAY_FLOW_STARTED.

mborof: I see that ACTION_SUBSCRIBE is used in 1 place in the codebase: it might be preferable to replace it with ACTION_PAYMENT_FLOW_STARTED if that is in fact what it means.  It's a much better enum value

src/runtime/experiment-flags.js
```diff
@@ -43,12 +43,18 @@ export const ExperimentFlags = {
   CONTRIBUTIONS: 'contributions',
 
   /**
-   * Enables the Propensity feature
+   * Enables the Propensity feature.
```
sohanirao: this can be removed.

src/runtime/propensity-test.js
```diff
@@ -166,4 +172,60 @@ describes.realWin('Propensity', {}, env => {
       expect(propensityScore.body.result).to.equal(42);
     });
   });
+
+  it('should convert all propensity events to analytics events', () => {
+    let eventSent = null;
+    const params = /** @type {JsonObject} */ ({
+      'source': 'user-action',
```
sohanirao: lets remove is_active since we plan to clarify the API anyway.
And let's use a source that's more believable like  'email'

Test code can serve as an example of API usage so lets be careful, although I recognize it does not make any difference to the test logic.

src/runtime/propensity-test.js
```diff
@@ -126,24 +128,28 @@ describes.realWin('Propensity', {}, env => {
...
+  it('should send events to event manager', () => {
     let eventSent = null;
-    let paramsSent = null;
-    const params = /** @type {JsonObject} */ ({'source': 'user-action'});
-    const eventParams = JSON.stringify(params);
-    const propensityEvent = {
-      name: PropensityApi.Event.IMPRESSION_OFFERS,
+    const params = /** @type {JsonObject} */ ({
+      'source': 'user-action',
+      'is_active': false,
```
sohanirao: lets remove is_active and the value for 'source' to email.

src/runtime/experiment-flags.js
mborof: ok

src/runtime/propensity-test.js
mborof: ok

src/runtime/propensity-test.js
mborof: Ok

src/runtime/propensity-server.js
```diff
@@ -39,6 +67,12 @@ export class PropensityServer {
     this.xhr_ = new Xhr(win);
     /** @private @const {number} */
     this.version_ = 1;
+
+    eventManager.registerEventListener(event => this.sendClientEvent_(event));
+
+    /** @private @const {!boolean} */
+    this.logSwgEvents_ = isExperimentOn(win,
```
sohanirao: // TODO(mborof): b/133519525

src/runtime/propensity-server.js
```diff
@@ -125,6 +159,28 @@ export class PropensityServer {
     return this.xhr_.fetch(url, init);
   }
 
+  /**
+   *
+   * @param {!../api/client-event-manager-api.ClientEvent} event
+   */
+  sendClientEvent_(event) {
```
sohanirao: I would call this handleClientEvents_() since we are not always sending events.

src/runtime/propensity-server.js
```diff
@@ -116,7 +150,7 @@ export class PropensityServer {
     if (context) {
       eventInfo = eventInfo + ':' + encodeURIComponent(context);
     }
-    let url = ServiceUrl.adsUrl('/subopt/data?events=')
+    let url = adsUrl('/subopt/data?events=')
```
sohanirao: Line 143: Any reason this method should stay public?

if the only caller is from within we can rename this to sendEvent_() and make it private.

src/runtime/propensity.js
```diff
@@ -69,18 +89,19 @@ export class Propensity {
 
   /** @override */
   sendEvent(userEvent) {
-    if (!Object.values(PropensityApi.Event).includes(userEvent.name)) {
+    if (!isEnumValue(PropensityApi.Event, userEvent.name)
+        || !PropensityEventToAnalyticsEvent[userEvent.name]) {
       throw new Error('Invalid user event provided');
```
sohanirao: Do you see any issues in augmenting the error message on line 94 to include the userEvent.name?

src/runtime/propensity-server.js
sohanirao: my bad: This is the right mapping.
  [AnalyticsEvent.ACTION_PAYMENT_FLOW_STARTED]:
      Event.ACTION_PAYMENT_FLOW_STARTED,

ACTION_SUBSCRIBE tracks click on subscribe button. There's no equivalent of that in Propensity world. Please ignore this whole set.

src/runtime/propensity-server.js
mborof: Ok

src/runtime/propensity-server.js
mborof: Ok

src/runtime/propensity-server-test.js
```diff
@@ -296,4 +312,132 @@ describes.realWin('PropensityServer', {}, env => {
           expect(() => {throw reason;}).to.throw(/Invalid request/);
         });
   });
+
+  it('should listen for events from event manager', function*() {
+    //event registration happens during initialization so we need to stub it
+    //and then create a new propensity server
+    const eventMan = new ClientEventManager();
```
sohanirao: Why not use eventManager object created in line 43?

src/runtime/propensity-server-test.js
```diff
@@ -296,4 +312,132 @@ describes.realWin('PropensityServer', {}, env => {
...
+    let receivedEvent = null;
+    sandbox.stub(PropensityServer.prototype, 'sendClientEvent_', event => {
```
sohanirao: Line 321 and similar:
Instead of stubbing an internal method of the class under test, the right way to do is to mock the method of EventManager, registerEventListener and capture the argument to that call. and that verifies that PropensityServer listens for events from the event manager. None of the event manager specific code should be tested here.

A helper method that returns the event sent in the URL
/**
 * @param {string} capturedUrl
 * @return {?PropensityEvent}
 */
 getPropensityEventFromUrl(capturedUrl) {
        const queryString = capturedUrl.split('?')[1];
        const queries = parseQueryString(queryString);
        expect('events' in queries).to.be.true;

        const events = decodeURIComponent(queries['events'].split(':'));
       const eventName = events[1];
       const eventParam = parseJson(events[2]);
       const propensityEvent = /** @typedef {PropensityEvent} */({
         name: eventName,
         active: eventParam['is_active'],
         data: eventParam,
       } 
        expect(events).to.equal(JSON.stringify(eventParam));

The best way to test the mapping is to capture what gets sent in the fetch API.
    let capturedUrl;
    let capturedRequest;
    sandbox.stub(Xhr.prototype, 'fetch',
        (url, init) => {
          capturedUrl = url;
          capturedRequest = init;
          return Promise.reject('Server down');
        });

Then when you invoke the captured handler with different events, you can test your can test what was sent in the URL
return capturedListenerCallback(defaultAnalyticsEvent).then(() => {
  throw new Error('must have failed');
}).catch(() => {
        const propensityEventSent = this.getPropensityEventFromUrl(capturedUrl);
       // expectation on this event.
        expect(() => {
          throw reason;
        }).to.throw(/Server down/);
});

src/runtime/propensity-server.js
```diff
@@ -14,7 +14,34 @@
...
 import {Xhr} from '../utils/xhr';
-import * as ServiceUrl from './services';
+import {adsUrl} from './services';
+import {Event} from '../api/propensity-api';
+import {AnalyticsEvent,EventOriginator} from '../proto/api_messages';
+import {isObject,isBoolean} from '../utils/types';
+import {ExperimentFlags} from './experiment-flags';
+import {isExperimentOn} from './experiments';
+
+/** @private @const {!Object<number,string>} */
```
sohanirao: This is @private to which class? Should this be inside the constructor?

src/runtime/propensity-server-test.js
```diff
@@ -296,4 +312,132 @@ describes.realWin('PropensityServer', {}, env => {
...
+    new PropensityServer(win, pubId, eventMan);
+    eventMan.logEvent(defaultEvent);
+    yield eventMan.lastAction_;
+    expect(defaultEvent === receivedEvent).to.be.true;
+  });
+
+  it('should convert analytics events to propensity events', () => {
+    let receivedType = null;
+    let receivedContext = null;
+    sandbox.stub(PropensityServer.prototype, 'sendEvent',
```
sohanirao: I see you are trying to test the mapping here. Since sendEvent_ is going to be a private method and nobody outside is calling it, testing arguments passed in between private methods inside a class amounts to testing implementation. We need to test behavior. 

Do you need to test the mapping?

src/runtime/propensity-server-test.js
```diff
@@ -296,4 +312,132 @@ describes.realWin('PropensityServer', {}, env => {
...
+        case AnalyticsEvent.EVENT_CUSTOM:
+          expect(receivedType).to.equal(PropensityApi.Event.EVENT_CUSTOM);
+          break;
+        default:
+          expect(false).to.be.true; //add new analytics event to the switch
+      }
+    }
+  });
+
+  it('should respect the SwG logging experiment flag', () => {
```
sohanirao: Better to break this up into two tests.
// Mock the EventManager.registerEventListener() method and capture the callback

     numFetchApiCalls = 0;
    sandbox.stub(Xhr.prototype, 'fetch',
        (url, init) => {
          numFetchApiCalls++;
          return Promise.reject(new Error('Publisher not whitelisted'));
        });

One where it 'should send SwG events to Propensity Service'
// Enable experiment flag
// invoke the callback and catch
captureCallback(swgOriginatedEvent).then(() => {
  throw new Error('must have failed');
}).catch(reason => {
  expect(numFetchApiCalls).to.equal(0);
  expect(() => throw reason).to.throw(/Server down/);
});

One where it 'should not send SwG events to Propensity Service'
// same setup
captureCallback(swgOriginatedEvent).then(() => {
  throw new Error('must have failed');
}).catch(reason => {
  expect(numFetchApiCalls).to.equal(0);
  expect(() => throw reason).to.throw(/Server down/);
});

src/runtime/propensity-server-test.js
```diff
@@ -296,4 +312,132 @@ describes.realWin('PropensityServer', {}, env => {
...
+    let receivedType = null;
+    let receivedContext = null;
+    sandbox.stub(PropensityServer.prototype, 'sendEvent',
+        (eventType, context) => {
+          receivedType = eventType;
+          receivedContext = context;
+        });
+
```
sohanirao: Better to avoid testing the internal implementation if there's another way.

src/runtime/propensity-server-test.js
mborof: The comment above it explains why I had done it.  I've actually changed the implementation so it might not be necessary now - it depends how I register the listener.  

The syntax now: eventManager.registerEventListener(event => this.handleClientEvent_(event));
The way I had registered it initially: eventManager.registerEventListener(this.handleClientEvent_.bind(this));

For the second implementation, if I had stubbed it after the registration occurred then it would remember the original version of the function rather than the stubbed version (so it didn't work).  The way I'm doing it now doesn't matter.  I can do it either way.  I'm not sure which method is preferable.

src/runtime/propensity-server.js
mborof: It's a non-exported private const used for mapping different event types.  I could certainly put it inside the class if you think it should go there.  It's used exclusively by 1 line of code in propensity-server.

src/runtime/propensity-server.js
mborof: Agreed

src/runtime/propensity.js
mborof: I think its a good idea

src/runtime/propensity-server-test.js
mborof: I initially thought I didn't but then later thought it would be better if I did.  I'm not sure - do I need to test the mapping?    I lean towards yes in case someone accidentally changes it but I'm not really sure.

sohanirao: as discussed, lets create another class to centralize event mapping across different entities that use it.

src/runtime/propensity-server-test.js
mborof: Done!

src/runtime/propensity-server-test.js
mborof: Done!

src/runtime/propensity-server-test.js
mborof: Ok

src/runtime/propensity-server-test.js
mborof: I think this is done

src/runtime/propensity.js
```diff
@@ -69,18 +75,35 @@ export class Propensity {
...
     }
-    // TODO(sohanirao, mborof): Idenfity the new interface with event
-    // manager and update the lines below to adhere to that interface
-    let paramString = null;
-    if (userEvent.data) {
-      paramString = JSON.stringify(userEvent.data);
+
+    //event manager requires data to be a real object so fill it in if
+    //the publisher didn't pass the field in
+    if (!userEvent.data) {
```
sohanirao: Line 86, 87 feels like redundant code

sohanirao: if (userEvent.data && !isObject(userEvent.data)) {
      throw new Error('Event data must be an Object');
    }

Any reason why this code was not sufficient?

src/runtime/propensity-server.js
```diff
@@ -39,6 +48,16 @@ export class PropensityServer {
     this.xhr_ = new Xhr(win);
     /** @private @const {number} */
     this.version_ = 1;
+
+    eventManager.registerEventListener(event => this.handleClientEvent_(event));
+
+    // TODO(mborof): b/133519525
+    /** @private @const {!boolean} */
```
sohanirao: I think we can keep it as this.logSwgEvents_ = isExperimentOn(...)

and then and it with the enable logging flag. No?

Any reason we need to keep track of two flags?

src/runtime/propensity-server-test.js
```diff
@@ -76,7 +114,7 @@ describes.realWin('PropensityServer', {}, env => {
           return Promise.reject(new Error('Not sent from allowed origin'));
         });
     const eventParam = {'is_active': false, 'offers_shown': ['a', 'b', 'c']};
-    return propensityServer.sendEvent(
+    return propensityServer.sendEvent_(
```
sohanirao: this should happen in the then of the callback invocation. first capture the callback from mocking the registerEventListener interface.

src/runtime/propensity-server-test.js
sohanirao: you are still calling handleClientEvent_() which is a internal private method.

You need to capture the callback registered if you want to invoke the listener.

src/runtime/propensity-server-test.js
```diff
@@ -296,4 +334,81 @@ describes.realWin('PropensityServer', {}, env => {
...
+      return Promise.reject('Server down');
+    });
+
+    setExperiment(win, ExperimentFlags.LOG_SWG_TO_PROPENSITY, true);
+    eventManagerMock.expects('registerEventListener').once();
+    propensityServer = new PropensityServer(win, pubId, eventManager);
+
+    //both experiment and enable: ensure it actually logs
+    propensityServer.enableLoggingGoogleEvents();
+    propensityServer.handleClientEvent_(defaultEvent);
```
sohanirao: should not have to call internal private methods to do this test.

src/runtime/propensity.js
mborof: It caused the test code to fail so I figured it was supposed to work

mborof: It didn't fail before because nothing had a hard requirement that this was an object until event manager

sohanirao: Here's a piece of code that tested it already
it('should provide valid event', () => {
    const correctEvent = {
      name: PropensityApi.Event.IMPRESSION_PAYWALL,
      active: false,
    };
    expect(() => {
      propensity.sendEvent(correctEvent);
    }).to.not.throw('Invalid user event provided');
    const incorrectEvent = {
      name: 'user-redirect',
    };
    expect(() => {
      propensity.sendEvent(incorrectEvent);
    }).to.throw('Invalid user event provided');
    const incorrectEventParam = {
      name: PropensityApi.Event.IMPRESSION_OFFERS,
      active: true,
      data: 'all_offers',
    };
    expect(() => {
      propensity.sendEvent(incorrectEventParam);
    }).to.throw(/Event data must be an Object/);
  });

Still not sure how your code is different other than splitting up the checks into two lines?

src/runtime/propensity-test.js
```diff
@@ -70,32 +78,10 @@ describes.realWin('Propensity', {}, env => {
     }).throw(/Entitlements must be an Object/);
   });
 
-
-  it('should provide valid event', () => {
```
sohanirao: why is this test removed?

mborof: I reordered the tests to group logical items together.  It was just moved.

src/runtime/propensity.js
mborof: Here is the specific test that used to work that would no longer work if I throw an error without a data object:

    const correctEvent = {
      name: PropensityApi.Event.IMPRESSION_PAYWALL,
      active: false,
    };
    expect(() => {
      propensity.sendEvent(correctEvent);
    }).to.not.throw('Invalid user event provided');

src/runtime/propensity-server-test.js
mborof: I think I did what you wanted

src/runtime/propensity-server-test.js
mborof: Now I think I know what you mean.  Done?

src/runtime/propensity-server-test.js
mborof: Done?

src/runtime/propensity-server-test.js
sohanirao: just realized one more thing. since you are programming is_active field into the event parameters based on active field of the PropensityEvent, it's better to test that as well.

mborof: Done

src/runtime/propensity-server.js
```diff
@@ -201,4 +246,8 @@ export class PropensityServer {
           return this.parsePropensityResponse_(response);
         });
   }
+
```
sohanirao: Since this is a public API needs documentation.

src/runtime/propensity-server.js
```diff
@@ -39,6 +48,16 @@ export class PropensityServer {
     this.xhr_ = new Xhr(win);
     /** @private @const {number} */
     this.version_ = 1;
+
+    eventManager.registerEventListener(event => this.handleClientEvent_(event));
```
sohanirao: We need to bind the callback to this instance of the class. 

eventManager.registerEventListener(this.handleClientEvent_.bind(this));

src/runtime/propensity-server.js
sohanirao: And lets rename to enableLoggingSwgEvents()

These are not google events.

mborof: Done

src/runtime/propensity-server.js
mborof: It seemed to work either way, I wasn't sure which was preferable.  I've changed it back to a bind

src/runtime/propensity.js
```diff
@@ -69,27 +76,44 @@ export class Propensity {
...
     }
-    if (userEvent.data) {
-      paramString = JSON.stringify(userEvent.data);
+
+    if (isBoolean(userEvent.active)) {
+      if (!data) {
+        data = {};
+      }
+      Object.assign(data, {'is_active': userEvent.active});
+    } else if (userEvent.active != null) {
```
sohanirao: Now that we are not enforcing userEvent.active to be null and not undefined in case the caller has no information about the event being from a user action or not, we need to update your Event Manager code to not fail validation of the event and corresponding tests. You can either make that change first as a different PR or roll that into this one.

sohanirao: I looked into this. Your point is valid. We need to send something, undefined won't do. However, I wanted to find a way to handle it that is consistent with the rest of our code based. I found several instances of '|| null' to cover cases where undefined could be passed into an API for the field of an object defined using typedef.

Whenever a typedef requires something to be set and won't accept undefined, we OR the incoming object with a null. In this case it would look like this.

    this.eventManager_.logEvent({
    this.propensityServer_.sendEvent(userEvent.name, paramString);	      eventType: analyticsEvent,
      eventOriginator: EventOriginator.PROPENSITY_CLIENT,
      isFromUserAction: userEvent.active || null,
      additionalParameters: data,
    });

I still think we don't have to have a hard check to ensure its not undefined because the code already doesn't allow it. So != is sufficient IMHO. But to your point, to make it comply, following the code pattern in the rest of the code and ORing with null should address your concerns. 
 Let me know your thoughts.

sohanirao: given that we have a new way to handle undefined, can we update the message on 101 to be Active field must be boolean. 
Because if its not boolean, both null or undefined are okay. we handle it in line 107.

mborof: I'm still quite convinced that only test code can ever pass in undefined because the compiler will reject anything that allows undefined.  The reason it works in the tests is that the compiler seems to be less strict for them.  Real code won't be able to do it (except perhaps through Propensity where the publishers might not be doing strict compilation checking).  

I'd also strongly prefer not allowing it to be either undefined or null and treating them as the same, but it isn't actually all that important so I've just gone ahead and made that change.

mborof: I didn't see the previous 2 comments.  I don't think the || null is necessary because Propensity.Event doesn't support not having active.

sohanirao: Believe we have a resolution to this.

PR #555, opened 2019-05-21
Adds Impression Logging for SwG Button and SmartBox

elijahsoria:
Adds impression logging for the SwG "Subscribe" button and the SmartBox. Modifies Runtime to allow for ConfiguredRuntime to be in charge of all buttons. Adds tests.


src/runtime/button-api.js
```diff
@@ -60,10 +62,17 @@ export class ButtonApi {
 
   /**
    * @param {!../model/doc.Doc} doc
+   * @param {!./deps.DepsDef} deps
    */
-  constructor(doc) {
+  constructor(doc, deps) {
```
dparikh: `deps` has `doc` so no need to pass `doc` param anymore.

src/runtime/button-api.js
```diff
@@ -116,6 +125,13 @@ export class ButtonApi {
     }
     button.setAttribute('title', msg(TITLE_LANG_MAP, button) || '');
     button.addEventListener('click', callback);
+    if (this.deps_.config().analyticsMode == AnalyticsMode.IMPRESSIONS) {
+      this.deps_.entitlementsManager().getEntitlements().then(entitlements => {
+        this.analyticsService_.setReadyToPay(entitlements.isReadyToPay);
+        this.analyticsService_.logEvent(
```
mborof: Please specify the future value of isFromUserAction in a comment

src/runtime/button-api.js
```diff
@@ -152,21 +168,25 @@ export class ButtonApi {
...
     const options = this.getOptions_(optionsOrCallback);
     const callback = /** @type {function()} */
         (this.getCallback_(optionsOrCallback, opt_callback));
 
     // Add required CSS class, if missing.
     button.classList.add('swg-smart-button');
 
+    let analyticsRequest = null;
+    if (this.deps_.config().analyticsMode == AnalyticsMode.IMPRESSIONS) {
+      analyticsRequest = this.analyticsService_.createLogRequest(
```
mborof: Please specify the future value of isFromUserAction in a comment

src/runtime/analytics-service.js
```diff
@@ -195,7 +195,7 @@ export class AnalyticsService {
    * @param {!../proto/api_messages.AnalyticsEvent} event
    * @return {!AnalyticsRequest}
    */
-  createLogRequest_(event) {
+  createLogRequest(event) {
```
mborof: Just so you are aware, I am modifying this function to take the new ClientEvent object for events manager in a different pull request.  This is just FYI - I'm not sure who will beat who to final commit but you may need to resolve it on merge.  You can resolve this conversation (plus the other ones once you add the comments I asked for).

src/runtime/button-api.js
elijahsoria: Specified that isFromUserAction should be false.

src/runtime/button-api.js
elijahsoria: Specified that isFromUserAction should be false.

src/runtime/analytics-service.js
elijahsoria: Got it. Thanks for the heads up.

src/runtime/button-api.js
```diff
@@ -116,6 +125,14 @@ export class ButtonApi {
     }
     button.setAttribute('title', msg(TITLE_LANG_MAP, button) || '');
     button.addEventListener('click', callback);
+    if (this.deps_.config().analyticsMode == AnalyticsMode.IMPRESSIONS) {
+      this.deps_.entitlementsManager().getEntitlements().then(entitlements => {
+        // isFromUserAction = False
+        this.analyticsService_.setReadyToPay(entitlements.isReadyToPay);
+        this.analyticsService_.logEvent(
+            AnalyticsEvent.IMPRESSION_SUBSCRIBE_BUTTON);
```
sohanirao: you can send the event without getting entitlements and ready to pay. 
The analytics Service will be set with ready to pay bit already if entitlements fetch was done.

Is there a reason why you need the ready to pay?

sohanirao: I'd just do if (this.deps_.config().analyticsMode == AnalyticsMode.IMPRESSIONS) {
                    this.analyticsService_.logEvent(AnalyticsEvent.IMPRESSION_SUBSCRIBE_BUTTON);
}

src/runtime/button-api.js
```diff
@@ -152,21 +169,26 @@ export class ButtonApi {
...
     const callback = /** @type {function()} */
         (this.getCallback_(optionsOrCallback, opt_callback));
 
     // Add required CSS class, if missing.
     button.classList.add('swg-smart-button');
 
+    let analyticsRequest = null;
+    if (this.deps_.config().analyticsMode == AnalyticsMode.IMPRESSIONS) {
+      // isFromUserAction = False
+      analyticsRequest = this.analyticsService_.createLogRequest(
```
sohanirao: I'd remove this from here. Since SmartSubscriptionButtonApi has deps, it can get access to AnalyticsService directly.

src/runtime/button-api.js
```diff
@@ -152,21 +169,26 @@ export class ButtonApi {
...
+          AnalyticsEvent.IMPRESSION_SMARTBOX).toArray();
+    }
     return new SmartSubscriptionButtonApi(
-        deps, button, options, callback).start();
+      this.deps_, button, options, analyticsRequest, callback).start();
```
sohanirao: no need to send analyticsRequest as an argument.

src/runtime/smart-button-api.js
```diff
@@ -42,9 +42,10 @@ export class SmartSubscriptionButtonApi {
    * @param {!./deps.DepsDef} deps
    * @param {!Element} button
    * @param {!../api/subscriptions.ButtonOptions} options
+   * @param {?Array<../proto/api_messages.AnalyticsRequest>} analyticsRequest
```
sohanirao: no need for htis argument.

src/runtime/smart-button-api.js
```diff
@@ -80,6 +81,7 @@ export class SmartSubscriptionButtonApi {
       'publicationId': this.deps_.pageConfig().getPublicationId(),
       'theme': this.options_ && this.options_.theme || 'light',
       'lang': this.options_ && this.options_.lang || 'en',
+      'analyticsRequest': analyticsRequest,
```
sohanirao: I'd send analyticsContext here.

this.deps_.analytics().getAnalyticsContext().then(context => {
    // use context
}

And this way the iframe has context. When the iframe sends the ready() signal is when it has rendered the smart box. The event should be generated from boq-web iframe.

src/runtime/analytics-service.js
sohanirao: I'd keep this private and expose another method
/**
 * @return {!Promise<!AnalyticsContext>}
 */
getContext() {
  return this.start_().then(() => {
    return this.context_;
  }
}

src/runtime/button-api.js
elijahsoria: isReadyToPay being logged was a request by Shams. He mentioned that we needed to log this information, and we should couple it in this launch.

elijahsoria: Leaving the entitlements promise check but removing the setReadyToPay call for now.

src/runtime/analytics-service.js
elijahsoria: Done.

src/runtime/smart-button-api.js
elijahsoria: Done.

src/runtime/smart-button-api.js
elijahsoria: Thanks!

src/runtime/button-api.js
elijahsoria: Done.

src/runtime/button-api.js
sohanirao: the pitfall of waiting for an entitlements fetch is
1. You will send the SUBSCRIBE BUTTON IMPRESSION when entitlements response comes back, not when the button is rendered
2. Ideally, you want to be honest about the data you are collecting. Meaning; If the publisher fetched entitlements, and the entitlements response was available at the time the subscribe button was rendered, then you log that information.  But if entitlements fetch was not done or the response was not available, when the subscribe button was rendered, your data should say so. 
For these two reasons, you should remove the entitlements check too.

src/runtime/analytics-service.js
```diff
@@ -191,11 +191,18 @@ export class AnalyticsService {
...
+  getContext() {
+    return this.context_;
+  }
+
   /**
    * @param {!../proto/api_messages.AnalyticsEvent} event
    * @return {!AnalyticsRequest}
    */
-  createLogRequest_(event) {
+  createLogRequest(event) {
```
sohanirao: I am unable to load our conversation. But I thought in exporting the getContext method, we could leave this method be private and unchanged. 

Did you mean to leave the _ in line 206 at the end of createLogRequest()?

src/runtime/button-api.js
elijahsoria: Done. Thanks for the explanation!

src/runtime/analytics-service.js
elijahsoria: My mistake!

PR #556, opened 2019-05-23
Event Manager / Analytics Client

mborof:


src/runtime/analytics-service.js
```diff
@@ -192,22 +204,31 @@ export class AnalyticsService {
...
   /**
    * @param {!../proto/api_messages.AnalyticsEvent} event
    */
   logEvent(event) {
-    this.lastAction_ = this.start_().then(port => {
-      port.message({'buf': this.createLogRequest_(event).toArray()});
+    this.eventManager_.logEvent({
+      eventType: event,
+      eventOriginator: EventOriginator.SWG_CLIENT,
+      isFromUserAction: null,
```
sohanirao: Why not let internal events to provide isFromUserAction and additionalParams?
Make them optional so existing calls will pass.
/**
 * @param {!../proto/api_messages.AnalyticsEvent} event
 * @param {!boolean=} active
 * @param {!Object=} params
 */
logEvent(event, active, params) {
  const userAction = (active != null): active : null;
  this.eventManager_.logEvent({
    eventType: event,
    eventOriginator: EventOriginator.SWG_CLIENT,
    isFromUserAction: userAction,
    additionalParameters: params || null,
});

src/runtime/entitlements-manager-test.js
```diff
@@ -64,6 +66,7 @@ describes.realWin('EntitlementsManager', {}, env => {
     sandbox.stub(deps, 'storage', () => storage);
     sandbox.stub(deps, 'pageConfig', () => pageConfig);
     sandbox.stub(deps, 'config', () => config);
+    sandbox.stub(deps, 'eventManager', () => eventManager);
     const analyticsService = new AnalyticsService(deps);
```
sohanirao: Line 70-72 can be removed since EntitlementsManager no longer interacts with analytics Service.

src/runtime/entitlements-manager-test.js
```diff
@@ -31,6 +31,7 @@ import {
 import {AnalyticsService} from './analytics-service';
```
sohanirao: I believe this file tests that analytics service was called with logEvent API.

Since it now calls event manager, this verification should happen on a mock of the event manager.

src/runtime/analytics-service-test.js
```diff
@@ -121,7 +144,12 @@ describes.realWin('AnalyticsService', {}, env => {
         const /* {?AnalyticsRequest} */ request =
           new AnalyticsRequest(firstArgument['buf']);
         expect(request.getEvent()).to.deep.equal(AnalyticsEvent.UNKNOWN);
```
sohanirao: maybe check event originator is set?

src/runtime/analytics-service-test.js
```diff
@@ -121,7 +144,12 @@ describes.realWin('AnalyticsService', {}, env => {
...
-        analyticsService.logEvent(AnalyticsEvent.IMPRESSION_PAYWALL);
+        registeredCallback({
+          eventType: AnalyticsEvent.IMPRESSION_PAYWALL,
+          eventOriginator: EventOriginator.SWG_CLIENT,
+          isFromUserAction: null,
```
sohanirao: maybe set userAction to true, set some parameters and check both event originator and user action on the other side?

src/runtime/analytics-service.js
```diff
@@ -72,6 +75,15 @@ export class AnalyticsService {
...
     /** @private {?Promise} */
     this.lastAction_ = null;
+
+    /** @private @const {!../api/client-event-manager-api.ClientEventManagerApi} */
+    this.eventManager_ = deps.eventManager();
+    this.eventManager_.registerEventListener(
+        this.handleClientEvent_.bind(this));
+
+    /** @private @const {!boolean} */
+    this.logPropensityEvents_ = isExperimentOn(deps.win(),
```
sohanirao: again, not enough protection. Maybe add a todo using the same bug
or introduce a method
enableLoggingForPropensity()

Ultimately we'll have to think of config names that the publisher can update. Will add a comment on the bug to work with Shams on closing the loop on this.

mborof: Ok

src/runtime/analytics-service.js
mborof: I actually wasn't aware of the =} syntax for optional params - very helpful to know thank you.

I'm generally preferring to redirect things directly to events manager.  I was effectively planning to do that in a different set of changes to keep things limited in scope.  This just keeps things working the way they were before.  My thinking is that the analytics service is a secret background thing that most people don't need to be aware of.  It is just a listener and nothing should really need to interact with it directly.  Also they can just get eventManager from deps and eventually the analytics service can be removed as something publicly available in deps since things will need to know about it.  It will certainly still need to be instantiated and stored somewhere.

mborof: I did improve the documentation for that and add the deprecated tag to make it clear.

src/runtime/entitlements-manager-test.js
mborof: EntitlementsManager is still telling the analytics service to handle the logging.  So it shouldn't test the internal implementation of analytics service by ensuring analytics service passes the event to event manager.  So I don't think any of this needs to change until/unless we circle back and actually have entitlements manager directly send it's event to event manager.

src/runtime/analytics-service.js
mborof: After digging into entitlements manager again: since it needs to directly call analytics service to set user state anyway I think it's appropriate to keep the helper log function here.  I added isFromUserAction as an optional parameter and removed the deprecated tag.  I think that in general the additionalParameters field existed for Propensity events and that I'd rather keep it out of the helper for now.  I could be convinced that it's useful though if you think its important.

src/runtime/analytics-service-test.js
mborof: Ok

src/runtime/analytics-service-test.js
mborof: Analytics service isn't using the parameters at all right now and so I think testing them isn't necessary.  I did improve tests for originator and userAction

src/runtime/entitlements-manager-test.js
sohanirao: You are right. You are only stubbing the deps so that analytics service gets it.

src/runtime/analytics-service.js
```diff
@@ -192,22 +208,37 @@ export class AnalyticsService {
...
-  logEvent(event) {
-    this.lastAction_ = this.start_().then(port => {
-      port.message({'buf': this.createLogRequest_(event).toArray()});
+  logEvent(eventTypeIn, isFromUserActionIn) {
+    this.eventManager_.logEvent({
+      eventType: eventTypeIn,
+      eventOriginator: EventOriginator.SWG_CLIENT,
+      isFromUserAction: /** @type {?boolean} */
+          (isBoolean(isFromUserActionIn) ? isFromUserActionIn : null),
+      additionalParameters: null,
```
sohanirao: have you decided to not provide additionalParameters as an optional parameter at this time?

src/runtime/analytics-service.js
sohanirao: I think we can always add the additional Parameters if necessary. Its good enough to add the isFromUserAction though since the calling APIs know what event they are sending, they can set this value which is very useful.

PR #562, opened 2019-05-29
Update is_active from propensity event active field

sohanirao:


src/runtime/propensity.js
```diff
@@ -75,12 +75,21 @@ export class Propensity {
     if (userEvent.data && !isObject(userEvent.data)) {
       throw new Error('Event data must be an Object');
     }
-    // TODO(sohanirao, mborof): Idenfity the new interface with event
-    // manager and update the lines below to adhere to that interface
     let paramString = null;
+    if (userEvent.active != null) {
+      if (!userEvent.data) {
+        const jsonData = Object.create(null);
+        jsonData['is_active'] = userEvent.active;
```
jpettitt: Seems a bit complex - how about
```
if (!userEvent.data) {
 userEvent.data = {}
}
userEvent.data.is_active = userEvent.active;
```

sohanirao: The reason I had to do it this way is that userEvent.data gets stringified. So it needs to be a Json Object.
I get the following presubmit failured when I added your code:
Expected : JsonObject|null
Found    : Object{}

        userEvent.data = {};

PR #566, opened 2019-06-03
Do not submit: Deprecate message and onMessage APIs and onResult APIs

sohanirao:


src/components/activities.js
```diff
@@ -0,0 +1,351 @@
...
+   * Accepts the result when ready. The client should verify the activity's
+   * mode, origin, verification and secure channel flags before deciding
+   * whether or not to trust the result.
+   *
+   * Returns the promise that yields when the activity has been completed and
+   * either a result, a cancelation or a failure has been returned.
+   *
+   * @return {!Promise<!web-activities/activity-ports.ActivityResult>}
+   */
+  acceptResult() {}
```
dvoytenko: Wait, i thought this was getting deprecated? Or later? Or should it change the type to return our `Message` interface?

src/components/activities.js
```diff
@@ -0,0 +1,351 @@
...
+  * })
+  *
+  * ports.open('request1', request1Url, '_blank');
+  * ```
+  *
+  * @param {string} requestId
+  * @param {!function(!web-activities/activity-ports.ActivityResult):boolean} verifier
+  * @param {function(!Promise<!Object>)} callback
+  */
+  onResult(requestId, verifier, callback) {
```
dvoytenko: Question: would things be easier for us if we also were to start with two of these? `onResult` with verifier and the `Promise<Message>` and also `onResultDeprecated` with exactly the same old style.

src/components/activities.js
sohanirao: no i didn't change this yet , it would be a follow up change.

src/components/activities.js
sohanirao: i can make it be onResultDeprecated(requestId, verifier, callback) for now. 
and then add onResult(requestId, verifier, messageType, callback)

sohanirao: is that okay?

jpettitt:
Obsolete, closing

PR #570, opened 2019-06-10
Event Manager / Available Sooner / Log Behind Promise

mborof:
This instantiates the event manager sooner in the pipeline and guards the event broadcasting behind the configured runtime promise to ensure events that occur early in the pipeline wait until everything is ready before attempting to log anything.

### DIFF
--- a/src/runtime/client-event-manager.js
+++ b/src/runtime/client-event-manager.js
@@ -60,7 +60,11 @@ function validateEvent(event) {
 
 /** @implements {../api/client-event-manager-api.ClientEventManagerApi} */
 export class ClientEventManager {
-  constructor() {
+  /**
+   *
+   * @param {!Promise} configuredPromise
+   */
+  constructor(configuredPromise) {
     /** @private {!Array<function(!../api/client-event-manager-api.ClientEvent)>} */
     this.listeners_ = [];
 
@@ -69,6 +73,9 @@ export class ClientEventManager {
 
     /** @private {?Promise} */
     this.lastAction_ = null;
+
+    /** @private @const {!Promise} */
+    this.isReadyPromise_ = configuredPromise;
   }
 
   /**
@@ -91,37 +98,24 @@ export class ClientEventManager {
     this.filterers_.push(filterer);
   }
 
-
   /**
    * @overrides
    */
   logEvent(event) {
     validateEvent(event);
     this.lastAction_ = new Promise(resolve => {
-      for (let filterer = 0; filterer < this.filterers_.length; filterer++) {
-        if (this.filterers_[filterer](event) === FilterResult.CANCEL_EVENT) {
-          resolve();
-          return;
+      this.isReadyPromise_.then(() => {
+        for (let filterer = 0; filterer < this.filterers_.length; filterer++) {
+          if (this.filterers_[filterer](event) === FilterResult.CANCEL_EVENT) {
+            resolve();
+            return;
+          }
         }
-      }
-      for (let listener = 0; listener < this.listeners_.length; listener++) {
-        this.listeners_[listener](event);
-      }
-      resolve();
-    });
-  }
-
-  /**
-   * This function exists for the sole purpose of allowing the code to be
-   * presubmitted.  It can be removed once there is code generating a real
-   * event object somewhere.
-   */
-  useValidateEventForCompilationPurposes() {
-    validateEvent({
-      eventType: AnalyticsEvent.UNKNOWN,
-      eventOriginator: EventOriginator.UNKNOWN_CLIENT,
-      isFromUserAction: null,
-      additionalParameters: {},
+        for (let listener = 0; listener < this.listeners_.length; listener++) {
+          this.listeners_[listener](event);
+        }
+        resolve();
+      });
     });
   }
 }
--- a/src/runtime/client-event-manager.js
+++ b/src/runtime/client-event-manager.js
@@ -103,19 +103,16 @@ export class ClientEventManager {
    */
   logEvent(event) {
     validateEvent(event);
-    this.lastAction_ = new Promise(resolve => {
-      this.isReadyPromise_.then(() => {
-        for (let filterer = 0; filterer < this.filterers_.length; filterer++) {
-          if (this.filterers_[filterer](event) === FilterResult.CANCEL_EVENT) {
-            resolve();
-            return;
-          }
+    this.lastAction_ = this.isReadyPromise_.then(() => {
+      for (let filterer = 0; filterer < this.filterers_.length; filterer++) {
+        if (this.filterers_[filterer](event) === FilterResult.CANCEL_EVENT) {
+          return Promise.resolve();
         }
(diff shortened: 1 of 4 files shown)


mborof:
I wanted to leave a brief note explaining that I had to pass in false to this.configured because a large number of tests broke if I didn't.  Passing false gives us the promise and eventually something else kicks off the resolution of it.

src/runtime/client-event-manager.js
```diff
@@ -91,37 +98,24 @@ export class ClientEventManager {
...
    * @overrides
    */
   logEvent(event) {
     validateEvent(event);
     this.lastAction_ = new Promise(resolve => {
-      for (let filterer = 0; filterer < this.filterers_.length; filterer++) {
-        if (this.filterers_[filterer](event) === FilterResult.CANCEL_EVENT) {
-          resolve();
-          return;
+      this.isReadyPromise_.then(() => {
```
sohanirao: can you not do this.lastAction_ = this.isReadyPromise_.then(resolve => {

src/runtime/runtime.js
```diff
@@ -177,6 +177,9 @@ export class Runtime {
     /** @private {?PageConfigResolver} */
     this.pageConfigResolver_ = null;
 
+    /** @private @const {!ClientEventManager} */
+    this.eventManager_ = new ClientEventManager(this.configured_(false));
```
sohanirao: did you mean true?

mborof: If I use true it breaks a bunch of tests.  I think it isn't ready to be called with true at this point.

mborof: And just to note, calling it will false returns the same promise it would have either way - it just doesn't kick of resolution yet.

src/runtime/client-event-manager.js
mborof: The isReadyPromise is set to the configuredruntime promise which means the resolve function won't get passed (I think?); I believe it will be a ConfiguredRuntime object.  It does seem to work if I return Promise.resolve() instead of calling a passed resolve function from a new promise though.  I'm not very comfortable with promises yet and I'm not as sure it will work consistently so you will need to let me know about that.

mborof: I did some testing of this and "resolve" becomes undefined for a large number of circumstances during testing.  I think you can't rely on the resolve function being passed in for some promise circumstances (such as during testing when something just passes in an already resolved promise and for promises where the resolution of the promise calls another promise with a specific value such as ConfiguredRuntime which I'm fairly certain is the case here).

So I think what I'm doing is safer and will work in all circumstances instead of an unknown subset.

sohanirao: sure, here's one approach. why don't you tell me your thoughts on this and if required, we can modify further. Basically we are chaining promises here... 
this.lastAction_ = this.isReadyPromise_.then(() => {
                                  return Promise(resolve => {
                                      // filter events and call resolve()
                                      // OR for events not filtered, call listeneres and then call resolve()
                                   });
What this does is, when isReadyPromsie gets resolved, it creates another promise and assigns it to lastAction_. This should have no difference to your tests.

Another approach.
/**
 * @param {Event} event
 * @return {!Promise<boolean>}
 */
filterEvent(event) {
  let cancelEvent = false;
  for (let filterer = 0; filterer < this.filterers_.length; filterer++) {
        if (this.filterers_[filterer](event) === FilterResult.CANCEL_EVENT) {
          cancelEvent = true;
          break;
       }
  }
  return Promise.resolve(cancelEvent);
}

/**
   * @overrides
   */
  logEvent(event) {
    validateEvent(event);
    this.lastAction_ = this.isReadyPromise_.then(() => {
        return this.filterEvent(event);
      }).then(cancelled => {
        if (!cancelled) {
          // invoke listeners
        });
        return Promise.resolve();
      });
    });
I liked this second version better because this could even more asynchronous. Take a look and see how it may or may not impact your tests at all.

The bottom line is to ensure the code reflects our decision that everything waits on isReadyPromise to get resolved and hence lives in then of that.

mborof: I've adjusted the code to eliminate the extra sub-promise and am returning Promise.resolve() instead.  I lean against having the chain of promises because I suspect that every bit of async probably adds overhead and that we should only do it if necessary or logical.  It feels superfluous to break it up further than it is.

src/runtime/runtime.js
sohanirao: makes sense. I don't think we will remember why we did it this way even a few months down, can you add comments?

src/runtime/runtime.js
```diff
@@ -537,6 +552,10 @@ export class ConfiguredRuntime {
       this.analyticsService_.logEvent(AnalyticsEvent.EVENT_PAYMENT_FAILED);
       this.jserror_.error('Redirect error', error);
     });
+
+    if (doneConstructingResolve != null) {
```
sohanirao: i dont see dontConstrcutingResolve being defined. did you mean to removethis?
I think putting event manager at the top and propensity as soon as possible is good for now.

src/runtime/runtime.js
```diff
@@ -439,6 +442,10 @@ export class Runtime {
       return runtime.getPropensityModule();
     });
   }
+
+  getEventManager() {
```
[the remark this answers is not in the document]
sohanirao: sure

src/runtime/runtime.js
mborof: I forgot to stage that change, thanks for noticing!

src/runtime/runtime.js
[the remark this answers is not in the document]
mborof: Agreed / done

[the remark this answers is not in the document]
sohanirao: sure

src/runtime/client-event-manager.js
```diff
@@ -91,37 +98,21 @@ export class ClientEventManager {
     this.filterers_.push(filterer);
   }
 
-
   /**
    * @overrides
```
elijahsoria: nit: Please add a comment describing the type of the "event" argument.

mborof: I'd previously been told by Dima to take that comment out because it's documented in the API and that's where the documentation should be, rather in than in the override which just matches the API

src/runtime/runtime.js
```diff
@@ -888,6 +903,7 @@ function createPublicRuntime(runtime) {
     attachSmartButton: runtime.attachSmartButton.bind(runtime),
     getPropensityModule: runtime
         .getPropensityModule.bind(runtime),
+    eventManager: runtime.eventManager.bind(runtime),
```
sohanirao: Just noticed this change when reviewing the other change. Is there any reason to expose this outside of SwG? Looks like AMP is already able to send in the event manager to ConfiguredRuntime and we create one internally if required. I suggest removing line 906

mborof: If you say it is not necessary, I believe you.  I'm not sure why those are there it just seemed to fit along with everything else quite well.  Removed.

PR #571, opened 2019-06-11
Smartbox Impression Logging

elijahsoria:
Add impression logging to the SmartBox using AnalyticsService.

### DIFF
--- a/src/runtime/smart-button-api.js
+++ b/src/runtime/smart-button-api.js
@@ -109,7 +109,6 @@ export class SmartSubscriptionButtonApi {
     });
     this.button_.appendChild(this.iframe_);
     if (this.deps_.config().analyticsMode == AnalyticsMode.IMPRESSIONS) {
-      // isFromUserActsion = False
       const analyticsContext = this.deps_.analytics().getContext().toArray();
       this.args_['analyticsContext'] = analyticsContext;
     }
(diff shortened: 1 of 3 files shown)


src/runtime/smart-button-api.js
```diff
@@ -107,6 +108,11 @@ export class SmartSubscriptionButtonApi {
       'width': '100%',
     });
     this.button_.appendChild(this.iframe_);
+    if (this.deps_.config().analyticsMode == AnalyticsMode.IMPRESSIONS) {
+      // isFromUserActsion = False
```
mborof: At this point you could just start logging isFromUserAction (note mispelling in your comment)

elijahsoria: Thanks! I'll remove the comment because I'll actually be setting this server-side.

src/runtime/smart-button-api.js
```diff
@@ -107,6 +108,10 @@ export class SmartSubscriptionButtonApi {
       'width': '100%',
     });
     this.button_.appendChild(this.iframe_);
+    if (this.deps_.config().analyticsMode == AnalyticsMode.IMPRESSIONS) {
```
sohanirao: may not need this. sending context is not so controversial.

elijahsoria: Let's confirm with John and Shams.

elijahsoria: Resolved this offline. Removing if statement.

PR #572, opened 2019-06-14
Event man config

mborof:
Some notes on this:

1) It is branched off the branch where we don't log anything until configured runtime is finished constructing.
2) We are still working on the names for the new config elements.  I suspect at least the _'s should go away since our compiler doesn't like them.


src/runtime/runtime.js
```diff
@@ -888,6 +903,7 @@ function createPublicRuntime(runtime) {
     attachSmartButton: runtime.attachSmartButton.bind(runtime),
     getPropensityModule: runtime
         .getPropensityModule.bind(runtime),
+    eventManager: runtime.eventManager.bind(runtime),
```
sohanirao: based on comments in previous review, this should not be here.

src/runtime/propensity.js
```diff
@@ -28,17 +28,18 @@ export class Propensity {
   /**
    * @param {!Window} win
    * @param {!../model/page-config.PageConfig} pageConfig
-   * @param {!../api/client-event-manager-api.ClientEventManagerApi} eventManager
+   * @param {!./deps.DepsDef} deps
    */
```
sohanirao: Propensity class must remain independent of rest of SwG boot strapping. Hence the API call to enableLoggingSwgEvents() is necessary to get the configuration.

src/runtime/analytics-service.js
```diff
@@ -86,8 +86,11 @@ export class AnalyticsService {
     this.logPropensityExperiment_ = isExperimentOn(deps.win(),
         ExperimentFlags.LOG_PROPENSITY_TO_SWG);
 
+    const config = deps.config();
+    const analyticsConfig = (config && config.analyticsConfig) || {};
```
sohanirao: So since AnalyticsService is constructed in ConfiguredRuntime's constructor, the caller would have to somehow to call subscriptions.configure() on Runtime in order to set the configuration before page configuration is resolved in order for the analytics service to get that configuration before the event manager invokes event listeners, correct?
Is this what you intend?

src/runtime/runtime.js
mborof: Ok

src/runtime/analytics-service.js
mborof: It looked to me like configuration was passed to ConfiguredRuntime (via opt_config).  That may not be how people actually configure SwG though, I don't really know.

My concern with late configuration is that some events may have already occurred meaning that if you turn it on after construction it won't pass the already occurred events along to the clients you tried to enable.

sohanirao: If we don't send events occuring before the configuration was set to send those events, then that is intended behavior. Often getting that config setting to be set might require user consent and therefore, we should honor the config setting. This was one of your initial ideas actually: to check config in the event handler. Do you want to consider that again?

mborof: I thinks it's worth considering all options here, especially in light of a vaguely similar issue I am having with AMP (to be discussed during today's design):

1) Check config each time an event is fired.
2) When setting config, see if they changed that specific setting and enable/disable the sending of events accordingly.  (followup: can the user revoke consent?)

I lean slightly towards option 2 because it is slightly more efficient.

I think it is worth at least considering maintaining a list of all events not sent and sending them if/when the config changes.  This requires option 2 above because something will need to proactively send the saved events as soon as the config changes.

I don't have a clear preference for any of these things but I do wonder: if the user set consent previously will the configuration be available immediately or do some publishers still set it after ConfiguredRuntime is instantiated?

src/runtime/propensity.js
mborof: Are you saying that I can't pass in a deps?  I don't understand the second sentence: if it can't have deps then how can enableLoggingSwgEvents() get the configuration?  Or is something else supposed to call it when the configuration changes?

PR #574, opened 2019-06-19
♻️ ESLint Rules: Allow ExportSpecifier

ChrisAntaki:
Merges in a PR from `amphtml` that allows the ExportSpecifier https://github.com/ampproject/amphtml/pull/22913

This will allow for cleaner re-exporting of symbols.

Example:
```js
export {obj as deprecatedObj} from 'lib';
```

Instead of:
```js
import {obj} from 'lib';
export const deprecatedObj = obj;
```

### DIFF
--- a/build-system/eslint-rules/no-export-side-effect.js
+++ b/build-system/eslint-rules/no-export-side-effect.js
@@ -19,21 +19,22 @@ module.exports = function(context) {
   return {
     ExportNamedDeclaration: function(node) {
       if (node.declaration) {
-        var declaration = node.declaration;
+        const {declaration} = node;
         if (declaration.type === 'VariableDeclaration') {
           declaration.declarations
-              .map(function(declarator) {
-                return declarator.init
-              }).filter(function(init) {
-                return init && /(?:Call|New)Expression/.test(init.type)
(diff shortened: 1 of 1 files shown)


build-system/eslint-rules/no-export-side-effect.js
```diff
@@ -19,22 +19,23 @@ module.exports = function(context) {
...
               });
+            });
         }
-      } else if (node.specifiers) {
-        context.report(node, 'Side-effect linting not implemented');
       }
     },
   };
-};
+};
```
sohanirao: can you add new line and in the description what this PR allows us to do or not do.. some examples maybe and describe the motivation for this change?

sohanirao: I missed that. We are syncing from amphtml, is that all?

ChrisAntaki: Yep, earlier today I made a PR to `amphtml`. It basically allows us to use the `export {obj as deprecatedObj}` pattern.

ChrisAntaki: Just updated the description :)

PR #575, opened 2019-06-19
SwG Button Impressions with EventManager

elijahsoria:
Sets up button impression in the ButtonApi using the EventManager.


package.json
```diff
@@ -135,4 +135,4 @@
     "requiresKeyFile": true,
     "requiresProjectId": true
   }
-}
+}
```
sohanirao: Looks good. Can you remove changes to package.json. You dont need to check this in.

src/runtime/button-api-test.js
```diff
@@ -29,13 +30,17 @@ describes.realWin('ButtonApi', {}, env => {
...
+  let eventManagerMock;
   let buttonApi;
+  let eventManager;
   let handler;
 
   beforeEach(() => {
     win = env.win;
     doc = env.win.document;
-    buttonApi = new ButtonApi(resolveDoc(doc));
+    eventManager = new ClientEventManager(Promise.resolve());
```
mborof: I'm not really clear on test code and best practices so take this with a grain of salt, but you might consider taking the one out of ConfiguredRuntime and then mocking that one rather than creating your own.

elijahsoria: I think in this case it won't matter which EventManager we are using. The tests in this file all use the ButtonApi object created in line 43, which uses the EventManager that is passed to it upon construction. Since runtime.eventMan() will never be used within ButtonApi, we can go with the standalone mock EventMan approach here.

sohanirao: So taking eventManager() out of ConfiguredRuntime is the typical pattern we follow in our code. But in this case, since ConfiguredRuntime may not exist when ButtonApi is constructed or used, we cannot rely on that. That is the whole reason event manager had to event be passed into this Class constructor, yes? So I agree with Elijah, it doesn't matter where the event manager came from as long as its behavior is mocked for test purposes.

PR #580, opened 2019-06-25
Event manager config - DO NOT SUBMIT UNTIL WE AGREE ON CONFIG NAMES

mborof:


mborof:
I still need to re-write the tests on this, but I wanted Sohani's take on the approach before I do that.

src/runtime/propensity-server.js
```diff
@@ -56,8 +59,8 @@ export class PropensityServer {
     this.logSwgEventsExperiment_ = isExperimentOn(win,
         ExperimentFlags.LOG_SWG_TO_PROPENSITY);
 
-    /** @private {!boolean} */
-    this.logSwgEventsConfig_ = false;
+    /** @private {!function():?boolean} */
+    this.logSwgEventsCallback_ = logSwgEventsCallback || (() => false);
```
sohanirao: Do you want a function that returns true or false or config object?
Why not get entire config returned in the callback and then you can access that exact field

mborof: I guess I'm not clear on what propensity can rely on.  It can rely on a properly formed config object but not a depsdef?

mborof: I think using a boolean keeps Propensity's dependence on SwG as a whole minimal.

src/runtime/runtime.js
```diff
@@ -492,7 +492,8 @@ export class ConfiguredRuntime {
 
     /** @private @const {!Propensity} */
     this.propensityModule_ = new Propensity(this.win_,
-      this.pageConfig_, this.eventManager_);
+      this.pageConfig_, this.eventManager_,
+      () => this.config().analyticsConfig.enable_propensity_in_swg || false);
```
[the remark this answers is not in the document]
mborof: Makes sense

PR #581, opened 2019-06-26
ConfiguredRuntime now accepts the configuredPromise instead of event manager

mborof:
AMP wasn't allowed to declare a new event manager, so this minor tweak allows AMP to pass in a configuration promise to ConfiguredRuntime that it will resolve after it is done registering as a filterer.

SwG will also work by passing in its configuration promise to ConfiguredRuntime.  This ensures that no logging is done until it is ready. 

A helper method of logging against Runtime (before event manager has been declared) has also been added for the button that needs to log immediately.


src/runtime/runtime.js
```diff
@@ -501,8 +505,7 @@ export class ConfiguredRuntime {
     this.jserror_ = new JsError(this.doc_);
 
     /** @private @const {!Fetcher} */
-    this.fetcher_ = opt_integr && opt_integr.fetcher ||
-        new XhrFetcher(this.win_);
+    this.fetcher_ = opt_integr.fetcher || new XhrFetcher(this.win_);
```
sohanirao: what happens if opt_integr is undefined? why remove the opt_integr && clause?

src/runtime/runtime-test.js
```diff
@@ -1515,4 +1486,58 @@ describes.realWin('ConfiguredRuntime', {}, env => {
...
+
+    const event = {
+      eventType: AnalyticsEvent.IMPRESSION_PAYWALL,
+      eventOriginator: EventOriginator.SWG_CLIENT,
+      isFromUserAction: true,
+      additionalParameters: null,
+    };
+
+    beforeEach(() => {
+      activityResultCallbacks = {};
```
sohanirao: is this getting used?

src/runtime/runtime-test.js
```diff
@@ -1515,4 +1486,58 @@ describes.realWin('ConfiguredRuntime', {}, env => {
   it('should return events manager', () => {
     expect(runtime.eventManager() instanceof ClientEventManager).to.be.true;
   });
+
+  describe('configPromise', () => {
```
sohanirao: one more test that could be added here, if you think useful, is to demonstrate how one may attach a filterer before resolving a promise and as a result have the filterer do its magic on events that are queued that should be filtered.

For example.

stub filterer registration api in event manager to save filterer 
stub logEvent method to count events after the saved filterer is invoked
create configured runtime with a promise
get event manager from it and attach filterer
queue 2 events
check event count is 1

Often test code serves as documentation on use cases supported by an API surface. An example of this sort is a good reminder as to why did it this way documented by way of test code. IMHO

src/runtime/runtime-test.js
mborof: There is an error thrown if I do not include this statement.

src/runtime/runtime-test.js
mborof: Good idea, done

mborof:
I've redone all the tests so I could avoid the odd set to {} in the new tests I added.  I did it the way I suggested I would (re-sandboxing everything) instead of the way you had suggested (re-wiring the way every individual test got its configuredRuntime object).  I think my approach is better because it was just moving a bunch of code around instead of actually diving into all individual tests and figuring out how they worked to re-write them.

I think John has probably seen the upstream on these changes since he reviewed my AMP branch Friday which uses the configuration promise I created here.

PR #582, opened 2019-06-27
PropensityScore can be bucketed

sohanirao:


src/api/propensity-api.js
```diff
@@ -214,9 +214,12 @@ export const PropensityType = {
  *       If there are any errors which prevented the server from
  *       generating and providing a valid score, this field will have a
  *       string describing why score was not available.
+ *  - bucketed: Optional. When a numeric score is provided, this field
+ *       will be true or false to indicate if this score is bucketed or not.
  *
  *  @typedef {{
- *    result: (number|string)
+ *    result: (number|string),
```
jpettitt: Not a big fan of this - it leaves publishers having to parse a string.  Can we just return

result: {?number}
bucketedResult: {?Array}

sohanirao: no no, the string is the error case. otherwise its a number. publishers don't have to parse a string. Can I clarify the comments to communicate this or would you prefer to not have them be a one-of?
regarding bucketed result: they will provide one or the other (a raw score or bucket score). it wont be an array and we will never have both at the same time.

one response:
result: (number)
bucketed: true (means number is bucket score ) or false (means raw score)

another response:
result: error_message string

sohanirao: Here's an alternate proposal
/**
 * The score itself
 * Properties:
 * - product: Required. Indicates the publication_id:product_id for which the score is provided
 * - raw: Optional. When the score is a number in the range [0-100]
 * - bucketed: Optional. When the score is a number indicate the bucket in which the score lands [0-10]
 * - error_message: Optional. When no score is avaialble, an error message describing why no score is provided.
 *
 * @typedef {{
 *   product: string,
 *   raw: ?number,
 *   bucketed: ?number,
 *   error_message: ?string,
 * }}
 */
export let Score;

/**
 * The Body field of the Propensity Score.
 * Properties:
 * - scores: Optional, an array of scores
 * - error: Optional, string describing why no scores were provided
 *
 *  @typedef {{
 *    scores: ?Array<Score>,
 *    error: ?string,
 * }}
 */
export let Body;

Thoughts?

jpettitt: Ah ok.  

So it's a bucket # between [0..n] - how does the publisher know what n is given that it may change per the design doc?

Also I hate overloading results with error messages, it's not a good pattern (my basd I should have caught this earlier).

Let's also add a `success: {boolean}` and and `error: {string}` that's present if success is false.  We should deprecate the string and make it `result: {?number}`

PR #583, opened 2019-07-11
Adds gulp fix task

ChrisAntaki:
Adds gulp fix task, which formats code using prettier.

```sh
# For example, this command fixed gulpfile.js
gulp fix --files=gulpfile.js
```


ChrisAntaki:
Needs more work

PR #584, opened 2019-07-11
SwG Button Impression Logging

elijahsoria:
Implements SwG Button impression logging with the Promise-based logEvent function in Runtime. Removes logging from getEntitlements call. Adds ability to log Paywall/Offers impressions from create/attachButton and createSmartBox.

### DIFF
--- a/src/api/subscriptions.js
+++ b/src/api/subscriptions.js
@@ -222,11 +222,11 @@ export class Subscriptions {
    *
    * @param {!ButtonOptions|function()} optionsOrCallback
    * @param {function()=} opt_callback
-   * @param {?boolean} optOnPaywall
-   * @param {?boolean} optOnSubscriptionsPage
+   * @param {?boolean=} opt_onPaywall
+   * @param {?boolean=} opt_onSubscriptionsPage
    * @return {!Element}
    */
-  createButton(optionsOrCallback, opt_callback, optOnPaywall, optOnSubscriptionsPage) {}
+  createButton(optionsOrCallback, opt_callback, opt_onPaywall, opt_onSubscriptionsPage) {}
 
   /**
    * Attaches the SwG button style and the provided callback to an existing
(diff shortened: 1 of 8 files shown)


src/api/subscriptions.js
```diff
@@ -222,9 +222,11 @@ export class Subscriptions {
    *
    * @param {!ButtonOptions|function()} optionsOrCallback
    * @param {function()=} opt_callback
+   * @param {?boolean=} opt_onPaywall
+   * @param {?boolean=} opt_onSubscriptionsPage
    * @return {!Element}
    */
-  createButton(optionsOrCallback, opt_callback) {}
+  createButton(optionsOrCallback, opt_callback, opt_onPaywall, opt_onSubscriptionsPage) {}
```
mborof: It might be better to create an enum value and pass that in.  That way when there are other options later it is easily expandable.  As 2 bool flags passed in I additionally have to wonder what you should do if you pass in both as true.

src/runtime/button-api.js
```diff
@@ -60,10 +75,14 @@ export class ButtonApi {
 
   /**
    * @param {!../model/doc.Doc} doc
+   * @param {!function(!../api/client-event-manager-api.ClientEvent)} logEventFunc
    */
-  constructor(doc) {
+  constructor(doc, logEventFunc) {
```
mborof: NIT: I've seen "function" abbreviated as Fun many many times. You can leave it func if you want though.

src/runtime/button-api-test.js
```diff
@@ -183,6 +186,63 @@ describes.realWin('ButtonApi', {}, env => {
...
+        CreateButtonRelatedClientEvent(
+            AnalyticsEvent.IMPRESSION_SUBSCRIBE_BUTTON)).once();
+    eventManagerMock.expects('logEvent').withExactArgs(
+        CreateButtonRelatedClientEvent(
+            AnalyticsEvent.IMPRESSION_PAYWALL)).once();
+    const button = doc.createElement('button');
+    buttonApi.attach(button, {}, handler, true);
+  });
+
+  it('should log offers impression on create.', () => {
```
mborof: I'm just wondering why it logs offers impression when you create it.  Don't they have to click it to view offers?  Or if it has a default offer there is no offers impression?

src/runtime/button-api.js
```diff
@@ -51,6 +52,20 @@ const TITLE_LANG_MAP = {
 };
 
 
+/**
+   * @param {!AnalyticsEvent} analyticsEvent
+   * @return {!../api/client-event-manager-api.ClientEvent}
+   */
+export function CreateButtonRelatedClientEvent(analyticsEvent) {
```
mborof: Seems like a useful helper function and not actually related to a button at all.  Perhaps this should be called createSwgClientEvent(eventType, isFromUserAction)

With isFromUserAction optional and defaulting to false.

I'd also move it to a more centralized location.

src/runtime/runtime.js
```diff
@@ -863,6 +875,13 @@ export class ConfiguredRuntime {
   eventManager() {
     return this.eventManager_;
   }
+
+  /**
+   * @param {!../api/client-event-manager-api.ClientEvent} event
+   */
+  logEvent(event) {
```
sohanirao: still not convinced we need this new API

PR #586, opened 2019-07-15
Smart box message proto added

sohanirao:


sohanirao:
SmartBoxMessage might not be the right name for it, but its a communication from iframe to client indicating a click was done. This is to replace the JSON data 'clicked': true with a protobuf.

PR #588, opened 2019-07-16
Logger api

mborof:
This implements a logger API and makes changes to Propensity API so it be be integrated with logger later.


src/api/logger-api.js
```diff
@@ -0,0 +1,252 @@
+/* eslint-disable no-unused-vars */
```
jpettitt: Why?

src/api/propensity-api.js
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable no-unused-vars */
```
jpettitt: Why?

mborof: Looks like its just my add-on to visual studio code that's complaining.  Removed.

src/api/logger-api.js
mborof: Looks like its just an add-on complaining.  Removed.

src/api/propensity-api.js
```diff
@@ -263,38 +83,21 @@ export let PropensityScore;
...
 export let PropensityEvent;
 
+export const Event = LoggerApi.Event;
+export const SubscriptionState = LoggerApi.SubscriptionState;
+
 /**
+ * @extends {LoggerApi.LoggerApi}
  * @interface
  */
 export class PropensityApi {
```
sohanirao: Don't have to delete lines 277:to 297, you can simply at @override to lines 288 and 295

src/api/propensity-api.js
```diff
@@ -13,187 +13,7 @@
...
-   *   active: true,
-   *   data: {
-   *     'event_name': 'social_share',
-   *     'platform_used': 'whatsapp'
-   *   }
-   *  }
-   */
-  EVENT_CUSTOM: 'custom',
-}
+import * as LoggerApi from './logger-api';
```
sohanirao: Lets not delete this here, it serves as Propensity documentation

src/api/propensity-api.js
```diff
@@ -263,38 +83,21 @@ export let PropensityScore;
...
  *  @typedef {{
- *    name: string,
+ *    name: !Event,
  *    active: boolean,
  *    data: ?JsonObject,
  * }}
  */
 export let PropensityEvent;
 
+export const Event = LoggerApi.Event;
```
sohanirao: You would do the opposite. 
Like I have commented on the whole PR.

sohanirao:
Why not do it the other way around?

export PublisherEvent = PropensityApi.Event;
export SubscriptionState = PropensityApi.SubscriptionState;

mborof:
1) Logger API must have event and SubscriptionState (currently defined in Propensity) in order to create functions that accept those enums.
2) Propensity must rely on Logger API so it can implement the interface declared in it

So if I do not move the primary definitions of Event and SubscriptionState to Logger API it creates a circular reference.

Redeclaring Event and SubscriptionState in Propensity is simply done so no one needs to change references to continue using SwG.

src/api/propensity-api.js
mborof: Logger API must have event and SubscriptionState (currently defined in Propensity) in order to create functions that accept those enums.
Propensity must rely on Logger API so it can implement the interface declared in it
So if I do not move the primary definitions of Event and SubscriptionState to Logger API it creates a circular reference.

Redeclaring Event and SubscriptionState in Propensity is simply done so no one needs to change references to continue using SwG.

src/api/propensity-api.js
mborof: Why re-write the same exact thing with an override?

sohanirao: two reasons: 
1. completeness, when we extend interfaces, overriding existing interfaces is a common practice and it documents the new interface fully
2. looks like in this case even that might not work since, the argument to sendEvent() and sendSubscriptionState() now uses PropensityApi which is a clone to LoggerAPI but not the same. Does the closure allow you to send PropensityApi.Event when the method its overriding wants LoggerApi.Event?

src/api/propensity-api.js
sohanirao: I didn't mean create circular reference, I mean keep Propensity and Logger referes to Propensity API.

mborof:
But Propensity API needs to reference Logger API so it can implement its
interface.

src/api/propensity-api.js
mborof: Yes because they are identical.  If I change one then it breaks though.

src/api/propensity-api.js
mborof: Sorry, I left the word API out above.  Let me try to explain why you are asking for a circular reference again:  

1) PropensityAPI MUST reference LoggerAPI to implement its interface.  This cannot change.
2) LoggerAPI needs the enums currently defined in PropensityAPI because they are part of the interface it defines.  

If I do not move those enums out of Propensity API it is a circular reference.

src/api/propensity-api.js
mborof: I've added those functions back now.

mborof:
I've done 2 things:
  1) Removed the repeating of the logging definitions from propensity-api.  It is better without them.
  2) Added comments telling publishers to look in logger-api for the logging enums and object defs.

src/api/propensity-api.js
```diff
@@ -287,38 +110,28 @@ export let PropensityScore;
  *         enum listed in the Event enum above.
  *
  *  @typedef {{
- *    name: string,
+ *    name: !Event,
```
sohanirao: this might be a problem since publishers dont have access to enum types. a good way to confirm is try using this api is scenic demo.

PR #590, opened 2019-07-17
Move ACTION_SUBSCRIBE logging request to button click

elijahsoria:
Currently, the ACTION_SUBSCRIBE logging event is being logged on pay-flow start. This PR moves that logging event to a callback for each SwG button click. The previously logging event is changed to ACTION_PAYMENT_FLOW_STARTED.

### DIFF
--- a/src/runtime/button-api.js
+++ b/src/runtime/button-api.js
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+import {AnalyticsEvent, EventOriginator} from '../proto/api_messages';
 import {createElement} from '../utils/dom';
 import {msg} from '../utils/i18n';
 import {SmartSubscriptionButtonApi, Theme} from './smart-button-api';
@@ -50,6 +51,17 @@ const TITLE_LANG_MAP = {
   'zh-tw': '透過 Google 訂閱',
 };
 
+/**
+ * The ClientEvent object to log on button clicks.
+ */
+/** @type {!../api/client-event-manager-api.ClientEvent} */
+export const BUTTON_CLICK_EVENT = {
+  eventType: AnalyticsEvent.ACTION_SUBSCRIBE,
+  eventOriginator: EventOriginator.SWG_CLIENT,
+  isFromUserAction: true,
+  additionalParameters: null,
+}
+
 
 /**
  * The button stylesheet can be found in the `/assets/swg-button.css`.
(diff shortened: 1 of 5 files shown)


elijahsoria:
Shams seems to be okay with this change as long as we tell Carter before the change is submitted and he is okay with it as well. If we create a new event, we will need to propagate that event to all of our SQL queries, which would also be problematic since folks running analysis will have to change their analysis methods depending on the dataset.

It might be good to create a new log column that specifies the version of the logging proto used. That way, folks can tell the accuracy of the data by the version number associated to the data they are looking at. 

What about asking Carter to see what he prefers?

mborof:
You can ask Carter which he prefers.  The problem right now is that you might misinterpret ACTION_SUBSCRIBE to mean payment flow started when all they did was look at the offers screen and cancel.  And if you just stop interpreting ACTION_SUBSCRIBE to mean payment_flow_start than any log event recorded before the publisher updates their code doesn't indicate the user started the payment flow.

src/runtime/runtime.js
```diff
@@ -863,6 +863,13 @@ export class ConfiguredRuntime {
   eventManager() {
     return this.eventManager_;
   }
+
+  /**
+   * @param {!../api/client-event-manager-api.ClientEvent} event
+   */
+  logEvent(event) {
+    this.eventManager_.logEvent(event);
```
sohanirao: is there anyway to do this without exposing this? Since eventManager() is already exposed? Why can't the caller do runtime.eventManager().logEvent(event)?

src/runtime/runtime.js
```diff
@@ -538,7 +538,7 @@ export class ConfiguredRuntime {
     this.offersApi_ = new OffersApi(this.pageConfig_, this.fetcher_);
 
     /** @private @const {!ButtonApi} */
-    this.buttonApi_ = new ButtonApi(this.doc_);
+    this.buttonApi_ = new ButtonApi(this.doc_, this.logEvent.bind(this));
```
sohanirao: why don't we sent runtime or the event manager here?

src/runtime/button-api.js
```diff
@@ -50,6 +51,17 @@ const TITLE_LANG_MAP = {
   'zh-tw': '透過 Google 訂閱',
 };
 
+/**
+ * The ClientEvent object to log on button clicks.
+ */
+/** @type {!../api/client-event-manager-api.ClientEvent} */
+export const BUTTON_CLICK_EVENT = {
+  eventType: AnalyticsEvent.ACTION_SUBSCRIBE,
```
sohanirao: I guess once you add a new event name, you are going to add this?

src/runtime/button-api.js
```diff
@@ -60,10 +72,14 @@ export class ButtonApi {
 
   /**
    * @param {!../model/doc.Doc} doc
+   * @param {!function(!../api/client-event-manager-api.ClientEvent)} logEventFun
    */
-  constructor(doc) {
+  constructor(doc, logEventFun) {
```
sohanirao: if constructor was runtime or event manager, then the click event listener would just send event to it directly.

PR #593, opened 2019-07-19
Update Propensity event references to Logger Event (internally)

mborof:
My methodology for this was:

1) comment out the Event and SubscriptionState enums in propensity-api
2) Get everything to compile by ensuring things reference logger-api instead

I also renamed the propensity-type-mapping file to be event-type-mapping and made name changes related to generalizing the logging of publishers events.  This is a non-functional change to clean up code.


mborof:
Also, I branched this off the logger-api pull request so it has all those changes in it already.

elijahsoria:
Only comment would be that "Event" is generic. It might be better to use a more specific name, but I understand that you don't want to use "PropensityEvent" or "LoggerEvent" due to both classes using the same thing. Up to you.

PR #595, opened 2019-07-19
Use Runtime's EventManager instead of AnalyticsService

elijahsoria:
Uses EventManager's logEvent function in all places where AnalyticsService's logEvent function originally was. Removes logEvent function from Runtime.


src/runtime/runtime.js
```diff
@@ -863,6 +863,12 @@ export class ConfiguredRuntime {
   eventManager() {
     return this.eventManager_;
   }
+
```
sohanirao: wasnt this already there

mborof: Hold off on reviewing this.  It will use logger instead

mborof:
Just noting that we've decided that this code should rely on logger instead of a logEvent function on the runtime object.

src/runtime/runtime.js
elijahsoria: It is ready for review. We decided offline to use EventManager's logEvent method instead of using LoggerApi or Runtime's logEvent function to (1) reduce the number of methods that do logging code and (2) to delineate between external and internal logging by saying "external loggers use LoggerApi, internal loggers use EventManager".

src/runtime/entitlements-manager-test.js
```diff
@@ -39,15 +40,16 @@ describes.realWin('EntitlementsManager', {}, env => {
...
   let deps;
   let encryptedDocumentKey;
-  let eventManager;
 
   beforeEach(() => {
     win = env.win;
     pageConfig = new PageConfig('pub1:label1');
     fetcher = new XhrFetcher(win);
-    eventManager = new ClientEventManager();
+    const eventManager = new ClientEventManager();
```
mborof: I'm slightly surprised this works through testing without passing Promise.resolve() to ClientEventManager as the first parameter.  For robustness it would be best to do so.

src/runtime/client-event-manager.js
```diff
@@ -29,6 +29,28 @@ function createEventErrorMessage(valueName, value) {
...
 
+/**
+ * Utility function to create a ClientEvent proto.
+ * @param {!AnalyticsEvent} eventType
+ * @param {!EventOriginator} eventOriginator
+ * @param {?boolean} isFromUserAction
+ * @param {?Object} additionalParameters
+ * @returns {!../api/client-event-manager-api.ClientEvent}
+ */
+export function CreateClientEvent(
```
mborof: Since this function is for internal use, I would suggest one of the following:

1) Make eventOriginator the last parameter and make it optional.  Default to SWG_CLIENT
2) Create a second createSwgEvent where originator is not passed in and have it call createClientEvent with SWG_CLIENT.  I would expect almost nothing to use this without it being SWG_CLIENT excluding a few lines of code in the logger class.

also, do our utility functions generally start with a lower case or capital letter?  I feel like generally it is pascalCase but maybe I'm wrong.

mborof: I'd also make isFromUserAction and additionalParameters optional.  The syntax for this is:

{boolean?} which means it can be undefined / not even passed in.

src/runtime/entitlements-manager-test.js
elijahsoria: I believe it works because I am mocking out calls to logEvent, which is the function in ClientEventManager that uses the promise. In any case, I'll add it for robustness.

src/runtime/runtime-test.js
```diff
@@ -467,19 +467,6 @@ describes.realWin('Runtime', {}, env => {
         }
       );
     });
-
```
mborof: why remove these?

src/runtime/runtime-test.js
```diff
@@ -1643,19 +1630,5 @@ describes.realWin('ConfiguredRuntime', {}, env => {
     it('should return events manager', () => {
       expect(runtime.eventManager() instanceof ClientEventManager).to.be.true;
     });
-
-    it('should let event manager send events without a promise', () => {
```
mborof: Why remove these?

src/runtime/runtime.js
```diff
@@ -440,16 +440,6 @@ export class Runtime {
       return runtime.getPropensityModule();
     });
   }
-
```
mborof: I'm removing this in a different PR.  Just FYI.  I don't really care if you do also.

src/runtime/runtime-test.js
elijahsoria: I am removing Runtime's logEvent, so this test is no longer valid.

src/runtime/runtime-test.js
elijahsoria: This one's an accident.

src/runtime/runtime.js
elijahsoria: Ack. I'll remove it as well.

elijahsoria: I'll remove it as well.

src/runtime/client-event-manager.js
elijahsoria: Went with making eventOriginator, isFromUserAction, and additionalParameters all optional by marking them as having default values.

mborof: I think originator should be last because it is the least likely to be used.  isFromUserAction I would actually like to see everything use so it should be second.  additionalParameters will probably not be used internally (much) so it should be third.

elijahsoria: I like putting isFromUserAction second but left eventOriginator as third position because I don't like the idea of putting an additional param before an enum.

mborof: Sounds fine to me

src/runtime/entitlements-manager-test.js
```diff
@@ -16,6 +16,7 @@
 
 import {Callbacks} from './callbacks';
 import {DepsDef} from './deps';
+import {createClientEvent} from './client-event-manager.js';
```
mborof: It's weird that this has .js and nothing else does.  It's also weird that there's the same file here twice.

src/runtime/entitlements-manager-test.js
```diff
@@ -25,7 +26,7 @@ import {XhrFetcher} from './fetcher';
 import {base64UrlEncodeFromBytes, utf8EncodeSync} from '../utils/bytes';
 import {AnalyticsService} from './analytics-service';
 import {defaultConfig, AnalyticsMode} from '../api/subscriptions';
-import {AnalyticsEvent} from '../proto/api_messages';
+import {AnalyticsEvent, EventOriginator} from '../proto/api_messages';
 import {ClientEventManager} from './client-event-manager';
```
mborof: The other instance of that include

src/runtime/pay-flow-test.js
```diff
@@ -15,11 +15,12 @@
  */
 
 import {ActivityPort} from '../components/activities';
-import {AnalyticsEvent} from '../proto/api_messages';
+import {AnalyticsEvent, EventOriginator} from '../proto/api_messages';
 import {ConfiguredRuntime} from './runtime';
 import {Entitlements} from '../api/entitlements';
 import {ProductType, ReplaceSkuProrationMode} from '../api/subscriptions';
 import {PageConfig} from '../model/page-config';
+import {createClientEvent} from './client-event-manager.js';
```
mborof: just noting the .js

mborof:
Elijah and I talked about potentially moving the createClientEvent function into a more centralized helper file location.  I wasn't sure what the appropriate spot would be and he is going to look into that.

I feel comfortable approving the function API and it's usage but not comfortable approving whether or not the isUserAction flag is accurate for all cases where he is now setting it so I think someone who knows the code base better than I should look at that.

src/runtime/client-event-manager.js
```diff
@@ -29,6 +29,28 @@ function createEventErrorMessage(valueName, value) {
...
 
+/**
+ * Utility function to create a ClientEvent proto.
+ * @param {!AnalyticsEvent} eventType
+ * @param {?boolean=} isFromUserAction
+ * @param {!EventOriginator=} eventOriginator
+ * @param {?Object=} additionalParameters
+ * @returns {!../api/client-event-manager-api.ClientEvent}
+ */
+export function createClientEvent(
```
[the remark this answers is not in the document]
elijahsoria: Good point. In talking with Matt, we came up with the plan of adding a "logSwgEvent" method to the EventManager that accepts AnalyticsEvent, isFromUserAction, and additionParams. That way, we can just call eventManager().logSwgEvent directly instead of having this wrapper function, and since EventManager is only being used by the SWG client, we don't have to expose EventOriginator.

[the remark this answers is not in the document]
sohanirao: Well it shouldn't be necessary to add a new API to default one parameter. Again logEvent() should be able to handle it. Unless I"m missing something.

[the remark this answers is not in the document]
elijahsoria: I'm a bit confused. Are you proposing to alter logEvent's arguments such that it accepts (AnalyticsEvent, isFromUserAction, EventOriginator, and additionalParams) instead of accepting a ClientEvent object?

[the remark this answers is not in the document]
sohanirao: that's one way. 
// logEvent(event, active, originator, jsonParams)
OR
// If ClientEvent must be fully formed before sending to EM, then define default params instead of a helper function
DEFAULT_ACTIVE = false
DEFAULT_ORIGINATOR = SWG_CLIENT
DEFAULT_PARAMS = null
so the callers can construct a default event using these instead of invoking a helper function
logEvent({AnalyticsEvent.IMPRESSION_PAYWALL, ClientEventManager.DEFAULT_ACTIVE, ClientEventManager.DEFAULT_ORIGINATOR, ClientEventManager.DEFAULT_PARAMS})

OR
// allow logEvent to just accept AnalyticsEvent as an optional argument
/**
 * Either event or optEvent should be provided by the caller
 * @param {?ClientEvent} event
 * @param {AnalyticsEvent=} optEvent
 */
logEvent(clientEvent, optEvent) {
  let event = clientEvent;
  if (!event) {
    event = {optEvent, DEFAULT_ACTIVE, DEFAULT_ORIGINATOR, DEFAULT_PARAMS}
  }
  // do other stuff.
}
The caller would then do
logEvent(null, AnalyticsEvent.IMPRESSION_PAYWALL);

OR
// make ClientEvent a class with no helper functions but public fields
export class ClientEvent {
  /**
   * @param {!ClientEventType} type
   */
  constructor(type) {
    /** {!ClientEventType} */
    this.eventType = null;
    /** {boolean} */
    this.isFromUserAction = false;
    /** {EventOriginator} */
    this.eventOriginator = SWG_CLIENT;
    /** {?JsonObject} */
    this.additionalParams = null;
  }
}

So callers may do:
logEvent(new ClientEvent(AnalyticsEvent.IMPRESSION_PAYWALL));
OR do this instead:
 const event = new ClientEvent(AnalyticsEvent.IMPRESSION_PAYWALL);
  event.isFromUserAction = true;
  event.eventOriginator = PUBLISHER_CLIENT
  event.additionalParams = myJsonParams;
  logEvent(event);

All of the above are some API change or other. But logEvent() API IMHO is the one that needs fixing, not adding new helpers or APIs to solve the problem.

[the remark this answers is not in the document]
elijahsoria: My issue with changing the logEvent method or the ClientEvent object is that AMP is already using this API as is. Changing the public API would make it so that we would have to change both the SwG code and the AMP code concurrently, which I believe to be out of the scope of this PR. Although this could be done, I think it would be overkill to do so because this change would make the API only slightly more convenient and I would guess that the only users of this API will be in SwG anyways as EventManager will be getting wrapped by LoggerApi for publisher use.

The reason I like the logSwgEvent method is because it finds a middle ground between not changing a public API and making it easier to use inside of SwG. By adding this method to the implementation and not the API, we can use it inside of Runtime--which is where EventManager will mostly be used--and hide it from the public. This makes it so I can just use one function call (logSwgEvent(...) instead of two (logEvent(createClientEvent(...), and we won't have to change any AMP code.

What do you think?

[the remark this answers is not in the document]
sohanirao: I agree that logSwGEvent() or createClientEvent() is a good solution without having to change API. Because internally we can get away with a helper or alternate API like logSwgEvent. But the need for this internally indicates that maybe even AMP code is finding this API difficult to use? If that's the case, then we are better off fixing this sooner than later, before too many callers are calling it. Agree that whatever we decide is out of scope of this PR we just need to document our decision and then follow up with another change if needed. I really feel like decision should be based on ease of use in AMP. I have approved this PR so that you can document your final decision and merge it.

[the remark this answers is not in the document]
elijahsoria: Thank you for your approval. I will add the logSwgEvent and plan on leaving the logEvent/EventManager API as is as I don't think it is worth the effort given that we are behind on the logging project right now.

PR #597, opened 2019-07-23
Logger - PR 1 - send subscription state via event manager

mborof:
Sends subscription state changes via event manager so that setSubscriptionState doesn't need to directly reference propensity-server.


mborof:
We discussed what to do about errors and we decided they should get logged to the console.  This is necessary to decouple sendSubscriptionState from PropensityServer so that we can have logger communicate it to PS without actually having PS passed to it (as you had requested)

sohanirao:
My request is to simply hold off on this change. It has not been concluded that publishers dont need programmatic errors. We can come back to this. Logger can still communicate by sending event, yes. But this decoupling is not necessary since getPropensity() is only available in PropensityServer.

mborof:
I will hold off on that for now

src/runtime/propensity-test.js
```diff
@@ -151,16 +152,10 @@ describes.realWin('Propensity', {}, env => {
 
   it('should validate events sent to it and set appropriate defaults', () => {
     let hasError;
-    let receivedEvent;
-
```
sohanirao: i dont understand changes to this test file.

sohanirao:
Changes to propensity server look good. Don't understand changes to propensity-test.

mborof:
It was harder to undo the changes to PS than to just make a new one.
Please check this one now (612)

https://github.com/subscriptions-project/swg-js/pull/612/files

PR #599, opened 2019-07-24
Modify offersflow use oldsku

Kestrad:
OffersFlow class should now be able to receive an oldSku and pass it
on to the showOffers iframe. Implemented test to reflect new
functionality.


examples/sample-pub/views/article.html
```diff
@@ -70,7 +70,7 @@
          }
       },
       "description": "A most wonderful article",
-      "isAccessibleForFree": "False",
+      "isAccessibleForFree": "True",
```
jpettitt: Perhaps make a separate example for this rather than changing the default one?

src/runtime/offers-flow.js
```diff
@@ -51,20 +51,63 @@ export class OffersFlow {
...
+      'skus': (options && options.skus) || null,
+      'isClosable': isClosable,
+    };
+
+    if (options && options.oldSku) {
+      feArgsObj['oldSku'] = options.oldSku;
+    }
+
+    if (feArgsObj['oldSku']) {
+      if (!feArgsObj['skus']) {
```
jpettitt: use assert() from log.js for these sort of tests.

src/runtime/offers-flow.js
```diff
@@ -51,20 +51,63 @@ export class OffersFlow {
...
+        console.error('Need a sku list if old sku is provided!');
+        return;
+      }
+      // remove old sku from offers if in list
+      let skuList = feArgsObj['skus'];
+      const oldSku = feArgsObj['oldSku'];
+      skuList = skuList.filter(sku => sku !== oldSku);
+      if (skuList.length > 0) {
```
jpettitt: likewise assert();

examples/sample-pub/views/article.html
Kestrad: Sorry, I forgot that I could set a .gitignore file and accidentally committed changes to those files I made for test purposes only. I've reset them back to the correct versions.

src/runtime/offers-flow.js
Kestrad: Sorry, what is log.js?

src/runtime/runtime.js
```diff
@@ -709,6 +721,29 @@ export class ConfiguredRuntime {
   /** @override */
   showOffers(opt_options) {
     return this.documentParsed_.then(() => {
+      if (opt_options['oldSku']) {
+        console.error(
```
Kestrad: Is this the sort place where I should use assert as well, or is the current structure fine?

src/api/subscriptions.js
```diff
@@ -124,9 +130,15 @@ export class Subscriptions {
 
   /**
    * Starts subscription purchase flow.
-   * @param {string|SubscriptionRequest} skuOrSubscriptionRequest
+   * @param {string} sku
+   */
+  subscribe(sku) {}
```
dparikh: Why subscribe() API is being changed?
This API is already being used and could break existing implementation.

src/runtime/offers-flow.js
```diff
@@ -51,20 +52,54 @@ export class OffersFlow {
...
+      'isClosable': isClosable,
+    };
+
+    if (options && options.oldSku) {
+      feArgsObj['oldSku'] = options.oldSku;
+    }
+
+    if (feArgsObj['oldSku']) {
+      assert(feArgsObj['skus'], 'Need a sku list if old sku is provided!')
+      // remove old sku from offers if in list
```
dparikh: All comments start with Uppercase and ends with a period "."

src/runtime/offers-flow.js
Kestrad: Done, thanks!

src/runtime/offers-flow.js
Kestrad: Done!

src/api/subscriptions.js
Kestrad: Currently since update subscription is still behind an experimental flag, publishers should only have been passing in a string anyway (or else they would have encountered an error telling them the feature is not yet launched). This change just enforces subscribe taking only a string - and therefore only being used for a new subscription - and then updateSubscription will take the a SubscriptionRequest object.

chenshay: +1. This is fine and shouldn't break anything (the "SubscriptionRequest" was added by me but never launched)

src/runtime/offers-flow-test.js
```diff
@@ -156,6 +156,32 @@ describes.realWin('OffersFlow', {}, env => {
     return offersFlow.start();
   });
 
+  it('should have valid OffersFlow constructed with skus and oldSku', () => {
```
chenshay: add another test with expected behavior when a list of skus is not present

src/runtime/offers-flow.js
```diff
@@ -51,20 +52,54 @@ export class OffersFlow {
       isClosable = false; // Default is to hide Close button.
     }
 
+    const feArgsObj = {
```
chenshay: what does "fe" stand for in "feArgsObj"?

src/runtime/offers-flow.js
```diff
@@ -51,20 +52,54 @@ export class OffersFlow {
...
+      'publicationId': deps.pageConfig().getPublicationId(),
+      'showNative': deps.callbacks().hasSubscribeRequestCallback(),
+      'productType': ProductType.SUBSCRIPTION,
+      'list': (options && options.list) || 'default',
+      'skus': (options && options.skus) || null,
+      'isClosable': isClosable,
+    };
+
+    if (options && options.oldSku) {
+      feArgsObj['oldSku'] = options.oldSku;
```
chenshay: I think this can go above like:
'oldSku': (options && options. oldSku) || undefined,

src/runtime/offers-flow.js
chenshay: write a test for this (removing old sku from sku list)

src/runtime/offers-flow.js
```diff
@@ -51,20 +52,54 @@ export class OffersFlow {
...
+    }
+
+    if (feArgsObj['oldSku']) {
```
chenshay: write a test for this use case

src/runtime/runtime.js
jpettitt: yes any place you generate a console error

src/runtime/runtime.js
```diff
@@ -798,15 +833,19 @@ export class ConfiguredRuntime {
...
-    ) {
+  subscribe(sku) {
+    return this.documentParsed_.then(() => {
+      return new PayStartFlow(this, sku).start();
+    });
+  }
+
+  /** @override */
+  updateSubscription(subscriptionRequest) {
+    if (!isExperimentOn(this.win_, ExperimentFlags.REPLACE_SUBSCRIPTION)) {
```
jpettitt: user assert here too

src/runtime/offers-flow.js
```diff
@@ -51,20 +52,54 @@ export class OffersFlow {
...
+      // remove old sku from offers if in list
+      let skuList = feArgsObj['skus'];
+      const /** @type {String} */ oldSku = feArgsObj['oldSku'];
+      skuList = skuList.filter(sku => sku !== oldSku);
+      assert(skuList.length > 0, 'Sku list only contained offer user already has');
+      feArgsObj['skus'] = skuList;
+    }
+
+    // redirect to payments if only one upgrade option is passed
+    if (feArgsObj['skus'] && feArgsObj['skus'].length === 1) {
```
chenshay: write a test for this scenario

src/runtime/offers-flow.js
```diff
@@ -74,34 +109,47 @@ export class OffersFlow {
    * @return {!Promise}
    */
   start() {
-    // Start/cancel events.
-    this.deps_.callbacks().triggerFlowStarted(SubscriptionFlows.SHOW_OFFERS);
-    this.activityIframeView_.onCancel(() => {
-      this.deps_.callbacks().triggerFlowCanceled(SubscriptionFlows.SHOW_OFFERS);
-    });
+    if (this.activityIframeView_) {
+      // so no error is thrown if offers skipped
```
chenshay: Please clarify the comment (I'm not sure what you're trying to say here)

src/runtime/offers-flow.js
```diff
@@ -74,34 +109,47 @@ export class OffersFlow {
...
-        return;
-      }
-    });
+      // If result is due to OfferSelection, redirect to payments.
+      this.activityIframeView_.onMessageDeprecated(result => {
+        if (result['alreadySubscribed']) {
+          this.deps_.callbacks().triggerLoginRequest({
+            linkRequested: !!result['linkRequested'],
+          });
+          return new Promise();
```
chenshay: Why are you returning New Promises now?
Previously there were just return statements.

src/runtime/runtime.js
```diff
@@ -709,6 +721,29 @@ export class ConfiguredRuntime {
   /** @override */
   showOffers(opt_options) {
     return this.documentParsed_.then(() => {
+      if (opt_options['oldSku']) {
+        console.error(
+          'Use the showUpdateOffers method for subscription updates'
```
chenshay: const errorMessage = 
'The showOffers() method cannot be used to update a subscription. Please use the showUpdateOffers() method instead.';

console.error(errorMessage);
return new Error(errorMessage);

assign it to a const

src/runtime/runtime.js
```diff
@@ -709,6 +721,29 @@ export class ConfiguredRuntime {
...
+      }
+      const flow = new OffersFlow(this, opt_options);
+      return flow.start();
+    });
+  }
+
+  /** @override */
+  showUpdateOffers(opt_options) {
+    if (!isExperimentOn(this.win_, ExperimentFlags.REPLACE_SUBSCRIPTION)) {
+      throw new Error('Not yet launched!');
```
chenshay: throw new Error('Feature to replace subscriptions is not yet launched!');

src/runtime/runtime.js
```diff
@@ -294,6 +294,13 @@ export class Runtime {
     );
   }
 
+  /** @override */
+  showUpdateOffers(opt_options) {
```
chenshay: write test

src/runtime/runtime.js
```diff
@@ -337,9 +344,14 @@ export class Runtime {
   }
 
   /** @override */
-  subscribe(skuOrSubscriptionRequest) {
+  subscribe(sku) {
```
chenshay: update test for this

src/runtime/runtime.js
```diff
@@ -709,6 +721,29 @@ export class ConfiguredRuntime {
   /** @override */
   showOffers(opt_options) {
```
chenshay: update test for this

src/runtime/runtime.js
chenshay: @dparikh ?

src/runtime/runtime.js
```diff
@@ -709,6 +721,29 @@ export class ConfiguredRuntime {
...
+          'Use the showUpdateOffers method for new subscription updates'
+        );
+      }
+      const flow = new OffersFlow(this, opt_options);
+      return flow.start();
+    });
+  }
+
+  /** @override */
+  showUpdateOffers(opt_options) {
```
chenshay: add test

src/runtime/runtime.js
```diff
@@ -798,15 +833,19 @@ export class ConfiguredRuntime {
   }
 
   /** @override */
-  subscribe(skuOrSubscriptionRequest) {
-    if (
-      typeof skuOrSubscriptionRequest != 'string' &&
-      !isExperimentOn(this.win_, ExperimentFlags.REPLACE_SUBSCRIPTION)
-    ) {
+  subscribe(sku) {
```
chenshay: update test

src/runtime/runtime.js
```diff
@@ -798,15 +833,19 @@ export class ConfiguredRuntime {
...
+    return this.documentParsed_.then(() => {
+      return new PayStartFlow(this, sku).start();
+    });
+  }
+
+  /** @override */
+  updateSubscription(subscriptionRequest) {
```
chenshay: add test

src/runtime/runtime.js
dparikh: Instead of console.error() use Throw Error("msg")

src/runtime/runtime.js
```diff
@@ -709,6 +721,29 @@ export class ConfiguredRuntime {
...
+  }
+
+  /** @override */
+  showUpdateOffers(opt_options) {
+    if (!isExperimentOn(this.win_, ExperimentFlags.REPLACE_SUBSCRIPTION)) {
+      throw new Error('Not yet launched!');
+    }
+    return this.documentParsed_.then(() => {
+      if (!opt_options['oldSku']) {
+        console.error('Use the showOffers method for new subscribers');
```
dparikh: We don't use console.error() in PROD code. The `new Error("..")` is enough.

src/runtime/offers-flow.js
```diff
@@ -51,20 +51,61 @@ export class OffersFlow {
...
+      'isClosable': isClosable,
+    };
+
+    if (options && options.oldSku) {
+      feArgsObj['oldSku'] = options.oldSku;
+    }
+
+    if (feArgsObj['oldSku']) {
+      if (!feArgsObj['skus']) {
+        console.error('Need a sku list if old sku is provided!');
```
dparikh: `throw New Error("msg");` instead

src/api/subscriptions.js
```diff
@@ -86,6 +86,12 @@ export class Subscriptions {
    */
   showOffers(opt_options) {}
 
+  /**
+   * Starts the Offers flow for a subscription update.
+   * @param {!OffersRequest=} opt_options
+   */
+  showUpdateOffers(opt_options) {}
```
dparikh: Need to add documentation on new APIs.

Similar to: https://github.com/subscriptions-project/swg-js/blob/master/docs/subscribe-flow.md

src/runtime/offers-flow.js
```diff
@@ -51,20 +52,54 @@ export class OffersFlow {
...
+      // so we need to check for oldSku to decide what to send
+      if (oldSku) {
+        new PayStartFlow(this.deps_, {
+          skuId: sku,
+          oldSkuId: oldSku,
+          replaceSkuProrationMode: ReplaceSkuProrationMode.IMMEDIATE_WITH_TIME_PRORATION,
+        }).start();
+        return;
+      } else {
+        new PayStartFlow(this.deps_, sku).start();
```
dparikh: As discussed offline, this `else` clause requires some discussion.

showOffers() can be launched by a publisher without user interaction and for a single SKU, launching a `PlayFlow` seems odd to me (and may be for the users too).

src/runtime/runtime.js
```diff
@@ -709,6 +721,29 @@ export class ConfiguredRuntime {
   /** @override */
   showOffers(opt_options) {
     return this.documentParsed_.then(() => {
+      if (opt_options['oldSku']) {
+        console.error(
+          'Use the showUpdateOffers method for subscription updates'
+        );
+        return new Error(
```
dparikh: should be: `throw new Error("...");`

src/runtime/offers-flow.js
Kestrad: I'm returning promises here because the compiler yelled at me if I didn't return promises here, since the place that invoked this function expected a promise to be returned.

mborof: (happened to see this in email)

Did you mean return Promise.resolve()?  As in this promise is finished and complete now?

src/runtime/runtime.js
jpettitt: It'd rather we use `assert()` over `throw New Error()` to generate the Error because it localizes the error path through the logger so we can log and capture errors if we choose to.

src/runtime/runtime.js
jpettitt: use `assert()`

src/runtime/offers-flow.js
Kestrad: Oh, I'd tried `return new Promise.resolve()` and it didn't work, but in retrospect I probably should have done it without new. I think what you're suggesting would be better than what I have now.

src/runtime/offers-flow.js
Kestrad: Fixed, thanks.

src/runtime/runtime.js
Kestrad: I went ahead and turned all instances of `throw new Error` in runtime into `assert()` instead.

src/runtime/runtime.js
Kestrad: Done!

src/runtime/runtime.js
Kestrad: Done!

src/runtime/runtime.js
Kestrad: Done!

src/runtime/offers-flow.js
Kestrad: Done!

src/runtime/runtime.js
Kestrad: Done!

src/runtime/offers-flow.js
Kestrad: Unfortunately, doing this breaks our tests because they expect the object to not have an `oldSku` property at all if `oldSku` isn't relevant. I guess we could modify all the tests?

src/runtime/offers-flow.js
Kestrad: Done, hopefully clearer now.

src/runtime/offers-flow-test.js
Kestrad: Added, will be in next PR!

src/runtime/offers-flow.js
Kestrad: Added, will be in next PR!

src/runtime/offers-flow.js
Kestrad: I assume that addressing the comment on offers-flow-test.js also addressed this one.

src/runtime/offers-flow.js
Kestrad: Added - will be in next PR!

src/runtime/offers-flow.js
Kestrad: I've reset them to just say `return;` for now, since that really makes the most sense. Will re-evaluate depending on whether it passes presubmit or not.

src/runtime/offers-flow.js
Kestrad: I've commented out the auto-redirect for when there's no oldSku. (Actually I probably should have deleted the code outright - I'll fix that in the next push.)

src/runtime/offers-flow-test.js
Kestrad: And pushed!

src/runtime/offers-flow.js
Kestrad: Pushed!

src/runtime/offers-flow.js
Kestrad: Pushed!

src/runtime/offers-flow.js
```diff
@@ -51,20 +52,57 @@ export class OffersFlow {
...
+    if (feArgsObj['skus'] && feArgsObj['skus'].length === 1) {
+      const sku = feArgsObj['skus'][0];
+      const /** @type {String} */ oldSku = feArgsObj['oldSku'];
+      // Update subscription triggers experimental flag if oldSku is passed,
+      // so we need to check for oldSku to decide if it needs to be sent.
+      // Otherwise we might accidentally block a regular subscription request.
+      if (oldSku) {
+        new PayStartFlow(this.deps_, {
+          skuId: sku,
+          oldSkuId: oldSku,
```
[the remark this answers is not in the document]
Kestrad: Just to clarify, you're suggesting replacing oldSkuId with oldSku, right? oldSkuId was the name of the field as defined in SubscriptionRequest, but it looks like it's not used in too many places yet and should be relatively easy to clean up.

src/api/subscriptions.js
Kestrad: We definitely do! I'll start writing a documentation page. But since the new endpoints are still experimental right now, I don't think we need to have the documentation ready before submitting.

src/runtime/offers-flow.js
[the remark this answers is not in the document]
alin04: Correct.

src/runtime/offers-flow.js
chenshay: I see. In that case ignore this comment :)

src/runtime/runtime.js
chenshay: ok ignore :)

src/runtime/runtime.js
chenshay: probably :)

src/runtime/offers-flow.js
```diff
@@ -51,20 +52,57 @@ export class OffersFlow {
...
+    }
+
+    // Redirect to payments if only one upgrade option is passed.
+    if (feArgsObj['skus'] && feArgsObj['skus'].length === 1) {
+      const sku = feArgsObj['skus'][0];
+      const /** @type {String} */ oldSku = feArgsObj['oldSku'];
+      // Update subscription triggers experimental flag if oldSku is passed,
+      // so we need to check for oldSku to decide if it needs to be sent.
+      // Otherwise we might accidentally block a regular subscription request.
+      if (oldSku) {
```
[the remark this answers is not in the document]
Kestrad: Right, removed. I did leave a placeholder that should always resolve to undefined, though, because pay-flow is expecting a SubscriptionRequest object (which should have a replaceSkuProrationMode field), and just in case we allow publishers to pass in a proration mode in the future.

src/runtime/offers-flow.js
```diff
@@ -51,20 +52,57 @@ export class OffersFlow {
...
+        new PayStartFlow(this.deps_, {
+          skuId: sku,
+          oldSkuId: oldSku,
+          replaceSkuProrationMode: ReplaceSkuProrationMode.IMMEDIATE_WITH_TIME_PRORATION,
```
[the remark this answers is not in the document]
Kestrad: This is already implemented, but I've gotten rid of the duplicate logic, as mentioned above.

src/runtime/offers-flow.js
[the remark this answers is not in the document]
Kestrad: Done! Pretty sure I've found all instances of oldSkuId and replaced them with oldSku.

src/runtime/runtime.js
Kestrad: Added test!

src/runtime/runtime.js
Kestrad: Added test!

src/runtime/runtime.js
Kestrad: Added test!

jpettitt:
Ooops jumped the gun there, a few test fixes needed

src/api/subscriptions.js
```diff
@@ -428,7 +440,7 @@ export let SmartButtonOptions;
  *
  *  @typedef {{
  *    skuId: string,
- *    oldSkuId: (string|undefined),
+ *    oldSku: (string|undefined),
```
sohanirao: are you sure nobody is using the key oldSkuId already? Looks like an API change to me.

Kestrad: Yeah, doing a quick codesearch, it doesn't show up anywhere that wouldn't have been hidden behind an experimental flag.

PR #602, opened 2019-07-24
Allows async/await in tests

ChrisAntaki:


ChrisAntaki:
Yep! I just find async/await syntax more concise and readable 😃

PR #610, opened 2019-07-26
Cleans up `changelog` gulp task

ChrisAntaki:
- Removes a TODO
- Reduces nesting with async/await pattern
- Formats with `prettier`


build-system/tasks/changelog.js
```diff
@@ -44,30 +44,26 @@ const logger = require('fancy-log');
...
 function changelog(opt_options) {
   return getLastGithubRelease()
-      .then(getGitLog)
-      .then(getGithubPullRequestsMetadata)
-      .then(buildChangelog)
-      .then(function(response) {
-        logger(colors.blue('\n' + response.changelog));
-      })
-      .catch(errHandler);
+    .then(getGitLog)
```
dparikh: For continuation, it should be 4 chars indentation.

Here and everywhere.

dparikh: LGTM. We could configure prettier to have 4 chars indentation though.

PR #613, opened 2019-07-26
Create Logger

mborof:


mborof:
Currently event manager is not available until after configured runtime is created.  This was so we didn't need to allow it being passed in from runtime.  We wanted logging on runtime without having it wrapped in a promise.  So to solve that, I give logger a promise to event manager so it can exist independently without a promise.   That way runtime.getLogger().logEvent() works instead of needing to:

runtime.getLogger().then(logger => logger.logEvent());

I have no idea why deps exists so wouldn't add to it without a reason.  I figure it needs to be part of SubscriptionsAPI so publishers can call it directly.  Would it also need to be on deps for some reason?

mborof:
It looks like deps isn't available until ConfiguredRuntime. We want logger to be available on runtime and configuredruntime so deps doesn't really work. Event manager isn't available until after runtime is created and then later configured. So it really can't accept a deps object which isn't actually formed yet

mborof:
If we opt to hide logger inside a promise ( subscriptions.getLogger().then(logger => ... ) then we can go with the deps approach.

mborof:
We decided we would place logger behind a promise on the subscriptions API.  This means that Logger can take a deps, so this change is now made and this branch should be ready for submission.

src/runtime/logger.js
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  /**
+   *
+   * @param {!AnalyticsEvent} eventType
+   * @param {?boolean=} isFromUserAction
+   * @param {?Object=} additionalParameters
+   * @param {EventOriginator=} originator
+   */
+  logSwgEvent(eventType, isFromUserAction, additionalParameters, originator) {
```
sohanirao: Do you think it would be enough to call it logEvent and to document that API for SWG internal events than using logSwgEvent? Or perhaps have logEventInternal()?

Just a suggestion. logSwgEvent by itself as a name provides very little indication of how its different than sendEvent IMHO.

mborof: This was an oversight, I intended to delete this function anyway.  We will keep logger purely for external use and use event manager internally.

PR #614, opened 2019-07-26
Make Logger Available in Runtime

mborof:


mborof:
Just a note here Chris.  The logger.js and logger-test.js are both part of a different pull request so you can skip reviewing those two files.  This PR is focused on integrating logger into our runtime.

src/runtime/runtime-test.js
```diff
@@ -470,17 +470,17 @@ describes.realWin('Runtime', {}, env => {
       );
     });
 
-    it('should eventually log events', () => {
-      runtime.logEvent({
+    it('should resolve event manager promise', async function() {
```
ChrisAntaki: Going `async`, nice!

mborof:
The is exactly the intention and is exactly why logger takes a promise to get event manager instead of directly taking event manager.

sohanirao:
My question was more about how the API looks externally, not about the argument to the constructor. Do you intend for API calls to Propensity and Logger to look different, because they will.

mborof:
It looks like deps isn't available until ConfiguredRuntime.  We want logger to be available on runtime and configuredruntime so deps doesn't really work.  Event manager isn't available until after runtime is created and then later configured.  So it really can't accept a deps object which isn't actually formed yet

mborof:
I think its a valid point and worth considering.  With only 1 publisher currently using Propensity I'd personally lean towards the improved syntax of obscuring the promise from their view but it may make sense to hide logger behind a promise.  I think we should involve John and Elijah in this decision so its well thought out from multiple perspectives.

mborof:
We decided to keep the external API identical to propensity's so logger is now behind a promise.

We also decided to remove the helper log function - internal developers should use event manager directly.

src/runtime/runtime.js
```diff
@@ -881,6 +885,16 @@ export class ConfiguredRuntime {
   eventManager() {
     return this.eventManager_;
   }
+
+  /**
+   * Note that this returns Logger instead of LoggerApi for internal logging
```
sohanirao: nit: this comment is confusing to callers. What you are trying to say that it returns the implementation of LoggerApi?

/**
 * Returns the Logger module that implements the LoggerApi
 */

src/runtime/runtime.js
```diff
@@ -442,13 +443,13 @@ export class Runtime {
   }
 
   /**
-   * Log a subscription buy-flow event.
-   * @param {!../api/client-event-manager-api.ClientEvent} event
+   * Note that this returns Logger instead of LoggerApi for internal logging
+   * purposes.
+   * @override
+   * @return {!Promise<Logger>}
```
sohanirao: /**

Returns the Logger module that implements the LoggerApi
*/

src/runtime/runtime.js
mborof: It was telling you that it was returning Logger, which had a slightly different API than LoggerAPI because we had the internal use function on there to logSwgEvents.  We took that out though making the comment out of date and the override unnecessary.  Removed.

PR #617, opened 2019-07-30
Transition contributions and offers flow to on() and execute() APIs

sohanirao:


src/proto/api_messages_test.js
```diff
@@ -56,6 +60,20 @@ function isEqual(thisArray, otherArray) {
 
 describe('api_messages', () => {
 
+  describe('test_AlreadySubscribedResponse', () => {
+    it('should deserialize correctly', () => {
+      const /** !AlreadySubscribedResponse  */ alreadysubscribedresponse = new AlreadySubscribedResponse();
```
ChrisAntaki: Would it be easy to make these local test variables camelCase?

src/runtime/contributions-flow-test.js
```diff
@@ -29,7 +33,7 @@ describes.realWin('ContributionsFlow', {}, env => {
   let callbacksMock;
   let pageConfig;
   let port;
-  let messageCallback;
+  const messageMap = {};
```
ChrisAntaki: This is a cool pattern, but I worry it leaks state across tests. Could the messageMap be declared with `let` instead, and then cleared `beforeEach` test?

src/proto/api_messages_test.js
ChrisAntaki: This comment is more of a nit, for readability

sohanirao: Sure please feel free to bug on Protobuf Library generator code and assign to me. Would be happy to address it during fix it.

src/runtime/contributions-flow-test.js
sohanirao: good idea. done

src/proto/api_messages_test.js
ChrisAntaki: Sounds good

sohanirao:
@googlebot I fixed it.

sohanirao:
This PR was actually submitted using P#646

PR #621, opened 2019-07-31
Redirect propensity to logger

mborof:
This is the last PR for the new publisher logging tool and just has propensity API use logger for the logging functions.  I think we could consider:

1) removing some of the tests in propensity-test because they are ultimately duplicated by logger
2) passing in a deps to propensity instead of 4 individual parameters but I think Sohani objected to this once and I don't recall why (or might be misremembering).

I wanted Sohani's feedback on those 2 things before proceeding with those changes.

### DIFF
--- a/src/runtime/propensity.js
+++ b/src/runtime/propensity.js
@@ -14,12 +14,7 @@
  * limitations under the License.
  */
 import * as PropensityApi from '../api/propensity-api';
-import {Event, SubscriptionState} from '../api/logger-api';
 import {PropensityServer} from './propensity-server';
-import {isObject, isEnumValue} from '../utils/types';
-import {EventOriginator} from '../proto/api_messages';
-import {publisherEventToAnalyticsEvent} from './event-type-mapping';
-import {isBoolean} from '../utils/types';
 
 /**
  * @implements {PropensityApi.PropensityApi}
@@ -29,8 +24,9 @@ export class Propensity {
    * @param {!Window} win
    * @param {!../model/page-config.PageConfig} pageConfig
    * @param {!../api/client-event-manager-api.ClientEventManagerApi} eventManager
+   * @param {!../api/logger-api.LoggerApi} logger
    */
-  constructor(win, pageConfig, eventManager) {
+  constructor(win, pageConfig, eventManager, logger) {
     /** @private @const {!Window} */
     this.win_ = win;
     /** @private {PropensityServer} */
@@ -40,33 +36,13 @@ export class Propensity {
       eventManager
     );
 
-    /** @private @const {!../api/client-event-manager-api.ClientEventManagerApi} */
-    this.eventManager_ = eventManager;
+    /** @private @const {!../api/logger-api.LoggerApi} */
+    this.logger_ = logger;
   }
 
   /** @override */
   sendSubscriptionState(state, jsonProducts) {
-    if (!Object.values(SubscriptionState).includes(state)) {
-      throw new Error('Invalid subscription state provided');
-    }
-    if (
-      (SubscriptionState.SUBSCRIBER == state ||
-        SubscriptionState.PAST_SUBSCRIBER == state) &&
-      !jsonProducts
-    ) {
-      throw new Error(
-        'Entitlements must be provided for users with' +
-          ' active or expired subscriptions'
-      );
-    }
-    if (jsonProducts && !isObject(jsonProducts)) {
-      throw new Error('Entitlements must be an Object');
-    }
-    let productsOrSkus = null;
-    if (jsonProducts) {
-      productsOrSkus = JSON.stringify(jsonProducts);
-    }
-    this.propensityServer_.sendSubscriptionState(state, productsOrSkus);
+    this.logger_.sendSubscriptionState(state, jsonProducts);
   }
 
   /** @override */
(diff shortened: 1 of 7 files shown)


src/runtime/propensity.js
```diff
@@ -40,33 +36,13 @@ export class Propensity {
...
-    }
-    if (jsonProducts && !isObject(jsonProducts)) {
-      throw new Error('Entitlements must be an Object');
-    }
-    let productsOrSkus = null;
-    if (jsonProducts) {
-      productsOrSkus = JSON.stringify(jsonProducts);
-    }
-    this.propensityServer_.sendSubscriptionState(state, productsOrSkus);
+    this.logger_.sendSubscriptionState(state, jsonProducts);
```
sohanirao: so we are also not talking to PropensityServer directly?

src/runtime/runtime.js
```diff
@@ -532,6 +525,14 @@ export class ConfiguredRuntime {
     /** @private @const {!Logger} */
     this.logger_ = new Logger(this);
 
+    /** @private @const {!Propensity} */
+    this.propensityModule_ = new Propensity(
+      this.win_,
+      this.pageConfig_,
+      this.eventManager_,
```
sohanirao: why does Propensity need both event manager and logger?

src/runtime/propensity.js
mborof: Well eventually if something else cares to know about a subscription state change then the current method would not be able to inform it. Logger uses an event so whomever is interested can simply subscribe.

sohanirao: Right now subscritpion state change is something PropensityServer needs to know. So I think the handling of when someone calls loggerAPI and PropensityServer handles it is necessary. However, this event is not useful in our Analytics. So the only advantage of doing this would be for AnalyticsService to get this event, which I don't know is going to add a lot. You need to acknowledge the fact that by doing this change you are changing the API behavior. sendSubscriptionState() programmatically returns/throws an error to the caller. The reason I wanted it to stay in place was because it would be an important milestone in the integration process for publishers to ensure they can send messages to DRX server without any errors.

mborof: We've had this conversation multiple times already.

sendSubscriptionState was added to the logger API in the first place because you had indicated analytics would eventually need or want to know this information.  If that is not actually true then the appropriate thing to do might be to remove it from the API entirely rather than maintain it in two different locations with identical code.  

It might be arguable that logger is a more logical place to send subscription state to anyway (if you are going to have a logger in the first place that is) and so it should stay in logger.

The alternative was to pass around propensity server, which was the initial proposal and would facilitate direct communications and not modify the error throwing behavior (I would still redirect propensity to logger and would NEVER want to literally have identical code in two places).  You were abjectly opposed to that proposal when it was made which is why we came up with the indirect event method in the first place. You were a part of those discussions.

As to the error handling issue, I'm not sure that throwing an error is the most appropriate thing that should happen - developers should be ensuring console doesn't have errors before they ship their product out.  I'm not entirely confident about that though John has indicated he thinks it should work this way.  If the error throwing really is a important piece of functionality then we absolutely have to pass around propensity server.  There is no alternative.

There was a long email chain between you and John about the error.  What was the result of that?

sohanirao: Short term:
Propensity API gets used by publishers
analytics BE doesn't care about subscription state
Logger api is complete and if that event comes through logger, it's handled.
Only Propensity Servers care about this information to provide to their model.

Long term:
Logger API replaces Propensity API.
Propensity API will not be used by publishers anyway, so LoggerApi will be used and will send Subscription state to both propensity and Analytics BE.
Analytics BE might start using this information in the future, right now it doesn't. 

In light of the changed API behavior, sendSubscriptionState() is the first message and programmatically throws an error vs logging an error after you make your change. Just want to understand and document why you think this change is important right now to justify the API behavior change.

jpettitt: We want the subscription state in the SwG logs for buyflow comparison so it should go through logger at some point soon.  I'm happy to leave it direct to propensity for this PR.

The error behavior will have to change anyway when we switch out the xhr for the version that amp can polyfill when we do the amp implementation.

jpettitt: otherwise LGTM

sohanirao: Might be a good idea to make this change when error handling is changed across the board anyway? jpettitt@ to make final call. I'll LGTM.

mborof: I'll just add that I think shorter-term new propensity publishers should be using logger for this.  I'm not saying we need to force anyone who already is to change, just that new publishers coming online should be using the longer term method so they can avoid switching (not that switching should be painful at all).

mborof: I'll also add that it isn't really an API change because it is potentially an error that a brand new implementing customer could encounter and not one an existing customer would be likely to encounter.  

If an existing customer did happen to encounter the error (of not being able to communicate with the propensity servers) then the current behavior of throwing the error could potentially be very bad because it might actively prevent a user from subscribing.  (depending on how each individual publisher had implemented things in their code).  A nice to have piece of functionality shouldn't block the most important workflow of the application (subscribing).  It should fail gracefully and log to console so someone can eventually figure out the error is occurring but the user doesn't potentially get hard-stopped in their tracks and prevented from continuing.

mborof:
Sounds to me like I'm not supposed to merge until Sohani resolves the conversation so despite both reviewers approving I'm going to hold off until that happens.

mborof:
Closing this so Sohani can make the change herself after October

PR #622, opened 2019-08-05
Log SwG button clicks with EventManager

elijahsoria:
Replaces ACTION_SUBSCRIBE logging with ACTION_PAYMENT_FLOW_STARTED in pay-flow.js. Adds EventManager as a private member to Runtime and uses that in ButtonApi to log SwG button click events.

### DIFF
--- a/src/runtime/button-api.js
+++ b/src/runtime/button-api.js
@@ -58,14 +58,14 @@ const TITLE_LANG_MAP = {
 export class ButtonApi {
   /**
    * @param {!../model/doc.Doc} doc
-   * @param {!./client-event-manager.ClientEventManager} eventManager
+   * @param {!Promise<!./runtime.ConfiguredRuntime>} configuredRuntimePromise
    */
-  constructor(doc, eventManager) {
+  constructor(doc, configuredRuntimePromise) {
     /** @private @const {!../model/doc.Doc} */
     this.doc_ = doc;
 
-    /** @private @const {!./client-event-manager.ClientEventManager} */
-    this.eventManager_ = eventManager;
+    /** @private @const {!Promise<!./runtime.ConfiguredRuntime>} */
+    this.configuredRuntimePromise_ = configuredRuntimePromise;
   }
 
   /**
@@ -122,12 +122,16 @@ export class ButtonApi {
     }
     button.setAttribute('title', msg(TITLE_LANG_MAP, button) || '');
     button.addEventListener('click', callback);
-    button.addEventListener('click', () =>
-      this.eventManager_.logSwgEvent(
-        AnalyticsEvent.ACTION_SWG_BUTTON_CLICK,
-        true
-      )
-    );
+    button.addEventListener('click', () => {
+      this.configuredRuntimePromise_.then(configuredRuntime => {
+        configuredRuntime
+          .eventManager()
+          .logSwgEvent(
+            AnalyticsEvent.ACTION_SWG_BUTTON_CLICK,
+            /* isFromUserAction */ true
+          );
+      });
+    });
     return button;
   }
 
@@ -186,9 +190,13 @@ export class ButtonApi {
     // Add required CSS class, if missing.
     button.classList.add('swg-smart-button');
     button.addEventListener('click', () =>
-      this.eventManager_.logSwgEvent(
-        AnalyticsEvent.ACTION_SWG_BUTTON_CLICK,
-        true
+      this.configuredRuntimePromise_.then(configuredRuntime =>
+        configuredRuntime
+          .eventManager()
+          .logSwgEvent(
+            AnalyticsEvent.ACTION_SWG_BUTTON_CLICK,
+            /* isFromUserAction */ true
+          )
       )
     );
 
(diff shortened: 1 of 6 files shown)


src/runtime/pay-flow-test.js
```diff
@@ -235,7 +235,7 @@ describes.realWin('PayStartFlow', {}, env => {
     analyticsMock.expects('setSku').withExactArgs('newSku');
     eventManagerMock
       .expects('logSwgEvent')
-      .withExactArgs(AnalyticsEvent.ACTION_SUBSCRIBE, true, null);
+      .withExactArgs(AnalyticsEvent.ACTION_PAYMENT_FLOW_STARTED, true, null);
```
mborof: could we drop the final , null for all these calls?  If we are just going to use the default anyway I feel like we might as well.

src/runtime/runtime.js
```diff
@@ -161,8 +161,12 @@ export class Runtime {
     /** @private {?PageConfigResolver} */
     this.pageConfigResolver_ = null;
 
+    /** @private @const {!ClientEventManager} */
+
+    this.eventManager_ = new ClientEventManager(this.configuredPromise_);
```
mborof: Unfortunately, we cannot create an event manager in runtime unless we also create a way to pass that event manager into configured runtime.  This is because there should only ever be 1 event manager (otherwise things might not all subscribe to the same one).  

We can create a method to pass event manager into configuredruntime, and we did have a way of doing this briefly but we went with the configuration promise instead because AMP couldn't instantiate event manager anyway.

The easiest thing you can do for your API is to pass in this.configuredPromise_.  Then, within your code:

this.configuredPromise_.then( configuredRuntime => 
  configuredRuntime.eventManager().then( em => 
    em.logSwgEvent( ... )
  )
);

src/runtime/button-api.js
```diff
@@ -117,6 +122,12 @@ export class ButtonApi {
     }
     button.setAttribute('title', msg(TITLE_LANG_MAP, button) || '');
     button.addEventListener('click', callback);
+    button.addEventListener('click', () =>
+      this.eventManager_.logSwgEvent(
+        AnalyticsEvent.ACTION_SWG_BUTTON_CLICK,
+        true
```
sohanirao: maybe a little comment on what this true is for //From user Action

src/runtime/button-api.js
```diff
@@ -174,6 +185,12 @@ export class ButtonApi {
 
     // Add required CSS class, if missing.
     button.classList.add('swg-smart-button');
+    button.addEventListener('click', () =>
+      this.eventManager_.logSwgEvent(
+        AnalyticsEvent.ACTION_SWG_BUTTON_CLICK,
```
sohanirao: is this a smart button event? or same as Line 127

sohanirao: do you want to distinguish a button click to a smart button than a regular swg button?

elijahsoria: Both click events will log the same action: ACTION_SWG_BUTTON_CLICK.

The question of distinguishing between SmartButton and regular came up in discussion and we settled on not logging a separate event just for SmartButton clicks since we will be able to distinguish them by looking at the impression that happened before the click anyhow.

src/runtime/runtime.js
elijahsoria: Of course, I remember this conversation now. Sorry for the mental lapse (long weekend!).

src/runtime/button-api.js
elijahsoria: Done.

src/runtime/button-api.js
sohanirao: You don't need a second event but you could pass in an event parameter though. Whatever helps you filter it in the past without having the depend on a previous event.

elijahsoria: I agree it would be easier to filter out by adding an additional parameter. I believe I brought this up in a conversation with John, Carter and Shams and we landed on not adding any SmartBox-specific parameters though. I can confirm this with John tomorrow when he gets back.

sohanirao: All you need to know is if it was a smart box click or not? Actually you can get that from boq-web iframe too if you log it there. Either ways, only useful to have this distinction if it helps you in your analytics.

src/runtime/runtime.js
```diff
@@ -547,7 +547,7 @@ export class ConfiguredRuntime {
     this.offersApi_ = new OffersApi(this.pageConfig_, this.fetcher_);
 
     /** @private @const {!ButtonApi} */
-    this.buttonApi_ = new ButtonApi(this.doc_);
+    this.buttonApi_ = new ButtonApi(this.doc_, opt_integr.configPromise);
```
mborof: Use Promise.resolve(this) instead.  The one you're passing isn't going to return a runtime object.  Should you add a test for this in runtime-test?

elijahsoria: Great catch, Matt! Switched to Promise.resolve(this) and added a test.

src/runtime/button-api.js
elijahsoria: Talked to John this morning and we decided that we shouldn't be adding any additional information into logs and just look at the impression that came before the SmartButton click.

PR #624, opened 2019-08-06
SwG Button Impression Logging

elijahsoria:
Adds impression logging to ButtonApi on attach of SwG button. Adds tests to verify that impression logging is being done.

Note that we decided to add impression logging on our end as well as have publishers log button impression with LoggerApi because we will then be able to grab all impressions done on the SwG button and differentiate between publisher logged impressions and SwG logged impressions by looking at the event origin.

### DIFF
--- a/src/runtime/button-api-test.js
+++ b/src/runtime/button-api-test.js
@@ -15,6 +15,7 @@
  */
 
 import {ButtonApi} from './button-api';
+import {ClientEventManager} from './client-event-manager';
 import {ConfiguredRuntime} from './runtime';
 import {PageConfig} from '../model/page-config';
 import {Theme} from './smart-button-api';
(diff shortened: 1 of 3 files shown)


src/runtime/runtime-test.js
```diff
@@ -1687,7 +1687,7 @@ describes.realWin('ConfiguredRuntime', {}, env => {
 
       const button = runtime.createButton();
       await button.click();
-      expect(count).to.equal(1);
+      expect(count).to.not.equal(0);
```
sohanirao: I dont understand the reason for this change. Is this carried over from a previous change or something changed in how button clicks are counted?

ChrisAntaki: I mentioned this too, it's been fixed 👍

src/runtime/button-api-test.js
```diff
@@ -187,21 +188,42 @@ describes.realWin('ButtonApi', {}, env => {
...
+    let count = 0;
+    sandbox
+      .stub(ClientEventManager.prototype, 'logSwgEvent')
+      .withArgs(AnalyticsEvent.ACTION_SWG_BUTTON_CLICK, true)
+      .callsFake(() => count++);
+    await button.click();
+    expect(count).to.equal(1);
+  });
+
+  it('should log button impression on create', async () => {
```
sohanirao: And this is because create calls attach. yes?

elijahsoria: Yes that is correct. I just wanted to test both functions explicitly.

PR #626, opened 2019-08-07
switch propensity to using passed in fetcher for AMP compatibility

jpettitt:
Uses passed in `Fetcher` which allows AMP to override the built in `XhrFetcher` to maintain privacy guarantees in pre-render.  This is a requirement for the upcoming implementation of Propensity in AMP.


src/runtime/propensity.js
```diff
@@ -29,15 +29,17 @@ export class Propensity {
    * @param {!Window} win
    * @param {!../model/page-config.PageConfig} pageConfig
    * @param {!../api/client-event-manager-api.ClientEventManagerApi} eventManager
+   * @param {!./fetcher.Fetcher} fetcher
    */
-  constructor(win, pageConfig, eventManager) {
+  constructor(win, pageConfig, eventManager, fetcher) {
```
sohanirao: I would have expected changes to Propensity-test.js where the Propensity object is created.

jpettitt: Added it, it never actually calls the fetcher in propensity-test because it's using a fake propensity-server. Fetcher is tested in propensity-server-test

src/runtime/propensity-server.js
```diff
@@ -29,18 +28,20 @@ export class PropensityServer {
...
+  constructor(win, publicationId, eventManager, fetcher) {
     /** @private @const {!Window} */
     this.win_ = win;
     /** @private @const {string} */
     this.publicationId_ = publicationId;
     /** @private {?string} */
     this.clientId_ = null;
-    /** @private @const {!Xhr} */
-    this.xhr_ = new Xhr(win);
+    /** @private @const {!./fetcher.Fetcher} */
```
sohanirao: I didn't realizer fetcher had a Fetcher. I thought fetcher was a Fetcher.

jpettitt:
Good catch, It didn't complain because it's using a stubbed `PropensityServer` in `propensity-test`

src/runtime/propensity-server.js
jpettitt: Fetcher is a class XhrFetcher is an implementation of it.  I added `fetch()` to the class to match the one in AMP

PR #630, opened 2019-08-16
Remove IMPRESSION_PAYWALL logging from getEntitlements.

elijahsoria:
Removes the IMPRESSION_PAYWALL logging in getEntitlements call as it is incorrectly placed and non-informative.

### DIFF
--- a/src/runtime/entitlements-manager-test.js
+++ b/src/runtime/entitlements-manager-test.js
@@ -38,7 +38,6 @@ describes.realWin('EntitlementsManager', {}, env => {
   let storageMock;
   let config;
   let analyticsMock;
-  let eventManagerMock;
   let deps;
   let encryptedDocumentKey;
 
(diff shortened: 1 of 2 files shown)


src/runtime/entitlements-manager.js
```diff
@@ -144,12 +132,6 @@ export class EntitlementsManager {
       if (response.isReadyToPay != null) {
         this.analyticsService_.setReadyToPay(response.isReadyToPay);
       }
-      if (
-        this.config_.analyticsMode == AnalyticsMode.IMPRESSIONS ||
-        this.isGoogleUtmSource_()
```
mborof: I believe this was the only place this function was being used.  Should we remove it?

src/runtime/entitlements-manager-test.js
```diff
@@ -79,6 +78,7 @@ describes.realWin('EntitlementsManager', {}, env => {
     xhrMock.verify();
     jwtHelperMock.verify();
     analyticsMock.verify();
+    eventManagerMock.verify();
```
mborof: Nothing seems to have any expectations of event manager mock any more.  Should it just go away?

elijahsoria: Yes I removed it. I still need to instantiate a ClientEventManager here because AnalyticsService uses it though.

src/runtime/entitlements-manager.js
elijahsoria: Yes, good catch.

PR #631, opened 2019-08-16
Revert "switch propensity to using passed in fetcher for AMP compatibility"

jpettitt:
Reverts subscriptions-project/swg-js#626 since it break the AMP build - I missed some needed downstream changes in AMP subscriptions google and in the Fetcher class def.


jpettitt:
May not be needed after all.

jpettitt:
not needed -the change required needed to happen simultaneously with the swg release so I added it to that PR in the amp repo.

PR #633, opened 2019-08-16
Event Manager / Analytics Service / Respect Publisher Events

mborof:
Updates analytics service to be aware of PUBLISHER_CLIENT when attempting to ignore publisher logged events.

Also adds some unit tests for the new logSwgEvent helper function.

### DIFF
--- a/src/runtime/analytics-service.js
+++ b/src/runtime/analytics-service.js
@@ -28,6 +28,7 @@ import {setImportantStyles} from '../utils/style';
 import {uuidFast} from '../../third_party/random_uuid/uuid-swg';
 import {ExperimentFlags} from './experiment-flags';
 import {isBoolean} from '../utils/types';
+import {ClientEventManager} from './client-event-manager';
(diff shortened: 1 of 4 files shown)


src/runtime/analytics-service.js
```diff
@@ -265,14 +266,21 @@ export class AnalyticsService {
     });
   }
 
+  /**
+   * @return {boolean}
+   */
+  shouldLogPublisherEvents_() {
```
elijahsoria: I believe this makes it so publishers can't log to propensity without specifying logFromPublisherConfig_. Is that to be expected?

mborof: This makes it so the analytics service only receives events from publisher clients if a configuration setting and an experiment are both active.

This does not effect what propensity gets, that's controlled by propensity-server

elijahsoria: I see now! LGTM

PR #634, opened 2019-08-16
remove logEvent from analytics service

mborof:
Does some final cleanup to remove the old logging function from analytics service.  Everything should now use event manager.

### DIFF
--- a/src/runtime/runtime-test.js
+++ b/src/runtime/runtime-test.js
@@ -953,6 +953,7 @@ describes.realWin('ConfiguredRuntime', {}, env => {
     let activityResultCallbacks;
     let offersApiMock;
     let redirectErrorHandler;
+    let eventManagerMock;
 
     beforeEach(() => {
       activityResultCallbacks = {};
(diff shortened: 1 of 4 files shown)


src/runtime/runtime.js
```diff
@@ -562,7 +562,10 @@ export class ConfiguredRuntime {
     // Report redirect errors if any.
     this.activityPorts_.onRedirectError(error => {
       this.analyticsService_.addLabels(['redirect']);
-      this.analyticsService_.logEvent(AnalyticsEvent.EVENT_PAYMENT_FAILED);
+      this.eventManager().logSwgEvent(
```
elijahsoria: I would use this.eventManager_ instead of this.eventManager() since we are in the constructor.

src/runtime/runtime-test.js
```diff
@@ -1178,9 +1181,14 @@ describes.realWin('ConfiguredRuntime', {}, env => {
         .expects('addLabels')
         .withExactArgs(['redirect'])
         .once();
-      analyticsMock
+      eventManagerMock
         .expects('logEvent')
```
elijahsoria: Shouldn't this be .expects('logSwgEvent') since that is the function we are calling in runtime? Although logEvent gets called internally, that implementation might change in the future.

mborof: done

PR #635, opened 2019-08-16
create config setting for buyflow comparison

mborof:
Creates a new configuration setting to enable buy flow comparison


src/api/subscriptions.js
```diff
@@ -278,11 +278,16 @@ export const SubscriptionFlows = {
  *   force redirect flow for any window.open operation, including payments.
  *   The "auto" value either uses a redirect or a popup flow depending on
  *   what's possible on a specific environment. Defaults to "auto".
- *
+ * - enableSwgAnalytics - if set to true then events logged by the publisher's
+ *   client will be sent to Google's SwG analytics service.  This information is
+ *   used to compare the effectiveness of Google's buy-flow events to those
+ *   generated by the publisher's client code.  This includes events sent to
+ *   both propensity and logger.
```
elijahsoria: nit: "both PropensityApi and LoggerApi".

src/api/subscriptions.js
```diff
@@ -278,11 +278,16 @@ export const SubscriptionFlows = {
...
  * @typedef {{
  *   experiments: (!Array<string>|undefined),
  *   windowOpenMode: (!WindowOpenMode|undefined),
  *   analyticsMode: (!AnalyticsMode|undefined),
+ *   enableSwgAnalytics: (boolean|undefined),
```
elijahsoria: Should this be (!boolean|undefined)? I don't know if this matters or not.

mborof: boolean is a non-nullable value so no.

Boolean is nullable so it would need the !

src/api/subscriptions.js
mborof: ok

PR #638, opened 2019-08-19
New impression and action logging

mborof:
Adds new log events

### DIFF
--- a/src/runtime/pay-flow.js
+++ b/src/runtime/pay-flow.js
@@ -271,6 +271,10 @@ export class PayCompleteFlow {
    */
   complete() {
     this.eventManager_.logSwgEvent(AnalyticsEvent.ACTION_ACCOUNT_CREATED, true);
+    this.eventManager_.logSwgEvent(
+      AnalyticsEvent.IMPRESSION_SUBSCRIPTION_COMPLETE,
+      true
+    );
     this.deps_.entitlementsManager().unblockNextNotification();
     this.readyPromise_.then(() => {
       this.activityIframeView_.messageDeprecated({'complete': true});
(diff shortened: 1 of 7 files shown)


mborof:
New events logged:
  ACTION_ALREADY_SUBSCRIBED - if you click "already subscribed" from the list of offers or from the second abbreviated "view offers" screen
  ACTION_OFFER_SELECTED - if you select an offer from the offers screen
  IMPRESSION_CLICK_TO_SHOW_OFFERS - user viewed an abbreviated "show offers" screen which just contains a button that says "show offers".
  IMPRESSION_CLICK_TO_SHOW_OFFERS_OR_ALREADY_SUBSCRIBED - user viewed the second abbreviated "show offers" screen.  It includes an "already subscribed" link and a "show offers" button.
  ACTION_VIEW_OFFERS - user clicked "view offers" from either abbreviated show offers screens
  IMPRESSION_SUBSCRIPTION_COMPLETE - logged when the user successfully pays for a subscription

src/runtime/pay-flow.js
```diff
@@ -271,6 +271,10 @@ export class PayCompleteFlow {
    */
   complete() {
     this.eventManager_.logSwgEvent(AnalyticsEvent.ACTION_ACCOUNT_CREATED, true);
```
elijahsoria: nit: The spacing here seems off. Could you reformat? Although I don't know if this is due to github or the actual code spacing.

src/runtime/pay-flow.js
```diff
@@ -271,6 +271,10 @@ export class PayCompleteFlow {
...
+    this.eventManager_.logSwgEvent(
+      AnalyticsEvent.IMPRESSION_SUBSCRIPTION_COMPLETE,
```
elijahsoria: Maybe this should mean that the entire subscription flow is complete, regardless as to the user's subscription status as time of start, and not that a subscription was created. What I mean is that maybe this event should be logged in the case where a flow is completed when a user has an account already. The name of the event is a bit wonky, but it makes sense if we aren't trying to double log.

mborof:
Reverted changes to pay-client

PR #642, opened 2019-08-21
Analytics ignores subscription state change (for now)

mborof:
The subscription state change event was not intended to be part of our analytics service.  The client code should avoid sending the event back to the analytics service.


mborof:
Now that I'm at it, I am wondering if analytics service should ignore any other event types.  The primary candidate that comes to mind is EVENT_CUSTOM which is a custom event publishers are allowed to send to Propensity (for reasons unknown as I can't imagine a custom event would help you determine whether the use is going to subscribe or not....)

PR #643, opened 2019-08-21
This PR does 3 things:

qidonna:
- use log.js instead of console.log, 
- using console.* will trigger failures instead of warnings
- exclude log.js and log-test.js from lint


src/runtime/client-event-manager.js
```diff
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
+/* eslint-disable */
```
mborof: Why is this necessary?

qidonna: I'll remove it. It was added by mistake. I deleted before pushing, but the file was not automatically saved. Good catch!

PR #648, opened 2019-08-29
Generate a better UUID

mborof:
Uses a better method of generating random values.


third_party/random_uuid/uuid-swg.js
```diff
@@ -17,8 +17,20 @@ Dual licensed under the MIT and GPL licenses.
...
+);
+
+export function getRandomFloat() {
+  try {
+    const arr = new Uint32Array(2);
+    crypto.getRandomValues(arr);
+    const mantissa = arr[0] * Math.pow(2, 20) + (arr[1] >>> 12);
+    return mantissa * Math.pow(2, -52);
+  } catch (e) {
+    return Math.random(); //for older browsers
```
dvoytenko: What kind of exception are we expecting here? Can we just test for `typeof crypto && crypto.getRandomValues == 'function'`?

mborof: Makes sense, I just validated crypto && crypto.getRandomValues since it seems unlikely it will be some other random thing that isn't a function.

third_party/random_uuid/Math.uuid.js
```diff
@@ -31,62 +31,87 @@ Dual licensed under the MIT and GPL licenses.
...
-      // rfc4122, version 4 form
-      var r;
+/**
+ * Returns a random number between 0 and 1.
+ */
+export function getRandomFloat() {
+  if (crypto && crypto.getRandomValues) {
+    const arr = new Uint32Array(2);
+    crypto.getRandomValues(arr);
+    const mantissa = arr[0] * Math.pow(2, 20) + (arr[1] >>> 12);
```
dvoytenko: Should we extract both `Math.pow` into constants up top?

Generally, it's a bit black-magicy. I'm wondering if our live would be easier if we just based our implementation on the `Math.uuid` or `Math.uuidCompact` from that package?

mborof: Agreed on the consts.

I'm slightly hesitant to change which of the 3 implementations we use to generate a UUID because, as you say, its black magicky and I don't understand everything going into these.  They seem to indicate that the functional output is identical so it would probably be safe?  I don't have a strong preference for any of them though.  Maybe a *slight* preference for the compact one because it reads the clearest and I don't think we are doing it enough times that we really care about performance (once per page load, maybe a few other uses here and there at a later time?)

mborof: Though I kind of wonder (based on comments) if "minimizing calls to random() to boost performance" is causing the higher than expected intersectionality of UUIDs.

mborof: I decided that I agreed an replaced UUIDFast with UUID.  I think the readability of uuid is much improved with my changes so it is preferable over the slower one with the regexp.

mborof:
I've removed the third_party code base and made enough changes to it all that I think we can call it our code.

mborof:
I decided to remove the non-RFC UUID since we aren't using it anyway.  I also beefed up the test expectations and improved naming in general.

src/utils/random.js
```diff
@@ -21,17 +21,25 @@
  * @param {!number} maxVal
  */
 export function getRandomInts(numInts, maxVal) {
-  const arr = new Uint32Array(numInts);
+  //ensure array type is appropriate for the max value (performance)
```
dvoytenko: Please add space and capitalize after `//`.

src/utils/string.js
```diff
@@ -14,6 +14,10 @@
  * limitations under the License.
  */
 
+import {getRandomInts} from './random';
+
+const CHARS = '0123456789ABCDEF'.split('');
```
dvoytenko: Minor, but `split` is not needed here. Chars in a string are addressable by `[i]`

src/utils/string.js
```diff
@@ -109,3 +113,34 @@ export function stringHash32(str) {
...
+  return CHARS[(v & 0x3) | 0x8];
+}
+
+/**
+ * Generates a rfc4122v4 uuid. Ex:
+ *   "92329D39-6F5C-4520-ABFC-AAB64544E172"
+ */
+export function getUuid() {
+  const uuid = new Array(36);
+  let i;
```
dvoytenko: Why is `i` not scoped to the for-each?

PR #649, opened 2019-08-29
Analytics config

jpettitt:
See b/137212492


src/runtime/runtime.js
```diff
@@ -646,27 +645,36 @@ export class ConfiguredRuntime {
...
+          }
+          break;
+        default:
+          error = 'Unknown config property: ' + k;
       }
     }
-    if (error) {
-      throw new Error(error);
-    }
+    // Throw error string if's set
```
sohanirao: if it's set?

src/runtime/runtime-test.js
```diff
@@ -865,11 +865,7 @@ describes.realWin('Runtime', {}, env => {
     });
 
     it('should return propensity module', () => {
-      const propensity = new Propensity(
-        win,
-        config,
-        configuredRuntime.eventManager()
-      );
+      const propensity = new Propensity(win, configuredRuntime, new Fetcher());
```
sohanirao: There's a fundamental issue with this: We  architected the Propensity module to be independent of runtime - in other words not have any dependency on internal SwG modules. This is why we pass in the event manager, which also has no dependency on SwG modules. Deps is useful for dependency injection during test, implemented in Runtime. So the idea was to insulate propensity from that. However, if you are passing in deps, you don't need to pass in win either you can get it internally as deps.win().

jpettitt: good point - added a #todo for fixit week.

PR #650, opened 2019-09-04
Make UUID Time-based

mborof:
Ensures the analytics transaction ID can never collide with any other value generated at a different MS of the month.


src/utils/string.js
```diff
@@ -123,17 +123,44 @@ function getChar19(v) {
...
+ */
+function getMsSinceMidnight(date) {
+  date = date || new Date();
+  return date.getTime() - date.setHours(0, 0, 0, 0);
+}
+
+/**
+ * The returned identifier will always be an 8 digit valid hexidecimal number
+ * and will be unique down to the MS within a given month.  For example,
+ * exactly midnight on the first of any month will be '00000000' and it will
```
jpettitt: why not just getTime() and take the least significant 8 hex digits of it?  We don't care about the absolute value just that it's distinct in any give 28 day period.

PR #652, opened 2019-09-05
include addToFixedLayer() in Doc for AMP compatibility

jpettitt:
fixes issue where swg dialogs are broken by amp fixed layer handling


jpettitt:
will require code from https://github.com/subscriptions-project/swg-js/pull/652 when we push this to AMP

src/components/dialog.js
```diff
@@ -152,11 +153,15 @@ export class Dialog {
     } else {
       this.show_();
     }
-
-    return iframe.whenReady().then(() => {
-      this.buildIframe_();
-      return this;
-    });
+    return this.doc_
+      .addToFixedLayer(iframe.getElement(), /* force transfer */ true)
```
dvoytenko: Hmm. Should we replace `this.doc_.getBody().appendChild(...)` above with `addToFixedLayer`. Basically that's the meaning of the `addToFixedLayer()` really - it adds the element into some sort of prioritized container.

src/model/doc.js
```diff
@@ -62,6 +62,17 @@ export class Doc {
...
+  /**
+   * Adds the element to the fixed layer.
+   * @param {!Element} unusedElement
+   * @param {boolean=} opt_forceTransfer If set to true, then the element needs
+   *    to be forcefully transferred to the fixed layer.
+   * @return {!Promise}
+   *
+   * This is a no-op for except in AMP on iOS < 13.0.
+   */
+  addToFixedLayer(unusedElement, opt_forceTransfer) {}
```
dvoytenko: API wise we should probably delegate `forceTransfer` to the `Doc` implementation in AMP and always make it `true`.

src/components/dialog.js
jpettitt: makes sense, will do

src/model/doc.js
jpettitt: will do

jpettitt:
Will need https://github.com/ampproject/amphtml/pull/24385 integrated withe it when pushed to AMP

src/components/dialog.js
```diff
@@ -140,22 +140,22 @@ export class Dialog {
...
-    this.graypane_.attach();
+    return this.doc_.addToFixedLayer(iframe.getElement()).then(() => {
+      this.graypane_.attach();
 
-    if (hidden) {
-      setImportantStyles(iframe.getElement(), {
-        'visibility': 'hidden',
-        'opacity': 0,
+      if (hidden) {
+        setImportantStyles(iframe.getElement(), {
```
dvoytenko: This branch of `if()` would be one thing I'd recommend moving above `return this.doc_.addToFixedLayer()...`. Otherwise, I'd have a concern that iframe can flash on the screen before hidden here. In other words, let's keep this part of code synchronous.

jpettitt: sgtm

jpettitt: moot as I had to revert to the old way where it's attached before the xfer to fixed laer - AMP will not take it if it's not attached to the doc.

PR #666, opened 2019-09-13
Updated api_messages_proto files with oldSku

Kestrad:


ChrisAntaki:
Hey! I just did some research on the cause. It appears like the `});` from line 195 was removed, which causes a syntax error. Then the indentation changes are a result of `prettier` autoformatting while being affected by the syntax error.

ChrisAntaki:
My guess is this happened while merging with the `master` branch. I bet recompiling the files would fix it.

PR #674, opened 2019-09-23
Result handling API to use Protocol buffers for Activity result data

sohanirao:


ChrisAntaki:
Thanks for your great work on this! I'm sorry we didn't have the bandwidth to finish it

PR #686, opened 2019-10-11
Script to encrypt HTML pages for SwG Encryption Project

elijahsoria:
Golang script which takes in an input HTML document and encrypts
all content within <section subscriptions-section="content" encrypted>
tags using AES-GCM. The key used to encrypt the content is added
to the output document's head inside of a 
<script cryptokeys type="application/json"> element. The encrypted
document is outputted to the output_file path given as a flag.


tools/encryption/encryptionutils/encryption-helper.go
```diff
@@ -0,0 +1,243 @@
...
+    return tinkpb.Keyset{
+		PrimaryKeyId: 1,
+		Key:          keys,
+	}
+}
+
+// Retrieves all encrypted content sections from the parsed HTML tree.
+func getAllEncryptedSections(parsed_html *html.Node) []*html.Node {
+	var encrypted_sections []*html.Node
+	for n := parsed_html.FirstChild; n != nil; n = n.NextSibling {
```
alin04: There are a lot of for loops in here. I got a little lost. Is this trying to traverse the entire tree depth first? There are a couple of ways do it. One is recursion. Or there's a simpler way to traverse the tree. I coded it for the amppackager [here](https://github.com/ampproject/amppackager/blob/releases/transformer/internal/htmlnode/htmlnode.go#L47).
Example usage [here](https://github.com/ampproject/amppackager/blob/releases/transformer/transformers/nodecleanup.go#L47).

alin04: Oh, I see, this is trying to find the <body> and then traverse everything within. Recommend splitting out FindBody from  the processing that's being done.

tools/encryption/encryptionutils/encryption-helper.go
```diff
@@ -0,0 +1,243 @@
...
+	tinkpb "github.com/google/tink/proto/tink_go_proto"
+	"net/http"
+	"strings"
+)
+
+const AES_GCM_KEY_URL string = "type.googleapis.com/google.crypto.tink.AesGcmKey" 
+const AES_GCM_KEY_SIZE uint32 = 16
+
+// Public function to generate an encrypted HTML document given the original.
+func GenerateEncryptedDocument(html_str string, public_key_url string, access_requirement string) (string, error) {
```
alin04: This method is stateless. Could it be more stateful so it can encrypt multiple documents? Seems reasonable that a user would want to encrypt more than one?

By state, I mean it can save the Google Public Key (so it's not re-retrieving it over and over). I'm not sure if the AesKey has to be regenerated per doc?

tools/encryption/encryptionutils/encryption-helper.go
```diff
@@ -0,0 +1,243 @@
...
+	if err != nil {
+		return "", err
+	}
+	ks := createAesGcmKeyset(key)
+	r := strings.NewReader(html_str)
+	parsed_html, err := html.Parse(r)
+	if err != nil {
+		return "", err
+	}
+	encrypted_sections := getAllEncryptedSections(parsed_html)
```
alin04: Move all parsing and encryption to as late as possible. IOW, set up your required dependencies first (key, keyset, publicKey).

tools/encryption/encryptionutils/encryption-helper.go
```diff
@@ -0,0 +1,243 @@
...
+}
+
+// Renders the input Node to a string.
+func renderNode(n *html.Node, trim bool) string {
+	var buf bytes.Buffer
+	w := io.Writer(&buf)
+	html.Render(w, n)
+	s := buf.String()
+	if trim {
+		s = strings.TrimPrefix(s, "<html><body>")
```
alin04: Don't think this will catch everything. What if html element has attributes?

tools/encryption/encryptionutils/encryption-helper.go
```diff
@@ -0,0 +1,243 @@
...
+					jsonData := fmt.Sprintf(`{"google.com":"%s"}`, encrypted_key)
+					text_node := &html.Node{Type: html.TextNode, Data: jsonData}
+					crypto_keys.AppendChild(text_node)
+					cn.AppendChild(crypto_keys)
+			    	return nil
+				}
+			}
+		}
+	}
+	return fmt.Errorf("Could not add cryptokeys to head.")
```
alin04: This also prints. Prefer to have callers handle the error. So instead return errors.New().

tools/encryption/encryption-script.go
```diff
@@ -0,0 +1,54 @@
+/* Script to encrypt documents for the SwG Encryption Project.
```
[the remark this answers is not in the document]
elijahsoria: Done.

tools/encryption/encryptionutils/encryption-helper.go
elijahsoria: Done.

tools/encryption/encryptionutils/encryption-helper.go
elijahsoria: I have to remove these strings from the output because html.Render adds them in automatically.

Do you know of a way to render the content inside of the html.Node that doesn't add any extra tags?

tools/encryption/encryptionutils/encryption-helper.go
```diff
@@ -0,0 +1,243 @@
...
+		PrimaryKeyId: 1,
+		Key:          keys,
+	}
+}
+
+// Retrieves all encrypted content sections from the parsed HTML tree.
+func getAllEncryptedSections(parsed_html *html.Node) []*html.Node {
+	var encrypted_sections []*html.Node
+	for n := parsed_html.FirstChild; n != nil; n = n.NextSibling {
+		if (n.Data == "html") && (len(n.Attr) != 0) {
```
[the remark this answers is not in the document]
elijahsoria: Done.

tools/encryption/encryptionutils/encryption-helper.go
```diff
@@ -0,0 +1,243 @@
...
+	if err != nil {
+		return "", err
+	}
+	key, err := generateNewAesGcmKey(keyManager)
+	if err != nil {
+		return "", err
+	}
+	ks := createAesGcmKeyset(key)
+	r := strings.NewReader(html_str)
+	parsed_html, err := html.Parse(r)
```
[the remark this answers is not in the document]
elijahsoria: Done.

tools/encryption/encryption-script.go
```diff
@@ -0,0 +1,54 @@
...
+	// Input flags.
+	html_file := flag.String("input_html_file", "", "Input HTML file to encrypt.")
+	out_file := flag.String("output_file", "", "Output path to write encrypted HTML file.")
+	access_requirement := flag.String("access_requirement", "", "The access requirement we grant upon decryption.")
+	google_public_key_url := flag.String("google_public_key_url",
+			"https://news.google.com/swg/encryption/keys/dev/tink/public_key",
+			"URL to Google's public key.")
+	flag.Parse()
+	if *html_file == "" || *out_file == "" || *google_public_key_url == "" || *access_requirement == "" {
+		fmt.Println("Missing flags!")
```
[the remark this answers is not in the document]
elijahsoria: Done.

tools/encryption/encryption-script.go
```diff
@@ -0,0 +1,54 @@
...
+		os.Exit(42)
+	}
+	// Read the input HTML file.
+	b, err := ioutil.ReadFile(*html_file)
+    if err != nil {
```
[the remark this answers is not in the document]
elijahsoria: Done.

tools/encryption/encryptionutils/encryption-helper.go
```diff
@@ -0,0 +1,243 @@
...
+	"github.com/google/tink/go/insecurecleartextkeyset"
+	"github.com/google/tink/go/keyset"
+	"github.com/google/tink/go/core/registry"
+	gcmpb "github.com/google/tink/proto/aes_gcm_go_proto"
+	tinkpb "github.com/google/tink/proto/tink_go_proto"
+	"net/http"
+	"strings"
+)
+
+const AES_GCM_KEY_URL string = "type.googleapis.com/google.crypto.tink.AesGcmKey"
```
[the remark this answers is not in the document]
elijahsoria: Done.

tools/encryption/encryption-script.go
```diff
@@ -0,0 +1,54 @@
...
+		os.Exit(42)
+	}
+	// Write the encrypted document to the output file.
+	f, err := os.Create(*out_file)
+	if err != nil {
+		fmt.Println(err)
+		os.Exit(42)
+	}
+	f.WriteString(encrypted_doc)
+	fmt.Println("Encrypted HTML file generated successfully")
```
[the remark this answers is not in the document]
elijahsoria: Done.

tools/encryption/encryptionutils/encryption-helper.go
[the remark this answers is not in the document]
elijahsoria: Shortened argument lists and some variable names. Thanks for the advice.

tools/encryption/encryption-script.go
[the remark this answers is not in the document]
elijahsoria: Thanks for the advice. I think I'll go with putting the script itself in cmd subdirectory and then the helper functions in an internal subdirectory.

tools/encryption/encryptionutils/encryption-helper.go
elijahsoria: I agree this is a confusing function. I moved the DFS function into it's own function--hopefully that clears things up.

elijahsoria: I agree it is quite confusing. I moved the DFS algorithm into its own function to clear things up.

tools/encryption/encryptionutils/encryption-helper.go
elijahsoria: Moved key handling up in the flow.

tools/encryption/encryptionutils/encryption-helper.go
elijahsoria: I modified GenerateEncryptedDocument so that it accepts a public key as input, which means that I can call GenerateEncryptedDocument multiple times with the same key in the script.

elijahsoria: To confirm, the AesKey has to be generated per document.

tools/encryption/encryptionutils/encryption-helper.go
alin04: Parse() is what adds the missing tags. Render on a node should simply render that node (and any of its children).

If you run this script, it outputs "<p>foobar</p>".

https://play.golang.org/p/RkjIaSEMGtK

elijahsoria: Thanks! Switched to Render and removed the trim option.

tools/encryption/pkg/encryption-helper.go
```diff
@@ -0,0 +1,263 @@
...
+		return "", err
+	}
+	if err = addEncryptedDocumentKeyToHead(encryptedKey, parsedHtml); err != nil {
+		return "", err
+	}
+	return renderNode(parsedHtml, false), nil
+}
+
+// Retrieves Google's public key from the given URL.
+func RetrieveGooglePublicKey(publicKeyUrl string) (tinkpb.Keyset, error) {
```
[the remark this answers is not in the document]
elijahsoria: I think I prefer the two function setup I have already. I worry that adding in an interface makes using the script more confusing because of Go's non-standard object-oriented implementation. Having only two functions to use seems easier for others to understand, which is important considering the folks using these scripts might not have any background in go (which is the reason we wanted to use JS originally).

tools/encryption/cmd/encryption-script.go
```diff
@@ -0,0 +1,78 @@
+/* Copyright 2018 The Subscribe with Google Authors. All Rights Reserved.
+ *
```
dvoytenko: 2019

tools/encryption/cmd/encryption-script.go
```diff
@@ -0,0 +1,78 @@
...
+* to the output document's head inside of a
+* <script cryptokeys type="application/json"> element. The encrypted
+* document is outputted to the output_file path given as a flag.
+ */
+func main() {
+	// Input flags.
+	inputHtmlFile := flag.String("input_html_file", "", "Input HTML file to encrypt.")
+	outFile := flag.String("output_file", "", "Output path to write encrypted HTML file.")
+	accessRequirement := flag.String("access_requirement", "", "The access requirement we grant upon decryption.")
+	googlePublicKeyUrl := flag.String("google_public_key_url",
```
dvoytenko: We probably need to instead provide a map of keys. E.g. google, internal, etc.

tools/encryption/pkg/encryption-helper.go
```diff
@@ -0,0 +1,263 @@
...
+		return nil, err
+	}
+	m, err := km.NewKey(p)
+	if err != nil {
+		return nil, err
+	}
+	return proto.Marshal(m)
+}
+
+// Creates an AES-GCM Keyset using the input key.
```
dvoytenko: Why does this need to be a keyset as opposed to a single key?

tools/encryption/pkg/encryption-helper.go
```diff
@@ -0,0 +1,263 @@
...
+			}
+		}
+	}
+	return nil
+}
+
+// Searches for all <section subscriptions-section="content" encrypted> nodes and returns them.
+func getEncryptedSectionsDfs(bodyNode *html.Node) []*html.Node {
+	var encryptedSections []*html.Node
+	var stack []*html.Node
```
dvoytenko: This looks more like a "queue" to me, not a stack per se.

tools/encryption/pkg/encryption-helper.go
```diff
@@ -0,0 +1,263 @@
...
+	if err != nil {
+		return err
+	}
+	for _, node := range encryptedSections {
+		var content []string
+		for c := node.FirstChild; c != nil; c = c.NextSibling {
+			content = append(content, renderNode(c, true))
+			node.RemoveChild(c)
+		}
+		encContent, err := cipher.Encrypt([]byte(strings.Join(content, "")), nil)
```
dvoytenko: My old question: can you confirm that the content joined here is all in UTF-8?

tools/encryption/pkg/encryption-helper.go
```diff
@@ -0,0 +1,263 @@
...
+func encryptDocumentKey(docKeyset, accessRequirement string, ks tinkpb.Keyset) (string, error) {
+	handle, err := keyset.NewHandleWithNoSecrets(&ks)
+	if err != nil {
+		return "", err
+	}
+	he, err := hybrid.NewHybridEncrypt(handle)
+	if err != nil {
+		return "", err
+	}
+	jsonStr := fmt.Sprintf("{\"accessRequirements\": [\"%s\"], \"key\": \"%s\"}", accessRequirement, docKeyset)
```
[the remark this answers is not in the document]
dvoytenko: Agree: either should be encoded or error'd out if contains not-allowed characters.

tools/encryption/pkg/encryption-helper.go
```diff
@@ -0,0 +1,263 @@
...
+						html.Attribute{Key: "type", Val: "application/json"},
+						html.Attribute{Key: "cryptokeys", Val: ""},
+					}
+					cryptoKeys := &html.Node{
+						Type:     html.ElementNode,
+						Data:     "script",
+						DataAtom: atom.Script,
+						Attr:     attrs,
+					}
+					jsonData := fmt.Sprintf(`{"google.com":"%s"}`, encryptedKey)
```
dvoytenko: Possibly coming in the followup PRs? But since we are doing one-pass thing here, the document will not be useful w/o publisher-own encrypted key here. And it won't be possible to add it on a second pass - they should all be prepared at the same time.

tools/encryption/cmd/encryption-script.go
```diff
@@ -0,0 +1,78 @@
+/* Copyright 2018 The Subscribe with Google Authors. All Rights Reserved.
```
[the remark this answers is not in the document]
elijahsoria: After offline conversation, we settled on the following structure:

swg-js/tools/golang/cmd/encrypt/script.go
swg-js/tools/golang/pkg/encryption/encryption.go

with encryption.go being "package encryption".

tools/encryption/pkg/encryption-helper.go
elijahsoria: This is due to a specification by the Tink team. I'll add you into the email discussion.

tools/encryption/pkg/encryption-helper.go
elijahsoria: Switched to using "queue" instead.

tools/encryption/pkg/encryption-helper.go
elijahsoria: I am looking to do this in a followup PR, yes. I would like to talk to some other folks about how publishers will be providing their public keys. I am thinking it would either be hosted on an SCS like our public key, or  by direct command line insertion.

tools/encryption/pkg/encryption-helper.go
dvoytenko: Any chance you could explicate in comments some examples of what serialized form of each part looks like?

tools/encryption/pkg/encryption-helper.go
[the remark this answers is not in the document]
elijahsoria: Added in a call to html/template.HTMLEscapeString before inserting accessRequirements into the JSON string.

tools/encryption/pkg/encryption-helper.go
dvoytenko: I don't think we will have their keys on our servers. I think, generically, it'd be a URL. But possibly more practically a keystore on the disk?

tools/encryption/cmd/encryption-script.go
elijahsoria: Right, I have some questions about how we expect publishers to post their public keys. Hopefully we can discuss offline.

tools/encryption/pkg/encryption-helper.go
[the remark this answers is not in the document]
dvoytenko: HTML escaping and JSON escaping are completely different :)

tools/encryption/pkg/encryption-helper.go
elijahsoria: Added a call to utf8.Valid before encrypting.

dvoytenko: This would seem to be a wrong place to check for valid UTF8 since you've already processed the input stream and parsed the HTML which inherently has to decode a byte stream into a char stream. Otherwise it'd not be able to scan characters such as `<` and `>`.

tools/golang/pkg/encryption/encryption.go
```diff
@@ -0,0 +1,277 @@
...
+						html.Attribute{Key: "type", Val: "application/json"},
+						html.Attribute{Key: "cryptokeys", Val: ""},
+					}
+					cryptoKeys := &html.Node{
+						Type:     html.ElementNode,
+						Data:     "script",
+						DataAtom: atom.Script,
+						Attr:     attrs,
+					}
+					jsonData := fmt.Sprintf(`{"google.com":"%s"}`, encryptedKey)
```
dvoytenko: Let's do that json.Marshal-y thing here as well.

tools/encryption/pkg/encryption-helper.go
dvoytenko: The only thing that really matters is the original stream that goes into html.Parse. I think we just need to make sure our tests (manual and/or otherwise) have characters from non-ASCII set.

tools/encryption/pkg/encryption-helper.go
dvoytenko: I definitely think we can start with URLs at this time, with `file:` URLs being a special case for now if we need for manual tests. Then we can expand. It's fine if we do it in followup PRs - this PR is already pretty big.

tools/encryption/pkg/encryption-helper.go
elijahsoria: Added an example proto in the comments of the function.

tools/encryption/pkg/encryption-helper.go
elijahsoria: I added support for publishers to input their domain and URL to the hosted public key as command line flags. PTAL.

tools/encryption/pkg/encryption-helper.go
elijahsoria: Acknowledged.

tools/golang/pkg/encryption/encryption.go
```diff
@@ -0,0 +1,295 @@
+/* Copyright 2019 The Subscribe with Google Authors. All Rights Reserved.
```
alin04: This file needs unit tests.

tools/golang/cmd/encrypt/script.go
```diff
@@ -0,0 +1,122 @@
...
+func main() {
+	// Input flags.
+	inputHtmlFile := flag.String("input_html_file", "", "Input HTML file to encrypt.")
+	outFile := flag.String("output_file", "", "Output path to write encrypted HTML file.")
+	accessRequirement := flag.String("access_requirement", "", "The access requirement we grant upon decryption.")
+	googlePublicKeyUrl := flag.String("google_public_key_url",
+		"https://news.google.com/swg/encryption/keys/dev/tink/public_key",
+		"URL to Google's public key.")
+	mf := make(mapFlags)
+	flag.Var(&mf, "publisher_public_key_urls", `Strings in the form of '<domain-name>,<url>', where url is
```
dvoytenko: Why is it plural?

dvoytenko: To confirm: the `googlePublicKeyUrl` appears to be a subset of this flag? Maybe we should just use one flag? We can always resolve defaults separately.

elijahsoria: I originally made it plural since this flag can be used more than once, but I'll change it to singular.

elijahsoria: I originally made it plural since this flag can be used more than once, but I'll change it to singular.

elijahsoria: Correct, `googlePublicKeyUrl` is used in the same way as `publisher_public_key_url`, but I prefer splitting them into two because it makes it clear that the google public key and the publisher public keys are both needed. I worry if we just leave it to one map then publishers will forget to input the google public key.

dvoytenko: My idea was actually to keep it plural. Name it `--encryption-key-url=<name>,url` and bake Google's one into that. E.g.

```
encrypt ... \
  --encryption-key-url=google.com,https://news.google.com/swg/key \
  --encryption-key-url=local,https://norcal.com/key \
  --encryption-key-url=...
```

Makes sense?

elijahsoria: Done.

elijahsoria: Made publisher key input a map with a default for the Google key.

tools/golang/pkg/encryption/encryption.go
elijahsoria: Added unit tests.

tools/golang/pkg/encryption/encryption_test.go
```diff
@@ -0,0 +1,134 @@
...
+		if n != len([]byte(googPublicKeyStr)) {
+			t.Errorf("Wrote %d bytes of fake response; want %d", n, len([]byte(googPublicKeyStr)))
+		}
+	}))
+	defer httpServer.Close()
+	pubKey, err := RetrieveTinkPublicKey(httpServer.URL)
+	if err != nil {
+		t.Errorf("Failed to retrieve Tink public key.")
+	}
+	if pubKey.PrimaryKeyId != 3962548922 {
```
alin04: Declare a const var instead of having these magic numbers in the test.

tools/golang/pkg/encryption/encryption_test.go
```diff
@@ -0,0 +1,134 @@
...
+		t.Errorf("Invalid primary key ID: %d", pubKey.PrimaryKeyId)
```
alin04: Use got/want semantics.

tools/golang/pkg/encryption/encryption_test.go
```diff
@@ -0,0 +1,134 @@
...
+	_, err := RetrieveTinkPublicKey(httpServer.URL)
+	if err == nil {
+		t.Errorf("Expected failure but call succeeded.")
+	}
+}
+
+func TestEncryptDocumentSuccess(t *testing.T) {
+	htmlStr := loadTestFileString("sample_encryption.html")
+	if htmlStr == "" {
+		t.Errorf("HTML file load failed.")
```
alin04: t.Fatalf

tools/golang/pkg/encryption/encryption.go
```diff
@@ -0,0 +1,295 @@
...
+	}
+	parsedHtml, err := html.Parse(strings.NewReader(htmlStr))
+	if err != nil {
+		return "", err
+	}
+	encryptedSections := getAllEncryptedSections(parsedHtml)
+	if err = encryptAllSections(parsedHtml, encryptedSections, kh); err != nil {
+		return "", err
+	}
+	encryptedKeys, err := encryptDocumentKey(base64.StdEncoding.EncodeToString(ksEnc), accessRequirement, pubKeys)
```
alin04: Return early if there are no encrypted sections.

tools/golang/pkg/encryption/encryption_test.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+	if !strings.Contains(encDoc, "google.com") {
+		t.Errorf("Missing google.com key.")
+	}
+}
+
+func TestEncryptDocumentEmptyKeyset(t *testing.T) {
+	htmlStr := loadTestFileString("sample_encryption.html")
+	if htmlStr == "" {
+		t.Errorf("HTML file load failed.")
```
alin04: t.Fatalf

tools/golang/pkg/encryption/encryption_test.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+	pubKeys := map[string]tinkpb.Keyset{
+		"google.com": tinkpb.Keyset{},
+	}
+	_, err := GenerateEncryptedDocument(htmlStr, "norcal.com:premium", pubKeys)
+	if err == nil {
+		t.Errorf("Error did not occur on empty Keyset.")
+	}
+}
```
alin04: For thoroughness, add a test for malformed/unparseable HTML. And a test if there are no encrypted sections (meaning no key needs to be added to head).

tools/golang/pkg/encryption/encryption.go
```diff
@@ -0,0 +1,295 @@
...
+			}
+		}
+	}
+	return errors.New("Could not add cryptokeys to head.")
+}
+
+// Renders the input Node to a string.
+func renderNode(n *html.Node) string {
+	var buf bytes.Buffer
+	w := io.Writer(&buf)
```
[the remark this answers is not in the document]
elijahsoria: Nice!

tools/golang/cmd/encrypt/script.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+	// Retrieve all public keys from the input URLs.
+	pubKeys := make(map[string]tinkpb.Keyset)
+	var pubKey tinkpb.Keyset
+	for domain, url := range mf {
+		pubKey, err = encryption.RetrieveTinkPublicKey(url)
+		if err != nil {
+			log.Fatal(err)
+		}
+		pubKeys[domain] = pubKey
```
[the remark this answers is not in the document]
elijahsoria: Good point. I'll add a ToLower call.

tools/golang/cmd/encrypt/script.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+	if _, ok := pubKeys["google.com"]; !ok {
```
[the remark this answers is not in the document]
elijahsoria: I agree the duplication doesn't look nice, but I added it because I don't want to override the user's chosen value for google.com, especially if they use the prod URL.

tools/golang/pkg/encryption/encryption_test.go
elijahsoria: I couldn't find a good failure case for unparsable HTML because the Go parser adds in a lot of missing tags when it can't find them. https://godoc.org/golang.org/x/net/html#Parse

I added a test for the case where there aren't any valid encryptable sections and when the content is in Hebrew (i.e. non-ASCII).

tools/golang/cmd/encrypt/script.go
```diff
@@ -0,0 +1,120 @@
...
+/* Script to encrypt documents for the SwG Encryption Project.
+*
+* Golang script which takes in an input HTML document and encrypts
+* all content within <section subscriptions-section="content" encrypted>
+* tags using AES-GCM. The key used to encrypt the content is added
+* to the output document's head inside of a
+* <script cryptokeys type="application/json"> element. The encrypted
+* document is outputted to the output_file path given as a flag.
+*
+* Example Usage:
```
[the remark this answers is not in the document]
elijahsoria: Added a README.md and added a comment about using go get.

tools/golang/cmd/encrypt/script.go
```diff
@@ -0,0 +1,120 @@
...
+*	--publisher_public_key_url=wp.com,www.wp.com/scs/publickey
+ */
+type mapFlags map[string]string
+
+const googleDevPublicKeyUrl string = "https://news.google.com/swg/encryption/keys/dev/tink/public_key"
+
+func (m *mapFlags) String() string {
+	var strs []string
+	for key, val := range *m {
+		strs = append(strs, key, ";", val)
```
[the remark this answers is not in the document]
elijahsoria: I'll use a comma in the String method here. I would like to avoid using the equals sign because I think that it can be confusing to have an equals sign for the flag and within the value itself. I can't use semi-colons in the flag because that is a command separator for the command line.

tools/golang/cmd/encrypt/script.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+	(*m)[s[0]] = s[1]
+	return nil
+}
+
+func main() {
+	// Input flags.
+	inputHtmlFile := flag.String("input_html_file", "", "Input HTML file to encrypt.")
+	outFile := flag.String("output_file", "", "Output path to write encrypted HTML file.")
+	accessRequirement := flag.String("access_requirement", "", "The access requirement we grant upon decryption.")
```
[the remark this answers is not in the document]
elijahsoria: I'm confused about the terminology to use here as well. In the SwG docs, it seems that we use product ID and entitlements for this, but we landed on "accessRequirement" as the key name within the JSON object which we store the keys. https://developers.google.com/news/subscribe/guides/set-up-skus#create-subscribe-with-google-skus

I would like to stick with "access_requirement" because this is consistent with the key name in the JSON object that gets added to the document.

tools/golang/cmd/encrypt/script.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+	// Read the input HTML file.
+	b, err := ioutil.ReadFile(*inputHtmlFile)
+	if err != nil {
+		log.Fatal(err)
+	}
+	// Retrieve all public keys from the input URLs.
+	pubKeys := make(map[string]tinkpb.Keyset)
+	var pubKey tinkpb.Keyset
+	for domain, url := range mf {
```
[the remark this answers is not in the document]
elijahsoria: We will only include the Google key in this case. During serving, we would then just call Google's key management system and not the publisher's. This will cause an issue once the document's keys are deprecated from Keystore, but that is in 5+ years. We will have to make note of this in the larger encryption document that we give to publishers.

tools/golang/cmd/encrypt/script.go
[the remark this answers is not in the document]
alin04: Can you check if they specify google.com before the for loop?

```
if _, ok := mf["google.com"]; !ok {
  mf["google.com"] = googleDevPublicKeyUrl
}
for domain, url := range mf {
  ...
}
```

tools/golang/cmd/encrypt/script.go
```diff
@@ -0,0 +1,102 @@
...
+	"github.com/subscriptions-project/swg-js/tools/golang/encryption"
+	"io/ioutil"
+	"log"
+	"os"
+	"strings"
+)
+
+type mapFlags map[string]string
+
+const googleDevPublicKeyUrl string = "https://news.google.com/swg/encryption/keys/dev/tink/public_key"
```
alin04: nit: googleDevPublicKeyURL

Acronyms in Go are constant case. https://github.com/golang/go/wiki/CodeReviewComments#initialisms

tools/golang/cmd/encrypt/script.go
```diff
@@ -0,0 +1,102 @@
...
+		return errors.New("Malformed value inserted: " + value)
+	}
+	(*m)[s[0]] = s[1]
+	return nil
+}
+
+// Script to encrypt documents for the SwG Encryption Project.
+func main() {
+	// Input flags.
+	inputHtmlFile := flag.String("input_html_file", "", "Input HTML file to encrypt.")
```
alin04: inputHTMLFile

tools/golang/pkg/encryption/encryption.go
```diff
@@ -0,0 +1,295 @@
...
+	"golang.org/x/net/html"
+	"golang.org/x/net/html/atom"
+	"net/http"
+	"strings"
+	"unicode/utf8"
+)
+
+// Helper functions for the SwG Encryption Script.
+
+const aesGcmKeyUrl string = "type.googleapis.com/google.crypto.tink.AesGcmKey"
```
alin04: aesGCMKeyURL

tools/golang/pkg/encryption/encryption.go
```diff
@@ -0,0 +1,295 @@
...
+	if err != nil {
+		return "", err
+	}
+	parsedHtml, err := html.Parse(strings.NewReader(htmlStr))
+	if err != nil {
+		return "", err
+	}
+	encryptedSections := getAllEncryptedSections(parsedHtml)
+	if len(encryptedSections) == 0 {
+		return "", errors.New("No encrypted sections found.")
```
alin04: Should this error or simply no-op?

tools/golang/cmd/encrypt/script.go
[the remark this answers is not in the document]
elijahsoria: Nice!

tools/golang/pkg/encryption/encryption.go
elijahsoria: Done. I think I got most of the others too.

tools/golang/pkg/encryption/encryption.go
elijahsoria: I like error out because the publisher is aware that nothing happened to the document instead of silently doing nothing or returning an empty string.

tools/golang/cmd/encrypt/README.md
```diff
@@ -0,0 +1,34 @@
...
+```
+    
+Or...
+
+```
+go run swg-js/tools/golang/cmd/encrypt/script.go \
+	--input_html_file=../tmp/sample-encryption.html \
+	--output_file=../tmp/sample-encryption-out.html \
+	--google_public_key_url=https://news.google.com/swg/encryption/keys/{dev|prod}/tink/public_key \
+	--access_requirement=norcal.com:premium \
```
alin04: Same here. thenews.com:premium

tools/golang/cmd/encrypt/README.md
```diff
@@ -0,0 +1,34 @@
...
+    --publisher_public_key_url=thenews.com,www.thenews.com/scs/publickey
+```
+    
+Or...
+
+```
+go run swg-js/tools/golang/cmd/encrypt/script.go \
+	--input_html_file=../tmp/sample-encryption.html \
+	--output_file=../tmp/sample-encryption-out.html \
+	--google_public_key_url=https://news.google.com/swg/encryption/keys/{dev|prod}/tink/public_key \
```
alin04: This isn't a flag anymore right? Should one of the public_key_urls below be google? Either replace example.com or thenews.com.

tools/golang/cmd/encrypt/script.go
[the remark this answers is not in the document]
alin04: Where is "access_requirement" used in JSON? I didn't find it in the link you referenced.

tools/golang/pkg/encryption/encryption.go
```diff
@@ -0,0 +1,295 @@
...
+	"unicode/utf8"
+)
+
+// Helper functions for the SwG Encryption Script.
+
+const aesGCMKeyURL string = "type.googleapis.com/google.crypto.tink.AesGcmKey"
+const aesGCMKeySize uint32 = 16
+
+// Public function to generate an encrypted HTML document given the original.
+func GenerateEncryptedDocument(htmlStr, accessRequirement string, pubKeys map[string]tinkpb.Keyset) (string, error) {
```
alin04: How do you plan to version/evolve this over time? Backwards compatibility? Versioned encryption tag?

tools/golang/pkg/encryption/encryption.go
```diff
@@ -0,0 +1,295 @@
...
+	ks := createAesGcmKeyset(key)
+	ksEnc, err := proto.Marshal(&ks)
+	if err != nil {
+		return "", err
+	}
+	kh, err := insecurecleartextkeyset.Read(&keyset.MemReaderWriter{Keyset: &ks})
+	if err != nil {
+		return "", err
+	}
+	parsedHtml, err := html.Parse(strings.NewReader(htmlStr))
```
alin04: nit: parsedHTML

yes, it's annoying.

tools/golang/pkg/encryption/encryption.go
```diff
@@ -0,0 +1,295 @@
...
+					return getEncryptedSectionsDfs(bn)
+				}
+			}
+		}
+	}
+	return nil
+}
+
+// Searches for all <section subscriptions-section="content" encrypted> nodes and returns them.
+func getEncryptedSectionsDfs(bodyNode *html.Node) []*html.Node {
```
alin04: nit: this should be DFS. Yes, annoying. literally every acronym should be constant case.

tools/golang/cmd/encrypt/script.go
alin04: Not changed? Or maybe i'm looking at an old branch.

tools/golang/cmd/encrypt/README.md
elijahsoria: You're right. Updated flag names and values.

tools/golang/pkg/encryption/encryption.go
elijahsoria: I imagine that this will largely happen on the backend or in the AMP runtime. We can support multiple versions inside of the AMP Cache, and then handle each version specifically with the AMP extension: https://github.com/ampproject/amphtml/blob/master/extensions/amp-subscriptions/0.1/crypto-handler.js

Right now, we are supporting encrypted documents for up to five years after encryption. After that, we will redirect to origin.

tools/golang/cmd/encrypt/README.md
```diff
@@ -0,0 +1,34 @@
...
+```<script cryptokeys type="application/json">``` element. The encrypted
+document is outputted to the output_file path given as a flag.
+
+## Example Usage:
+
+```
+go get github.com/subscriptions-project/swg-js/tools/golang/cmd/encrypt && \
+$GOPATH/bin/encrypt \
+    --input_html_file=../tmp/sample-encryption.html \
+	--output_file=../tmp/sample-encryption-out.html \
```
alin04: Indentation looks off for both these example commands.

elijahsoria: Fixed.

PR #696, opened 2019-10-21
update proto

mborof:
Add new enum value to proto


src/proto/api_messages.js
```diff
@@ -1053,3 +1054,4 @@ export {
   deserialize,
   getLabel,
 };
+
```
elijahsoria: Trailing newline.

PR #698, opened 2019-10-21
Pass Transaction ID to PaymentsAsyncClient before loadPaymentData

qidonna:
This PR does the following:
1. pass transaction ID in i param in PaymentsAsyncClient initialization option
2. pass the same transaction ID to loadPaymentData
3. set PaymentsAsyncClient.googleTransactionId_ = the same string before new PaymentsAsyncClient() is called

This should fix the issue that transaction ID dismatch between payment and SwG log for native buy flow


src/runtime/pay-client-test.js
```diff
@@ -407,7 +454,7 @@ describes.realWin('PayClientBindingPayjs', {}, env => {
     payClientStubs = {
       create: sandbox
         .stub(PayClientBindingPayjs.prototype, 'createClient_')
-        .callsFake((options, handler) => {
+        .callsFake((options, googleTransactionId, handler) => {
           responseHandler = handler;
```
mborof: Its probably not an issue (and not caused by your development if it is) but should we be using options or googleTransactionId here?  If we have some other method of using it, maybe it would be cleaner to do so here.

src/runtime/pay-client-test.js
```diff
@@ -438,14 +511,22 @@ describes.realWin('PayClientBindingPayjs', {}, env => {
...
+      ).getType()
+    ).to.equal('PAYJS');
   });
 
   it('should initalize correctly', () => {
     expect(payClientStubs.create).to.be.calledOnce.calledWith({
       'environment': '$payEnvironment$',
       'i': {
         'redirectKey': 'test_restore_key',
+        'googleTransactionId': 'testid123',
```
mborof: Let's replaced testid123 with a const (GOOGLE_TRANSACTION_ID?)

src/runtime/pay-client-test.js
```diff
@@ -378,17 +413,29 @@ describes.realWin('PayClientBindingSwg', {}, env => {
...
   let responseHandler;
+  let googleTransactionId;
 
   beforeEach(() => {
+    deps = new DepsDef();
     win = env.win;
     activityPorts = new ActivityPorts(win);
+    sandbox.stub(deps, 'win').callsFake(() => win);
+
+    googleTransactionId = 'testid123';
```
mborof: NIT: Let's replaced testid123 with a const (GOOGLE_TRANSACTION_ID?)

src/runtime/pay-client-test.js
```diff
@@ -75,15 +87,38 @@ describes.realWin('PayClientBindingSwg', {}, env => {
...
+    pageConfig = new PageConfig('pub1:label1');
+    sandbox.stub(deps, 'pageConfig').callsFake(() => pageConfig);
+
+    eventManager = new ClientEventManager(Promise.resolve);
+    sandbox.stub(deps, 'eventManager').callsFake(() => eventManager);
+
+    analyticsService = new AnalyticsService(deps);
+    analyticsMock = sandbox.mock(analyticsService);
+    sandbox.stub(analyticsService, 'getTransactionId').callsFake(() => {
+      return 'testid123';
```
mborof: NIT: Let's replaced testid123 with a const (GOOGLE_TRANSACTION_ID?)

src/runtime/pay-client.js
```diff
@@ -65,11 +65,14 @@ export class PayClient {
    * @param {!Window} win
    * @param {!../components/activities.ActivityPorts} activityPorts
    * @param {!../components/dialog-manager.DialogManager} dialogManager
+   * @param {!../runtime/analytics-service.AnalyticsService} analyticsService
    */
-  constructor(win, activityPorts, dialogManager) {
+  constructor(win, activityPorts, dialogManager, analyticsService) {
```
mborof: Will it ever need the entire analytics service or can we just pass the transaction ID?  Might make some of your tests simpler with just the ID.

src/runtime/pay-flow.js
```diff
@@ -138,7 +138,7 @@ export class PayStartFlow {
         'swg': swgPaymentRequest,
         'i': {
           'startTimeMs': Date.now(),
-          'googleTransactionId': this.analyticsService_.getTransactionId(),
+          'googleTransactionId': this.payClient_.googleTransactionId_,
```
mborof: Question: Are you absolutely certain that all paths set payClient_.googleTransactionId_?  It seemed like there were 2 major paths and one of them had fewer parameters.  I didn't trace the code closely enough to answer this question.  The existing code should always have the correct value so it seems safer to leave it as is.  Certainly arguable either way though.

src/runtime/runtime.js
```diff
@@ -535,9 +539,6 @@ export class ConfiguredRuntime {
     /** @private @const {!Logger} */
     this.logger_ = new Logger(this);
 
-    /** @private @const {!AnalyticsService} */
```
mborof: There's a big comment regarding the usage of "this" as a parameter.  It is currently:

    //NOTE: 'this' is passed in as a DepsDef.  Do not pass in 'this' before
    //analytics service and entitlements manager are constructed unless
    //you are certain they do not rely on them because they are part of that
    //definition.

It should minimally be before the first usage of this as a parameter.  

Not caused by you but you might want to consider how to make it more likely that a developer will actually read the comment too (or clean up the language) as the paradigm of passing "this" around as something that isn't really completely built yet seems quite delicate to me.

src/runtime/pay-client-test.js
```diff
@@ -378,17 +413,29 @@ describes.realWin('PayClientBindingSwg', {}, env => {
...
   let responseHandler;
+  let googleTransactionId;
 
   beforeEach(() => {
+    deps = new DepsDef();
     win = env.win;
     activityPorts = new ActivityPorts(win);
+    sandbox.stub(deps, 'win').callsFake(() => win);
+
+    googleTransactionId = 'GOOGLE_TRANSACTION_ID';
```
mborof: NIT: My opinion is that it would be cleaner to do this once at the top of the file, just after the includes:

const GOOGLE_TRANSACTION_ID = 'ANY_VALUE_YOU_WANT';

src/runtime/pay-flow.js
qidonna: Here the googleTransactionId is set in i param as a part of payment request, which will be passed to PaymentsAsyncClient.loadPaymentData. Before this step, PaymentsAsyncClient is initialized with another set of option, where I pass the googleTransactionId because the constructor in gpay_async.js checks if googleTransactionId is set in i param. If not, a new id will be generated by gpay. In summary the previous setup does not pass transaction id until loadPaymentData, but it was checked by gpay in the constructor. Hope this makes sense.

src/runtime/pay-client-test.js
qidonna: ah gotcha. I'll update it. Thanks!

src/runtime/pay-flow.js
mborof: I was curious about whether payClient.googleTransactionId_ would get set for whatever path is using PayClientBindingSwg.

qidonna: It is not being used in prod because flag GPAY_API is set to be 100% in prod so I did not debug that path.

src/runtime/pay-client-test.js
qidonna: I didn't set it in options because it needs to be passed back to pay-flow, so I set it as a member of PayClient.

src/runtime/pay-client.js
qidonna: I need this analyticsService here to generate a new ID.

build-system/tasks/closure-compile.js
```diff
@@ -18,6 +18,7 @@
 const fs = require('fs-extra');
 const argv = require('minimist')(process.argv.slice(2));
 const closureCompiler = require('gulp-closure-compiler');
+const colors = require('ansi-colors');
```
mborof: Is anything using this?

build-system/tasks/closure-compile.js
```diff
@@ -78,7 +79,7 @@ exports.closureCompile = function(
 };
 
 function compile(entryModuleFilenames, outputDir, outputFilename, options) {
-  return new Promise(function(resolve, reject) {
+  return new Promise(function(resolve, opt_reject) {
```
mborof: Why is this change here?

src/runtime/pay-flow.js
mborof: Sorry if this line of questioning is annoying, but is that true in AMP also?

src/runtime/pay-client.js
mborof: Where is it generating a new ID after the pay client gets instantiated?  I don't see that anywhere.  I wouldn't think it should ever do that.

Also, just considering possibilities but should PayClient take a deps object instead of all these individual parameters?

build-system/tasks/closure-compile.js
qidonna: colors.red in several places in this file.

src/runtime/pay-flow.js
qidonna: I think so. Also SwG in AMP does not use gpay_async.js for the payment.

src/runtime/pay-client.js
```diff
@@ -261,11 +267,13 @@ export class PayClientBindingPayjs {
 
   /**
    * @param {!Object} options
+   * @param {!string} googleTransactionId
```
dvoytenko: just `string`

src/runtime/pay-client.js
```diff
@@ -250,8 +254,10 @@ export class PayClientBindingPayjs {
         environment: '$payEnvironment$',
         'i': {
           'redirectKey': this.redirectVerifierHelper_.restoreKey(),
+          'googleTransactionId': googleTransactionId,
```
dvoytenko: Let's do one or the other to avoid confusion. Since you decided to update `PaymentAsyncClient.googleTransactionId`, let's go with that.

src/runtime/pay-flow.js
dvoytenko: Let's remove this.

src/runtime/runtime.js
qidonna: ACK. Thank you for pointing it out. I'm open to a better way.

src/runtime/pay-client.js
qidonna: Line 71 generates a new ID.

src/runtime/pay-flow.js
dvoytenko: Still here?

src/runtime/pay-client.js
```diff
@@ -65,11 +65,14 @@ export class PayClient {
    * @param {!Window} win
    * @param {!../components/activities.ActivityPorts} activityPorts
    * @param {!../components/dialog-manager.DialogManager} dialogManager
+   * @param {!../runtime/analytics-service.AnalyticsService} analyticsService
    */
-  constructor(win, activityPorts, dialogManager) {
+  constructor(win, activityPorts, dialogManager, analyticsService) {
+    this.googleTransactionId_ = analyticsService.getTransactionId();
```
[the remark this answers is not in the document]
qidonna: I'll move analyticsService.getTransactionId() to PayClientBindingPayjs new function call.

src/runtime/runtime.js
mborof: It doesn't look like you moved the comment above where you are instantiating analytics service.

src/runtime/pay-client.js
mborof: I think code may have shifted around.  Are you talking about: analyticsService.getTransactionId()?

mborof:
Looks like 2 of the discussion points had been marked as resolved and I forgot to unresolve them when I replied.  Sorry about that.  Nothing wrong with it as is (as far as I can tell) but I really want the comment moved at a minimum.  The other discussion could result in no changes, I mostly wanted to understand where/why/how the pay flow is generating a new transaction ID.

src/runtime/pay-client.js
qidonna: analyticsService.getTransactionId() was used in the original code in pay-flow.js to assign a transaction ID to i param in payment request. I moved it to pay-client.js to use it before PaymentsAsyncClient is initialized. Happy to talk offline.

qidonna: Comment has been addressed in the new commit.

src/runtime/runtime.js
qidonna: Comment has been addressed in the new PR.

src/runtime/pay-client-test.js
```diff
@@ -466,19 +444,8 @@ describes.realWin('PayClientBindingPayjs', {}, env => {
...
-
-    pageConfig = new PageConfig('pub1:label1');
-    sandbox.stub(deps, 'pageConfig').callsFake(() => pageConfig);
-
-    eventManager = new ClientEventManager(Promise.resolve);
-    sandbox.stub(deps, 'eventManager').callsFake(() => eventManager);
-
-    analyticsService = new AnalyticsService(deps);
+    googleTransactionId = GOOGLE_TRANSACTION_ID;
+    analyticsService = new AnalyticsService(runtime);
```
mborof: analyticsService = runtime.analytics()?

src/runtime/pay-client-test.js
```diff
@@ -415,28 +397,24 @@ describes.realWin('PayClientBindingSwg', {}, env => {
...
   beforeEach(() => {
-    deps = new DepsDef();
     win = env.win;
-    activityPorts = new ActivityPorts(win);
-    sandbox.stub(deps, 'win').callsFake(() => win);
+    pageConfig = new PageConfig('pub1:label1');
+    runtime = new ConfiguredRuntime(win, pageConfig);
 
-    googleTransactionId = GOOGLE_TRANSACTION_ID;
+    activityPorts = new ActivityPorts(win);
```
mborof: activityPorts = runtime.activities()?

src/runtime/runtime.js
```diff
@@ -518,24 +518,20 @@ export class ConfiguredRuntime {
...
-    this.payClient_ = new PayClient(
-      this.win_,
-      this.activityPorts_,
-      this.dialogManager_,
-      this.analyticsService_
-    );
+    this.payClient_ = new PayClient(this);
 
     /** @private @const {!Callbacks} */
     this.callbacks_ = new Callbacks();
```
mborof: Could you move this one so it is just below:

this.activityPorts_ = new ActivityPorts(this.win_);

The reason I prefer it there is that it is a part of DepsDef but doesn't take 'this' as a parameter.  So its slightly cleaner / less risky to declare it earlier.

src/runtime/pay-client-test.js
qidonna: done

src/runtime/pay-client-test.js
qidonna: done

src/runtime/pay-client-test.js
```diff
@@ -55,16 +58,24 @@ const INTEGR_DATA_OBJ_DECODED = {
...
   let payClient;
   let resultIdsAttached;
 
   beforeEach(() => {
     win = env.win;
+    pageConfig = new PageConfig('pub1:label1');
+    runtime = new ConfiguredRuntime(win, pageConfig);
+
     resultIdsAttached = [];
     activityPorts = new ActivityPorts(win);
```
mborof: activityPorts = runtime.activities()?

src/runtime/runtime.js
qidonna: done

src/runtime/pay-client-test.js
qidonna: done

PR #728, opened 2019-10-23
Subscription response

MattChilders92:


src/runtime/callbacks.js
```diff
@@ -144,50 +143,41 @@ export class Callbacks {
    * @param {function(!Promise<!../api/subscribe-response.SubscribeResponse>)} callback
    */
   setOnSubscribeResponse(callback) {
-    this.setCallback_(CallbackId.SUBSCRIBE_RESPONSE, callback);
+    console./*OK*/warn(`[swg.js:setOnSubscribeResponse]: This method has been deprecated, please swith usages to 'setOnPaymentResponse'`);
```
alin04: typo.  s/swith/switch

MattChilders92: Good catch, thanks

src/runtime/callbacks.js
```diff
@@ -144,50 +144,45 @@ export class Callbacks {
    * @param {function(!Promise<!../api/subscribe-response.SubscribeResponse>)} callback
    */
   setOnSubscribeResponse(callback) {
-    this.setCallback_(CallbackId.SUBSCRIBE_RESPONSE, callback);
+    console./*OK*/ warn(
```
jpettitt: add a warn method to log.js and call that rather than putting the console.warn here (ditto below)

MattChilders92: Done

PR #755, opened 2019-10-30
Change SKU Logging

mborof:
Change logging of SKU so that:

1) Context SKU is only set for SKUs the user has purchased (either already owns or just purchased).
2) Start logging event params SKU for offer selected, pay flow start, pay flow complete

Also, start logging pay flow cancel.


mborof:
Change logging of SKU so that:
  1) Context SKU is only set for SKUs the user has purchased (either already owns or just purchased).
  2) Start logging event params SKU for offer selected, pay flow start, pay flow complete, pay flow error, pay flow cancel (and start logging pay flow cancel).

PR #760, opened 2019-10-31
Log Account Linking

mborof:
Hey Dima.  These are the changes I discussed with you in the elevator.  If you could help certify the timing of each event I logged / make sure its appropriate.

I was in particular concerned about EVENT_LINKED_ACCOUNTS as I wasn't exactly sure the earliest moment I could log that.


mborof:
Also note that tests will still fail here because I haven't released the API with the new enum values yet and haven't updated the tests.  I'm just hoping to validate the basic code flow.

src/runtime/link-accounts-flow.js
```diff
@@ -186,6 +206,9 @@ export class LinkCompleteFlow {
    * @private
    */
   complete_(response) {
+    this.deps_
```
dvoytenko: Why DISMISS here?

src/runtime/link-accounts-flow.js
```diff
@@ -97,11 +102,17 @@ export class LinkCompleteFlow {
       );
       return promise.then(
         response => {
+          deps
+            .eventManager()
+            .logSwgEvent(AnalyticsEvent.ACTION_LINK_CONTINUE, true);
```
dvoytenko: Not sure CONTINUE is the right word here. But may be it's consistent with other names? But the flow doesn't really continue here - it just completes. As in "the action is done" by this time. There are just UX considerations at this point, I believe.

mborof: Ideally I would log this when the user clicks the continue button.  Is there a better location to do that?

I am open to a better naming convention.  The intention is that the user clicked the button to proceed to the next step (of account linking)

src/runtime/link-accounts-flow.js
mborof: The user clicked a button "dismissing" / "closing" the screen thanking them for updating google with their account information

mborof: I'm open to a better name if you have an idea

dvoytenko: DISMISS sounds like "cancel" to me. So maybe just CLOSE or COMPLETE?

src/runtime/link-accounts-flow.js
dvoytenko: That's the thing: user actually is not proceeding to the next step of account linking here IIUC. The user is all done - they are just acknowledging this fact. So maybe ACTION_LINK_ACK?

mborof: I definitely would like to log that the user clicked continue when they click continue.  Where can I log that?

dvoytenko: I think logging it is good. Just meaning of this event is "acknowledgement" of an action, not "continuation" of an action.

mborof: If the current meaning of the event is acknowledgement then it is being logged in the wrong place.  We need something to be logged when the user clicks the button to continue and initiate the actual linking flow.  Please clarify where I would log this event in code.

dvoytenko: Ok I might be misunderstanding this to some extent. Let's look at the actual flows.

mborof: Note that flow.start here will already log EVENT_GOOGLE_UPDATED and IMPRESSION_GOOGLE_UPDATED so we don't need a second thing logging that.  We just need something that logs that the user clicked the continue with linking button

mborof: We discussed and we will move this logging to the server in a separate pull

mborof:
The name of the dismiss event was changed to close.

We have no method of logging the user clicking next or canceling the workflow.  This will be fixed in a separate PR.

PR #767, opened 2019-11-04
Handle passed null promise

mborof:


src/runtime/wait-for-subscription-lookup-api-test.js
```diff
@@ -101,4 +101,19 @@ describes.realWin('WaitForSubscriptionLookupApi', {}, env => {
       }
     );
   });
+
+  it('should reject null account promise', async () => {
+    waitingApi = new WaitForSubscriptionLookupApi(runtime);
+    dialogManagerMock.expects('completeView').once();
+    return waitingApi.start().then(
```
ChrisAntaki: Could you simplify this using async/await syntax?

Ex:
```js
try {
  const foundAccount = await waitingApi.start();
  throw new Error('test failed. "' + foundAccount + '" should not be found');
} catch (reason) {
  expect(reason).to.equal('No account promise provided');
}
```

The `catch` portion might need to be tweaked a bit

PR #771, opened 2019-11-04
--DONOTMERGE--Activity Ports always pass analytics context to server

mborof:
To make it generically easier to engage in logging from within our iframes/popups/redirects so we can accurately log clicks to buttons within those iframes/popups/redirects.


src/components/activities.js
```diff
@@ -241,6 +241,33 @@ export class ActivityPorts {
...
+  }
+
+  /**
+   *
+   * @param {!../proto/api_messages.AnalyticsContext} context
+   */
+  setAnalyticsContext(context) {
+    this.analyticsContext_ = context;
+  }
+  /**
```
ChrisAntaki: Could you add a newline here?

src/components/activities.js
```diff
@@ -241,6 +241,33 @@ export class ActivityPorts {
...
+   * @param {?Object=} opt_args
+   * @return {!Object}
+   * @private
+   */
+  getArgs_(opt_args) {
+    opt_args = opt_args || {};
+    const context = this.analyticsContext_
+      ? this.analyticsContext_.toArray()
+      : null;
+    opt_args = Object.assign(opt_args, {
```
ChrisAntaki: Just a heads up, Object.assign mutates the first parameter, so there's no need to write `opt_args = ` here

ChrisAntaki: Perhaps this line could be combined with line 263?

ChrisAntaki: You could remove line 263, and then have this line be `opt_args = Object.assign({}, opt_args, {'analyticsContext': context})`

src/components/activities.js
```diff
@@ -241,6 +241,33 @@ export class ActivityPorts {
...
+  setAnalyticsContext(context) {
+    this.analyticsContext_ = context;
+  }
+  /**
+   * Adds universal arguments
+   * @param {?Object=} opt_args
+   * @return {!Object}
+   * @private
+   */
+  getArgs_(opt_args) {
```
ChrisAntaki: Could we call this method `getArgsWithDefaults`, just to make it's purpose more obvious?

src/runtime/analytics-service.js
```diff
@@ -184,10 +184,10 @@ export class AnalyticsService {
     if (!this.serviceReady_) {
       // TODO(sohanirao): Potentially do this even earlier
       this.doc_.getBody().appendChild(this.getElement());
+      this.setContext_();
```
ChrisAntaki: Just curious, why did this move up?

src/runtime/offers-flow-test.js
```diff
@@ -50,14 +50,12 @@ describes.realWin('OffersFlow', {}, env => {
...
   let messageCallback;
   let messageMap;
-  let analyticsContext;
 
   beforeEach(() => {
     win = env.win;
     messageMap = {};
     pageConfig = new PageConfig('pub1:label1');
     runtime = new ConfiguredRuntime(win, pageConfig);
-    analyticsContext = runtime.analytics().getContext();
```
ChrisAntaki: Great to see the tests getting slimmer!

src/components/activities.js
ChrisAntaki: On second thought, maybe something like `getArgsWithInvariants` would be better, since the `analyticsContext` arg isn't just a default value here, it would be overridden if it were passed in the `opt_args` object

src/components/activities.js
```diff
@@ -251,6 +278,7 @@ export class ActivityPorts {
    * @return {!Promise<!ActivityIframePort>}
    */
   openIframe(iframe, url, opt_args) {
+    opt_args = this.getArgs_(opt_args);
```
elijahsoria: I think that we need to do a call to AnalyticsService's getContext() before this line because it looks like AnalyticsContext can change throughout the user's journey. At least it seems that way given AnalyticsService's addLabels function.

src/components/activities.js
mborof: Now that you mention it, we should probably respect the value passed in if one is.

src/runtime/analytics-service.js
mborof: So the data put in context was there sooner (in case the activities are going to use it)

src/components/activities.js
mborof: AnalyticsContext is passed by reference so it should always reflect the current object in AnalyticsServer.

mborof:
This was a reasonable prototype but we need to roll this out in phases so this code should NOT BE MERGED.

ChrisAntaki:
👍

PR #798, opened 2019-11-16
Use oldsku

Kestrad:


Kestrad:
> Add a test?

Added a test to pay-flow-test to check that completion flow works with oldSku around in the response.

src/runtime/pay-flow-test.js
```diff
@@ -426,6 +427,49 @@ describes.realWin('PayCompleteFlow', {}, env => {
...
+      'email': 'test@example.org',
+    });
+    const response = new SubscribeResponse(
+      'RaW',
+      purchaseData,
+      userData,
+      null,
+      ProductType.SUBSCRIPTION,
+      null,
+      'oldSku'
```
alin04: use a different string to represent the old sku. It's confusing to read this and differentiate from the name of the object key.

Kestrad: Done!

PR #819, opened 2019-11-21
Refactor ButtonApi

mborof:
There will be some work done to add additional options to ButtonApi.  This refactor shouldn't modify any behavior but will make the work a bit easier by centralizing the button click events into 1 location.


src/runtime/button-api.js
```diff
@@ -186,32 +199,18 @@ export class ButtonApi {
...
-            /* isFromUserAction */ true
-          )
-      )
-    );
-
     return new SmartSubscriptionButtonApi(
       deps,
       button,
       options,
-      castedCallback
```
elijahsoria: Why do we send an empty function here?

mborof: Because I am manually calling the passed callback here:

  setupButtonAndGetOptions_(button, optionsOrCallback, callbackFun) {
    const options = this.getOptions_(optionsOrCallback);
    const callback = this.getCallback_(optionsOrCallback, callbackFun);
    button.addEventListener('click', event => {
      this.logSwgEvent_(AnalyticsEvent.ACTION_SWG_BUTTON_CLICK, true);
      if (typeof callback === 'function') {
        callback(event);
      }
    });
    return options;
  }

So SmartSubscriptionButtonApi shouldn't have to.  It does make sense to ask this question though: is it possible for someone to click the button in the iframe and not trigger the click event in the code I showed above?

src/runtime/button-api.js
```diff
@@ -178,6 +185,25 @@ export class ButtonApi {
...
+    const options = this.getOptions_(optionsOrCallback);
+    const callback = this.getCallback_(optionsOrCallback, callbackFun);
+    const clickFun = event => {
+      this.logSwgEvent_(AnalyticsEvent.ACTION_SWG_BUTTON_CLICK, true);
+      if (typeof callback === 'function') {
+        callback(event);
+      }
+    };
+    button.addEventListener('click', clickFun);
+    return {options, clickFun};
```
dparikh: why do you need to return `clickFn`?

PR #835, opened 2019-11-26
Return Promise.reject when decryptedDocumentKey is not found.

elijahsoria:
Returns a rejected Promise when getEntitlements is called with a non-null encryptedDocKey and a decryptedDocKey is not found in the response.

### DIFF
--- a/src/runtime/entitlements-manager.js
+++ b/src/runtime/entitlements-manager.js
@@ -120,9 +120,9 @@ export class EntitlementsManager {
         encryptedDocumentKey != null &&
         response.decryptedDocumentKey == null
       ) {
-        // Fail to avoid cases of malicious keys.
+        // Fail to avoid cases where we expose undecryptable ciphertext.
         return Promise.reject(
-          new Error('Missing encrypted doc key in entitlements response.')
+          new Error('Missing decrypted doc key in entitlements response.')
         );
       }
       if (response.isReadyToPay != null) {
--- a/src/runtime/entitlements-manager.js
+++ b/src/runtime/entitlements-manager.js
@@ -116,6 +116,10 @@ export class EntitlementsManager {
       this.responsePromise_ = this.getEntitlementsFlow_(encryptedDocumentKey);
     }
     return this.responsePromise_.then(response => {
+      if (encryptedDocumentKey != null && response.decryptedDocumentKey == null) {
+        // Fail to avoid cases of malicious keys.
+        return Promise.reject(new Error('Missing encrypted doc key in entitlements response.'));
+      }
       if (response.isReadyToPay != null) {
         this.analyticsService_.setReadyToPay(response.isReadyToPay);
       }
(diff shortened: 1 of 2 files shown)


src/runtime/entitlements-manager.js
```diff
@@ -116,6 +116,15 @@ export class EntitlementsManager {
       this.responsePromise_ = this.getEntitlementsFlow_(encryptedDocumentKey);
     }
     return this.responsePromise_.then(response => {
+      if (
+        encryptedDocumentKey != null &&
+        response.decryptedDocumentKey == null
+      ) {
+        // Fail to avoid cases of malicious keys.
+        return Promise.reject(
```
ChrisAntaki: Would it be friendlier to 3rd party developers (and users if a 3rd party's server mistakenly didn't return a decrypted key) if we logged a warning or error instead?

src/runtime/entitlements-manager.js
```diff
@@ -116,6 +116,15 @@ export class EntitlementsManager {
       this.responsePromise_ = this.getEntitlementsFlow_(encryptedDocumentKey);
     }
     return this.responsePromise_.then(response => {
+      if (
```
dvoytenko: Do you want to do this here or in AMP?

src/runtime/entitlements-manager.js
```diff
@@ -116,6 +116,15 @@ export class EntitlementsManager {
...
+        encryptedDocumentKey != null &&
+        response.decryptedDocumentKey == null
```
dvoytenko: We also need to check that entitlement is a granting one. It's totally ok to return a `null` decrypted key if the user doesn't have the required access.

src/runtime/entitlements-manager.js
elijahsoria: I think it would be better to fail since we don't want to run an encrypted script on a SwG site since it isn't validated.

src/runtime/entitlements-manager.js
elijahsoria: I think here would be easier since we make multiple getEntitlements calls in AMP, whereas all getEntitlements calls go through this method in SwG.

dvoytenko: But we still want to do something very similar in AMP as well for all other platforms, right? Otherwise, from the pov of SwG we here essentially testing ourselves. I'm pretty confident once we launch if we ever get a granting entitlement w/o decryption key it'd be purely due to a bug on our side and not a user error.

src/runtime/entitlements-manager.js
```diff
@@ -116,6 +116,15 @@ export class EntitlementsManager {
       this.responsePromise_ = this.getEntitlementsFlow_(encryptedDocumentKey);
     }
     return this.responsePromise_.then(response => {
+      if (
+        encryptedDocumentKey != null &&
```
dvoytenko: This would probably be better as:

```
if (granted && encryptedDocumentKey && !response.decryptedDocumentKey) {
  ...
}
```

src/runtime/entitlements-manager.js
elijahsoria: Truth. I will update AMP too. Thanks for the heads up.

elijahsoria:
Closing since this should only be done in AMP (https://github.com/ampproject/amphtml/pull/25829). Publishers can decide on this fail case during their own integration.

PR #836, opened 2019-11-26
One time contributions

Kestrad:
Updated skuSelectedResponse to include information about whether offer is one-time or not.


src/runtime/pay-flow.js
```diff
@@ -126,7 +126,15 @@ export class PayStartFlow {
     }
 
     // Start/cancel events.
-    this.deps_.callbacks().triggerFlowStarted(SubscriptionFlows.SUBSCRIBE, req);
+    if (this.productType_ == ProductType.SUBSCRIPTION) {
```
alin04: Revert this change and keep it as a separate PR (to fix the trigger).

Kestrad: Huh, I thought I'd reverted this change on this branch. Thanks for catching that!

Kestrad:
Looks like the thing in https://github.com/subscriptions-project/swg-js/pull/666 happened where I missed a bracket while pasting in the tests and then ran Prettier - the tests' indentations are all wrong. Going to try recompiling the tests and pasting them again. I use ctrl+a to copy the files over, so it's spooky that I've managed to do this twice now while modifying protos.

Kestrad:
Closing to see if reopening will fix codecov blocking merge.

PR #853, opened 2019-12-04
Make it easier to get an SKU from entitlements

mborof:
1) Add public method to get SKU from an entitlement.
2) Add public method to pick a single SKU from all entitlements.
3) Anytime someone gets entitlements, update analytics service with that information.


src/api/entitlements.js
```diff
@@ -263,4 +288,16 @@ export class Entitlement {
       : [json];
     return jsonList.map(json => Entitlement.parseFromJson(json));
   }
+
+  /**
+   * Returns the SKU associated with this entitlement.
+   * @return {?string}
+   */
+  getSku() {
```
alin04: A very similar method already exists. Please DRY.

https://github.com/subscriptions-project/swg-js/blob/master/src/runtime/pay-flow.js#L484

src/runtime/runtime.js
```diff
@@ -717,7 +717,17 @@ export class ConfiguredRuntime {
...
-      .then(entitlements => entitlements.clone());
+      .then(entitlements => {
+        // Auto update internal things tracking the user's current SKU.
+        let sku = null;
+        if (entitlements) {
+          try {
+            sku = entitlements.getSingleEntitlement().getSku();
+          } catch (ex) {}
+        }
+        this.analyticsService_.setSku(sku);
```
alin04: Is logging this info useful? If so, may as well log all the skus?

src/api/entitlements.js
```diff
@@ -175,6 +175,31 @@ export class Entitlements {
...
+   * TODO: come up with a better way of picking the right SKU when there are
+   * multiple
+   * @return {?string}
+   */
+  getSingleEntitlement() {
+    if (!this.entitlements) {
+      return null;
+    }
+    for (let i = 0; i < this.entitlements.length; i++) {
+      const ent = this.entitlements[i];
```
henel677: please use the full name, i.e. entitlement

src/api/entitlements.js
```diff
@@ -175,6 +175,31 @@ export class Entitlements {
...
+      const sku = ent.getSku();
+      if (!sku) {
+        continue;
+      }
+      if (this.product_ && !ent.enables(this.product_)) {
+        continue;
+      }
+      return sku;
+    }
+    return null;
```
henel677: Do we want to throw an exception instead?

src/api/entitlements.js
```diff
@@ -175,6 +175,31 @@ export class Entitlements {
...
   }
+
+  /**
+   * Converts the whole set of entitlements into a single SKU.
+   * Returns the first SKU that enables this page.
+   * TODO: come up with a better way of picking the right SKU when there are
+   * multiple
+   * @return {?string}
+   */
+  getSingleEntitlement() {
```
henel677: -> getSkuEntitlement  (or something like that)?

Would it make sense for this function to return all the SKUs and let the caller select which of the SKUs to use (then you wouldn't need your TODO)?

src/runtime/runtime.js
```diff
@@ -717,7 +717,17 @@ export class ConfiguredRuntime {
   getEntitlements(encryptedDocumentKey) {
     return this.entitlementsManager_
       .getEntitlements(encryptedDocumentKey)
-      .then(entitlements => entitlements.clone());
+      .then(entitlements => {
+        // Auto update internal things tracking the user's current SKU.
+        let sku = null;
+        if (entitlements) {
+          try {
+            sku = entitlements.getSingleEntitlement().getSku();
```
henel677: Doesn't getSingleEntitlement() return the sku? I.e. should be "sku = entitlements.getSingleEntitlement();"?

src/api/entitlements.js
mborof: The code is effectively the same but it is accepting different kinds of inputs.  One is a string associated with the purchase and the other is one associated with an entitlement.  Is there anything insisting these 2 things match or is it just happenstance?

I could create a generic utility function to read a single property out of a string that might be a json but it seems fairly unnecessary.

src/runtime/runtime.js
mborof: Currently the logs do not know if you are already subscribed or not.  It is useful to know if you were already subscribed (give providing the SKUs you are subscribed to).

I can log all of them though.  That makes sense.

src/api/entitlements.js
mborof: The whole point is to centralize this bad logic into a single place so if multiple things require it we only need to fix the 1 function instead of all the things that might rely on this bad logic.  I have a specific use case where I need to pick a single SKU and I'd rather the calling code not need to have that logic in it.

src/api/entitlements.js
mborof: You could certainly not have any entitlements so an exception would not be appropriate.

src/runtime/runtime.js
mborof: I've updated it to return the full entitlement instead

mborof: I've managed to avoid needing this funciton.

src/api/entitlements.js
mborof: I've managed to avoid needing this function.

src/api/entitlements.js
mborof: I went ahead and centralized this logic

src/utils/json.js
```diff
@@ -120,3 +120,15 @@ function hasOwnProperty(obj, key) {
     key
   );
 }
+
+/**
+ * Converts the passed string into a JSON object (if possible) and returns the
+ * value of the propertyName on that object.
+ * @param {string} jsonString
+ * @param {string} propertyName
+ * @return {*}
```
henel677: How about we return String or null?

@return {?string}

src/api/entitlements.js
```diff
@@ -263,4 +265,17 @@ export class Entitlement {
...
     return jsonList.map(json => Entitlement.parseFromJson(json));
   }
+
+  /**
+   * Returns the SKU associated with this entitlement.
+   * @return {?string}
+   */
+  getSku() {
+    return (
+      /** @type {?string} */ (getPropertyFromJsonString(
```
henel677: If getPropertyFromJsonString returns {?string} you won't need this cast

src/runtime/runtime.js
```diff
@@ -717,7 +717,21 @@ export class ConfiguredRuntime {
   getEntitlements(encryptedDocumentKey) {
     return this.entitlementsManager_
       .getEntitlements(encryptedDocumentKey)
-      .then(entitlements => entitlements.clone());
+      .then(entitlements => {
+        // Auto update internal things tracking the user's current SKU.
+        if (entitlements) {
+          try {
+            let skus = '';
```
henel677: How about:
const skus = entitlements.entitlements.map(entitlement => entitlement.getSku());
if (skus.length > 0) {
  this.analyticsService_.setSku(skus.join(','));
}

alin04: +1

src/utils/json.js
mborof: Because that would be a lie.  An object can contain anything.  For instance:
fun(" {val: true} ", "val") 

would return true which is not a string.

src/api/entitlements.js
mborof: In this case it returns a string but that is not generically guaranteed for any random object passed in.

src/runtime/runtime.js
mborof: Done

PR #855, opened 2019-12-05
Further work to propogate one-time status with sku

Kestrad:
For one-time contributions, but should also be compatible with one-time subscriptions in the future if those become relevant. Test currently doesn't pass yet.


src/runtime/pay-flow.js
```diff
@@ -125,6 +137,10 @@ export class PayStartFlow {
       swgPaymentRequest.replaceSkuProrationMode = this.prorationEnum;
     }
 
+    if (this.recurrenceEnum) {
+      swgPaymentRequest.oneTime = this.recurrenceEnum;
```
ChrisAntaki: Could you set this after line 121, when the `recurrenceEnum` prop is set?

Kestrad: I was going to say that I did this just to match the pattern we used to set proration mode - but looking at the blame layer, I was also the one who did that. Lemme check that PR to make sure there wasn't some reason I did that that I'm now forgetting.

Kestrad: I'm doing the suggested change but it broke some tests. Trying to figure out why.

Kestrad: Turns out the tests and flows are written in such a way that the easiest way to not break the test is to separate things out the way I had them. I'll look into how to make the tests better tomorrow morning.

src/runtime/contributions-flow.js
```diff
@@ -82,8 +82,20 @@ export class ContributionsFlow {
...
   startPayFlow_(response) {
     const sku = response.getSku();
+    const isOneTime = response.getOneTime();
     if (sku) {
-      new PayStartFlow(this.deps_, sku, ProductType.UI_CONTRIBUTION).start();
+      if (isOneTime) {
+        const contributionRequest = {};
+        contributionRequest.skuId = sku;
+        contributionRequest.oneTime = isOneTime;
+        new PayStartFlow(
```
dparikh: Could simplify as:

const contributionRequest = {
  skuId: sku,
  oneTiem: isOnetime,
}

src/runtime/contributions-flow.js
```diff
@@ -82,8 +82,20 @@ export class ContributionsFlow {
    */
   startPayFlow_(response) {
     const sku = response.getSku();
+    const isOneTime = response.getOneTime();
     if (sku) {
-      new PayStartFlow(this.deps_, sku, ProductType.UI_CONTRIBUTION).start();
+      if (isOneTime) {
+        const contributionRequest = {};
+        contributionRequest.skuId = sku;
+        contributionRequest.oneTime = isOneTime;
```
dparikh: Could simplify as:

const contributionRequest = {
  skuId: sku,
  oneTiem: isOnetime,
}

Kestrad: Fixed, thank you!

src/runtime/pay-flow.js
Kestrad: Going to figure out if the tests can be improved some other time, in order to submit this feature sooner. (In theory I can fix the breaking ones easily enough by just changing the expected value, but doing so feels like it goes against the spirit of the tests.)

PR #862, opened 2019-12-07
♻️ Refactors runtime.js a bit for readability + fixes a few tests

ChrisAntaki:
- Tries to improve readability by adding comments, renaming vars, etc.
- Fixes a flaky E2E test
- Fixes a few unit tests that weren't having their expectations checked, which I discovered by checking the coverage for the test itself (https://codecov.io/gh/subscriptions-project/swg-js/pull/862/src/src/runtime/runtime-test.js)


src/runtime/runtime.js
```diff
@@ -84,43 +87,46 @@ export function getRuntime() {
...
-    configurable: false,
-  });
-  win[RUNTIME_PROP] = dependencyInstaller;
-  win[RUNTIME_LEGACY_PROP] = dependencyInstaller;
-  if (waitingArray) {
-    waitingArray.forEach(pushDependency);
-  }
+
+  // Queue up any callbacks the publication might have provided.
+  [].concat(win[RUNTIME_PROP], win[RUNTIME_LEGACY_PROP]).forEach(push);
```
jpettitt: Nit: It would be marginally more readable if you left the `const waitingArray =`

src/runtime/runtime.js
```diff
@@ -84,43 +87,46 @@ export function getRuntime() {
...
   const publicRuntime = createPublicRuntime(runtime);
 
-  const dependencyInstaller = {};
-
   /**
+   * Executes a callback when SwG runtime is ready.
    * @param {function(!Subscriptions)} callback
    */
-  function pushDependency(callback) {
+  function push(callback) {
```
jpettitt: Nit: naming is confusing since there is an `Array.push` already.  More so since it's not actually pushing anything to an array.

ChrisAntaki: How about the name `callWhenRuntimeIsReady`?

src/runtime/runtime.js
ChrisAntaki: How about the var name `waitingCallbacks`?

ChrisAntaki: Would the var name `const waitingCallbacks` work well?

jpettitt: sure

src/runtime/runtime.js
jpettitt: yes

PR #877, opened 2019-12-12
Preserve paymentRequest locally.

alin04:
Preserve paymentRequest for native buy flow and non-native popup. Won't work for redirect.

Clean up SubscribeResponse so it only contains data for publishers' callback.


alin04:
Completely messed up this PR. Started over with #878

PR #878, opened 2019-12-12
Preserve paymentRequest locally.

alin04:
Preserve paymentRequest for native buy flow and non-native popup. Won't work for redirect.

Clean up SubscribeResponse so it only contains data for publishers' callback.


alin04:
Sending this out for some early feedback.

src/runtime/pay-flow.js
```diff
@@ -187,22 +187,30 @@ export class PayCompleteFlow {
...
         payPromise,
         flow.complete.bind(flow)
       );
-      deps.callbacks().triggerPaymentResponse(promise);
-      return promise.then(
-        response => {
-          const sku = parseSkuFromPurchaseDataSafe(response.purchaseData);
+
+      deps.callbacks().triggerPaymentResponse(subscribeResponsePromise);
+      return Promise.all([subscribeResponsePromise, payPromise]).then(
```
mborof: This is a notable change here, you're waiting for another promise.  Is that necessary or was not waiting causing any issues previously?

alin04: I'm not an expert on promises, but since validatePayResponse above also waits on the same payPromise, I didn't think this was a significant change.

I need payPromise because it has the paymentRequest in it. I pared down SubscribeReponse so it only has data that publishers may need. Previously we were "cheating" by using SubscribeReponse to shephard extra data we needed.

mborof:
I know you say this won't work for the redirect flow (and clearly it won't since it all relies on local memory) but there are a significant number of payment events that go through redirect so its not really a case we can just ignore.  Are there any plan for this for the redirect workflow?

It otherwise LGTM, it just won't work for 20-25% of our transactions...

alin04:
Yes, there are plans to fix redirect flow but that is much more involved and requires quite a bit more changes through Play/Pay stacks. 

This is a short term solution for fixing known issues with native buy flow.

alin04:
@Kestrad , FYI, I'm changing how we remember the request.

alin04:
Need a new approval please.

PR #891, opened 2019-12-16
Delay Redirect so Logging Can Finish

mborof:
This fixes an issue where we tell the client to log something and then immediately tell the client to redirect.  This causes the page to redirect before the logging occurs (causing broken logs).

Ultimately we really need a "logging promise" that we can wait to resolve prior to redirecting the page but it will take some time to implement that.  As a temporary solution I am proposing a brief delay before a page redirect.

### DIFF
--- a/src/runtime/pay-client.js
+++ b/src/runtime/pay-client.js
@@ -174,6 +174,22 @@ class PayClientBindingSwg {
 
   /** @override */
   start(paymentRequest, options) {
+    if (options.forceRedirect) {
+      // This resolves an issue with logging where the page redirects before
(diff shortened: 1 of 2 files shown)


src/runtime/pay-client.js
```diff
@@ -174,6 +174,22 @@ class PayClientBindingSwg {
 
   /** @override */
   start(paymentRequest, options) {
+    if (options.forceRedirect) {
```
alin04: PayClientBindingSwg is the obsolete lib and only used by AMP.

mborof: Ok, I've updated both objects now.  Thank you!

PR #896, opened 2019-12-17
Create SwG Logging promise

mborof:
Experimental code so we have a method of waiting to redirect until all logging is finished.

### DIFF
--- a/src/runtime/analytics-service-test.js
+++ b/src/runtime/analytics-service-test.js
@@ -19,6 +19,7 @@ import {
   AnalyticsEvent,
   EventOriginator,
   EventParams,
+  FinishedLoggingResponse,
 } from '../proto/api_messages';
 import {AnalyticsService} from './analytics-service';
 import {ClientEventManager} from './client-event-manager';
@@ -36,7 +37,7 @@ describes.realWin('AnalyticsService', {}, env => {
   let analyticsService;
   let pageConfig;
   let runtime;
-  let registeredCallback;
+  let eventManagerCallback;
 
   const productId = 'pub1:label1';
   const defEventType = AnalyticsEvent.IMPRESSION_PAYWALL;
@@ -50,7 +51,7 @@ describes.realWin('AnalyticsService', {}, env => {
   beforeEach(() => {
     sandbox
       .stub(ClientEventManager.prototype, 'registerEventListener')
-      .callsFake(callback => (registeredCallback = callback));
+      .callsFake(callback => (eventManagerCallback = callback));
     win = env.win;
     src = '/serviceiframe';
     pageConfig = new PageConfig(productId);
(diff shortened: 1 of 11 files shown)


src/runtime/analytics-service.js
```diff
@@ -89,6 +91,12 @@ export class AnalyticsService {
     this.eventManager_.registerEventListener(
       this.handleClientEvent_.bind(this)
     );
+
+    /** @private {!number} */
+    this.sent_ = 0;
```
ChrisAntaki: Could you describe this more with a comment and/or variable rename?

mborof: Done

src/runtime/analytics-service.js
```diff
@@ -37,6 +37,14 @@ const iframeStyles = {
   display: 'none',
 };
 
+// We will wait up to 750 ms for the initial log event (iframe must load)
+const MAX_WAIT_BEFORE_LOAD = 750;
```
ChrisAntaki: @jpettitt What do you think about this max? I'd prefer 100 personally... At a certain point, it's like do we want to lose a log event or a conversion? Just my 2c

mborof: Measuring the time it takes for logging from my computer indicates the FIRST log takes 500-600ms.  This is because an iframe has to load.  This case would only affect a user if the publisher linked to a page that immediately initiated a buy flow.

Once the iframe has loaded it takes 1-60ms to log in my measuring.  This would be the expected wait for most users (who are being redirected).

mborof:
Just a note that the inner frame code has propagated now so I am getting the "logging done" signal.

src/runtime/analytics-service.js
ChrisAntaki: > This case would only affect a user if the publisher linked to a page that immediately initiated a buy flow.

In that case, why not just limit the wait time to 100ms and lose the event, to increase the chances the publisher gets a subscription?

src/runtime/analytics-service-test.js
```diff
@@ -116,16 +145,22 @@ describes.realWin('AnalyticsService', {}, env => {
...
         eventOriginator: EventOriginator.UNKNOWN_CLIENT,
         isFromUserAction: null,
         additionalParameters: null,
       });
 
+      // These wait for analytics server to be ready to send data.
       await analyticsService.lastAction_;
       await activityIframePort.whenReady();
+
+      // These enusre the right event was communicated.
```
ChrisAntaki: enusre -> ensure

src/runtime/analytics-service.js
```diff
@@ -35,6 +37,23 @@ const iframeStyles = {
...
+// to the iframe code that's already loaded.
+const MAX_WAIT = 100;
+const TIMEOUT_ERROR = 'AnalyticsService timed out waiting for a response';
+
+/**
+ *
+ * @param {!string} error
+ * @param {boolean=} success
+ */
+function createResponse(error, success = false) {
```
ChrisAntaki: Could we call this "createErrorResponse"?

PR #821, opened 2019-11-21
Minor change to specify correct type of payment

Kestrad:

PR #898, opened 2019-12-18
Trigger flow fix

Kestrad:
Redo of PR 821 (https://github.com/subscriptions-project/swg-js/pull/821)

Test currently doesn't pass even though console logs show a different behavior than what the test seems to be seeing - trying to figure out why.

### DIFF
--- a/src/runtime/pay-flow-test.js
+++ b/src/runtime/pay-flow-test.js
@@ -169,7 +169,11 @@ describes.realWin('PayStartFlow', {}, env => {
       skuId: 'sku1',
       publicationId: 'pub1',
     };
-    const flow = new PayStartFlow(runtime, subscriptionRequest, ProductType.UI_CONTRIBUTION);
+    const flow = new PayStartFlow(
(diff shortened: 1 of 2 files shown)


src/runtime/pay-flow-test.js
```diff
@@ -164,6 +164,47 @@ describes.realWin('PayStartFlow', {}, env => {
     await expect(flowPromise).to.eventually.be.undefined;
   });
 
+  it('should trigger the contribution flow if given contribution productType', async () => {
```
alin04: This test won't pass because the response callback isn't getting triggered (which adds the original request data to the paymentResponse). See line 1005 of the original file for an example (the test called 'should start flow with data from request').

Kestrad: Turns out it actually wasn't passing because I was passing in the product type incorrectly.

alin04: Ahh, good to know.

PR #903, opened 2019-12-19
Fixing PayCompleteFlow's flow canceled reporting

Kestrad:
triggerFlowCanceled should now receive the correct product type in some cases when the user closes out of the payment iframe. It will only do so when the binding is Pay and not SwG. Still need to make it report accurately for SwG and also need tests.

Continuation of PR #898 (https://github.com/subscriptions-project/swg-js/pull/898)


src/runtime/pay-client.js
```diff
@@ -222,7 +223,15 @@ export class PayClient {
       )
       .catch(reason => {
         if (typeof reason == 'object' && reason['statusCode'] == 'CANCELED') {
-          return Promise.reject(createCancelError(this.win_));
+          const error = createCancelError(this.win_);
+          if (this.request_) {
```
alin04: Use the request param instead of this.request_.

build-system/tasks/karma.conf.js
```diff
@@ -121,9 +121,9 @@ module.exports = {
   },
 
   singleRun: true,
-  browserDisconnectTimeout: 10000,
+  browserDisconnectTimeout: 1000000,
```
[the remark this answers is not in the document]
Kestrad: Oops, did not mean to commit this file. This change was just for debugging with Karma, so that the browser wouldn't refresh immediately while I was trying to debug a test.

src/runtime/pay-client.js
```diff
@@ -222,7 +223,15 @@ export class PayClient {
...
+            error['productType'] = /** @type {!Object} */ (this.request_)['i'][
```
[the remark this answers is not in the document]
Kestrad: Fixed, thanks!

src/runtime/pay-client.js
Kestrad: Also fixed!

build-system/tasks/karma.conf.js
```diff
@@ -123,7 +123,7 @@ module.exports = {
   singleRun: true,
   browserDisconnectTimeout: 10000,
   browserDisconnectTolerance: 2,
-  browserNoActivityTimeout: 4 * 60 * 1000,
+  browserNoActivityTimeout: 400 * 60 * 1000,
```
alin04: Is this desired change too?

src/runtime/pay-client.js
```diff
@@ -222,7 +223,14 @@ export class PayClient {
       )
       .catch(reason => {
         if (typeof reason == 'object' && reason['statusCode'] == 'CANCELED') {
-          return Promise.reject(createCancelError(this.win_));
+          const error = createCancelError(this.win_);
+          if (request) {
+            error['productType'] =
+              /** @type {!PaymentDataRequest} */ (request)['i']['productType'];
+          } else {
+            error['productType'] = ProductType.SUBSCRIPTION;
```
alin04: Would prefer not to default to anything.

build-system/tasks/karma.conf.js
Kestrad: Also a leftover from testing. Sorry, I really need to check more thoroughly before submitting PRs in the future.

src/runtime/pay-client.js
Kestrad: Not defaulting to anything makes the test a bit trickier to write.. Would it be okay for me to set it to null? Either way, pay-flow.js ends up defaulting it to the subscription flow if it's not the contribution flow.

Kestrad: I've set it to be null by default if no product type is given.

PR #904, opened 2019-12-19
Stop throwing errors on window close

mborof:
When the user closes the gPay window:
* Stop throwing the abort error
* Stop triggering payment response with null response data (it should just trigger cancel workflow)

### DIFF
--- a/src/runtime/callbacks.js
+++ b/src/runtime/callbacks.js
@@ -177,9 +177,14 @@ export class Callbacks {
   triggerPaymentResponse(responsePromise) {
     this.paymentResponsePromise_ = responsePromise.then(
       res => {
-        this.trigger_(CallbackId.PAYMENT_RESPONSE, res);
+        debugger;
+        this.trigger_(
+          CallbackId.PAYMENT_RESPONSE,
+          Promise.resolve(res.clone())
+        );
       },
       reason => {
+        debugger;
         if (isCancelError(reason)) {
           return;
         }
@@ -259,6 +264,7 @@ export class Callbacks {
    * @private
    */
   trigger_(id, data) {
+    debugger;
     this.resultBuffer_[id] = data;
     const callback = this.callbacks_[id];
     if (callback) {
--- a/src/runtime/callbacks.js
+++ b/src/runtime/callbacks.js
@@ -177,14 +177,12 @@ export class Callbacks {
   triggerPaymentResponse(responsePromise) {
     this.paymentResponsePromise_ = responsePromise.then(
       res => {
-        debugger;
         this.trigger_(
           CallbackId.PAYMENT_RESPONSE,
           Promise.resolve(res.clone())
         );
       },
       reason => {
-        debugger;
         if (isCancelError(reason)) {
           return;
         }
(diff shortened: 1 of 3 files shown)


src/runtime/callbacks.js
```diff
@@ -172,10 +175,21 @@ export class Callbacks {
...
-      CallbackId.PAYMENT_RESPONSE,
-      responsePromise.then(res => res.clone())
+    this.paymentResponsePromise_ = responsePromise.then(
+      res => {
+        this.trigger_(
+          CallbackId.PAYMENT_RESPONSE,
+          Promise.resolve(res.clone())
+        );
+      },
+      reason => {
```
alin04: This prevents the publisher from capturing the error too and dealing with it, right?

src/runtime/callbacks.js
```diff
@@ -172,10 +175,21 @@ export class Callbacks {
...
+        if (isCancelError(reason)) {
+          return;
+        }
+        throw reason;
```
alin04: I'm not sure on the sequence of events, but if this throws, won't pay-flow.js execution halt at pay-flow.js:201 resulting in no logging of any failure?

src/runtime/callbacks.js
mborof: They shouldn't have to catch a cancel error, they should be informed through our trigger flow cancel workflow.

If you mean the non-cancel errors, I would think this duplicates existing behavior.  If nothing was catching the error before it would effectively just be thrown to the next level?  I could be wrong though, my grasp of promises is not 100%.  I basically did this development because there were errors thrown when I closed the gPay popup which seems inappropriate.

src/runtime/callbacks.js
mborof: It's an asynchronous error because it happens in the error part of a promise.  It was throwing these errors before also, I am just suppressing it if you close the popup.

src/runtime/callbacks.js
alin04: Hmm, I'm not 100% on promises either. Ok, if you're confident..

Kestrad: If I'm reading the code correctly, though, doesn't this throw out cancel errors as well?

mborof: If the user cancels (e.g. closes the popup window), the this.trigger code is never called.  It instead calls the part that starts with reason.  This is consuming an error thrown by the underlying promise.

In that catch, if it is the cancel error it just returns (basically canceling the error).  Otherwise, it throws the error so something higher level and be aware of it (ultimately it just goes to console).

mborof: I must admit when I read your comment I wondered if you were correct.  It can be difficult to parse through how promises actually function...  I think both things are catching the same error because they operate on the same promise.  E.g.:

x = new Promise();
x.then(successFun, failFun);
x.then(successFun, failFun);
failFun will get called twice even if it doesn't throw the error.

To be certain, I just validated by testing this change locally.  I did so by adding this to DemoPaywalController in the swg-js/examples/sample-pub/demo.js

  this.subscriptions.setOnFlowCanceled(a => {
    debugger;
  });

And validated closing the gPay window hit that debugger point.

mborof: Thanks for checking my dev though! I always appreciate multiple perspectives.

Kestrad: Ah, okay. Thank you for checking!

PR #908, opened 2019-12-19
Simplify internal pay-flow interfaces.

alin04:
- Force PayStartFlow to only accept a SubscriptionRequest
- Quote object keys to protect against minification mismatch.
- Move some logic out of constructor closer to where the values are
used.

### DIFF
--- a/src/runtime/offers-flow.js
+++ b/src/runtime/offers-flow.js
@@ -87,13 +87,8 @@ export class OffersFlow {
         .toArray(),
     };
 
-    this.prorationMode = feArgsObj['replaceSkuProrationMode'] || undefined;
-
     if (options && options.oldSku) {
       feArgsObj['oldSku'] = options.oldSku;
-    }
-
-    if (feArgsObj['oldSku']) {
       assert(feArgsObj['skus'], 'Need a sku list if old sku is provided!');
 
       // Remove old sku from offers if in list.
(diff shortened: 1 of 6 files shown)


jpettitt:
pay needs to be initialized to handle the redirect flow.   You'll need to check for the # fragment that is the return from redirect and lauch pay to handle it if it's present.

alin04:
> pay needs to be initialized to handle the redirect flow. You'll need to check for the # fragment that is the return from redirect and lauch pay to handle it if it's present.

Grrr. So the issue is that for tests, ConfiguredRuntime eagerly initializes ALL its dependencies. And right now in tests, the deprecated PayClientBindingSwg is bound in tests (because the gpay-api experiment isn't enabled everywhere in our tests).

Since this PR removes that bindning, now PayClientBindingJs is eagerly initialized and the second problem I run into is that tests don't resolve all the config vars (e.g. $payEnvironment$).

Sooo, I had started looking into a dependency injection framework, but thought that might be overkill and opted for this lazy instantiation. And redirect flow again bites us in the backside.

I think the easiest solution then is to come up with a way for tests to resolve config vars.... Ideas appreciated.

alin04:
Oh, and I *think* you were actually commenting on my other PR? This one doesn't change how any of the deps are initialized.

src/runtime/offers-flow.js
```diff
@@ -116,11 +111,10 @@ export class OffersFlow {
...
       if (oldSku) {
-        new PayStartFlow(this.deps_, {
-          skuId: sku,
-          oldSku,
-          replaceSkuProrationMode: this.prorationMode,
-        }).start();
+        const skuSelectedResponse = new SkuSelectedResponse();
+        skuSelectedResponse.setSku(sku);
+        skuSelectedResponse.setOldSku(oldSku);
+        this.startPayFlow_(skuSelectedResponse);
```
Kestrad: Doesn't this change mean that even if a publisher passes in a proration mode to the flow (which I realize currently isn't supported), offers-flow now has no way of passing it along?

alin04: Yes, I removed it because this.prorationMode was always undefined. I can put it back with a TODO.

Kestrad: That makes sense. So if we were to add that functionality in the future, we'd want to update skuSelectedResponse to have a proration mode field, right?

alin04: Yeah, that sounds right

PR #937, opened 2020-01-03
SwG early logging

mborof:
This development is being done to attempt to prevent SwG from losing log events.  The theory is that the page is being redirected to gPay before the "payment started" event is sent.

Measuring of "time to first log" was done on a local glinux machine.  Time varied between .3s and .6s.  Measuring of subsequent logs was also done, this was always less than .05s.

This development incorporates a number of changes to:
  1) Attempt to get the logging iframe loaded earlier in the system lifecycle
  2) Increase the amount of waiting for the "first log" to send (so the iframe has time to load) -> 250ms
       a) Question for John: can we increase this "first log wait" to .5s?  This is a corner case and will only effect a user if the publisher links to a page and immediately initiates a request to pay.  This is how long it seems to actually take for that first log.
  3) Add 25 additional MS of waiting when redirecting and a log didn't get sent in time.  This is necessary because SwG considers the log event sent when the iframe indicates it is done.  The iframe may not actually have a sent the log out yet.


jpettitt:
I think the initial 0.5s is fine,  I'm not sure it 25ms is going to be enough on a phone connection, 100ms might be better.  Or better still find a way for the iframe to signal it's done if that's not too complex.

mborof:
I added a "logging finished" signal in late December that we are already expecting.  I just discovered that the inner JavaScript object (the one provided by the clearcut team) also works asynchronously and so I need to add some more code to wait for that to finish before the finished signal is sent.  I have a CL for it and it will take 2-3 days to actually propagate once it is submitted.

As such, I removed that 25ms extra wait time since we should theoretically already be waiting for the log to be sent.

I also bumped the "first wait" up to 500ms.  I'm slightly concerned it might not be enough time on a phone but again this is the rare corner case of the publisher just immediately starting a buyflow.

I also bumped the "subsequent waits" up to 200ms.  We could *probably* safely lower this some but I wouldn't expect the system to ever actually wait this long unless perhaps someone's phone lost signal and the redirect to gPay is going to fail anyway...

PR #946, opened 2020-01-09
send new parameter structure to propensity

mborof:
In order to avoid double encoding some values, Propensity will eventually change their URL structure.  For now, this sends that data both ways.  Eventually the old method can be removed.

### DIFF
--- a/src/runtime/propensity-server.js
+++ b/src/runtime/propensity-server.js
@@ -103,15 +103,18 @@ export class PropensityServer {
       method: 'GET',
       credentials: 'include',
     });
-    const stateParam = 'states=' + this.publicationId_ + ':' + state;
-    const extrainfoParam = '&extrainfo=';
+    let stateParam = this.publicationId_ + ':' + state;
+    let extrainfoParam = '&extrainfo=';
     if (productsOrSkus) {
       const skus = encodeURIComponent(productsOrSkus);
       extrainfoParam += skus;
       //TODO: Remove this once propensity officially moves over to the new param
       stateParam += ':' + skus;
     }
-    const url = adsUrl('/subopt/data?') + stateParam + extrainfoParam;
+    const url =
+      adsUrl('/subopt/data?states=') +
+      encodeURIComponent(stateParam) +
+      extrainfoParam;
     return this.fetcher_.fetch(this.propensityUrl_(url), init);
   }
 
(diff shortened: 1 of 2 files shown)


src/runtime/propensity-server.js
```diff
@@ -103,11 +103,18 @@ export class PropensityServer {
       method: 'GET',
       credentials: 'include',
     });
-    let userState = this.publicationId_ + ':' + state;
+    let stateParam = this.publicationId_ + ':' + state;
+    let extrainfoParam = '&extrainfo=';
```
ChrisAntaki: Would it be possible to use a more descriptive parameter name? `extrainfo` could apply to all kinds of data :smile:

src/runtime/propensity-server.js
```diff
@@ -103,11 +103,18 @@ export class PropensityServer {
...
     if (productsOrSkus) {
-      userState = userState + ':' + encodeURIComponent(productsOrSkus);
+      const skus = encodeURIComponent(productsOrSkus);
```
ChrisAntaki: Why does the variable name switch from `productsOrSkus` to just `skus` here?

src/runtime/propensity-server.js
```diff
@@ -103,11 +103,18 @@ export class PropensityServer {
...
+      extrainfoParam += skus;
+      //TODO: Remove this once propensity officially moves over to the new param
+      stateParam += ':' + skus;
     }
-    const url = adsUrl('/subopt/data?states=') + encodeURIComponent(userState);
+    const url =
+      adsUrl('/subopt/data?states=') +
+      encodeURIComponent(stateParam) +
```
ChrisAntaki: Could you use the `addQueryParam` util method to add the query params? It handles URL encoding, checking whether `?` or `&` is needed, etc

src/runtime/propensity-server.js
mborof: The names they picked were much worse and they chose this one.  Its too late to change.  

It is a fair name because it is a generic JSON object that contains any number of different possible data elements and all of them are techincally "extra info" beyond the basic information passed to the service.

src/runtime/propensity-server.js
mborof: Its just a local const so it wasn't as important to have a clear name - I removed the const anyway because we don't need to send the value twice anyway (as they aren't using the JSON object right now).

ChrisAntaki: Could it still contain products then? Looks like the `const` is still around

mborof: I just hadn't updated the code yet

src/runtime/propensity-server.js
mborof: Done

PR #952, opened 2020-01-14
gPay Logging Expectation Changes

mborof:
This repairs the expectations when returning from gPay.  In order to determine whether a redirect occurred, we will now check the analytics service to see if the user was sent to gPay (if they weren't sent to gPay it means the user was redirected there because we lost JavaScript memory).  

This replaces a "had ever logged" flag which will not work anymore due to the introduction of a "page load" log event.

### DIFF
--- a/src/runtime/analytics-service.js
+++ b/src/runtime/analytics-service.js
@@ -86,7 +86,7 @@ export class AnalyticsService {
     this.doc_.getBody().appendChild(this.getElement());
 
     /** @private @type {!boolean} */
-    this.everStartedLog_ = false;
+    this.waitingOnGpay_ = false;
 
     /** @private @type {!boolean} */
     this.everFinishedLog_ = false;
@@ -267,11 +267,16 @@ export class AnalyticsService {
   }
 
   /**
-   * Returns true if any logs have already be sent to the analytics server.
+   * Returns true if the user was sent to gPay (and not redirected).
    * @return {boolean}
    */
-  getHasLogged() {
-    return this.everStartedLog_;
+  getWaitingOnGpay() {
+    return this.waitingOnGpay_;
+  }
+
+  /** */
+  setDoneWaitingOnGpay() {
+    this.waitingOnGpay_ = false;
   }
 
   /**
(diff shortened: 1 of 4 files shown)


src/runtime/analytics-service.js
```diff
@@ -322,22 +327,31 @@ export class AnalyticsService {
...
     if (
       ClientEventManager.isPublisherEvent(event) &&
       !this.shouldLogPublisherEvents_() &&
       !this.shouldAlwaysLogEvent_(event)
     ) {
       return;
     }
+
+    switch (event.eventType) {
+      //this event is just used to communicate information internally.  It should
```
henel677: // This...

src/runtime/analytics-service.js
```diff
@@ -322,22 +327,31 @@ export class AnalyticsService {
...
+      //not be reported to the SwG analytics service.
+      case AnalyticsEvent.EVENT_SUBSCRIPTION_STATE:
+        return;
+      case AnalyticsEvent.ACTION_PAYMENT_FLOW_STARTED:
+        if (this.waitingOnGpay_) {
+          //TODO: Consider a permanent log event for this
```
henel677: // TODO

src/runtime/analytics-service.js
```diff
@@ -267,11 +267,16 @@ export class AnalyticsService {
...
+   * Returns true if the user was sent to gPay (and not redirected).
    * @return {boolean}
    */
-  getHasLogged() {
-    return this.everStartedLog_;
+  getWaitingOnGpay() {
+    return this.waitingOnGpay_;
+  }
+
+  /** */
```
henel677: Do you need this line? If you do It is nicer if you add some description even if it is mostly restating what can be deduced from the name of the function

src/runtime/pay-flow.js
```diff
@@ -389,6 +392,7 @@ function validatePayResponse(deps, payPromise, completeHandler) {
       }
     }
     deps.eventManager().logSwgEvent(eventType, true, eventParams);
+    deps.analytics().setDoneWaitingOnGpay();
```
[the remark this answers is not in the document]
mborof: I am moving this dev to a new branch where I am making the pay logic self reliant (and removing this reliance on the logging system).  I did account for this comment in that new branch though.  Good idea!

PR #954, opened 2020-01-15
Ensure "WasRedirected" Calculation is Correct

mborof:
This makes the PayClient self reliant in determining if the user is returning from a popup or a redirect and removes code from the analytics client that was being used to make this determination.

### DIFF
--- a/src/runtime/analytics-service-test.js
+++ b/src/runtime/analytics-service-test.js
@@ -294,6 +294,25 @@ describes.realWin('AnalyticsService', {}, env => {
       expect(loggedErrors.length).to.equal(1);
       expect(loggedErrors[0]).to.equal('Error when logging: ' + err);
     });
+
+    it('Should work without any promises', async () => {
+      // This sends another event and waits for it to be sent
+      eventManagerCallback({
+        eventType: AnalyticsEvent.IMPRESSION_PAYWALL,
+        eventOriginator: EventOriginator.SWG_CLIENT,
+        isFromUserAction: true,
+        additionalParameters: {droppedData: true},
+      });
+      await analyticsService.lastAction_;
+      const loggingResponse = new FinishedLoggingResponse();
+      loggingResponse.setComplete(true);
+      // Simulate 1 logging response
+      iframeCallback(loggingResponse);
+      expect(analyticsService.unfinishedLogs_).to.equal(0);
+      return analyticsService.getLoggingPromise().then(val => {
+        expect(val).to.be.true;
+      });
+    });
   });
 
   it('should not log the subscription state change event', () => {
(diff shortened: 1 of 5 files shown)


src/runtime/analytics-service-test.js
```diff
@@ -118,6 +118,13 @@ describes.realWin('AnalyticsService', {}, env => {
       analyticsService.setTransactionId(txId);
       expect(analyticsService.getTransactionId()).to.equal(txId);
     });
+
+    it('should close', () => {
+      const activityIframe = analyticsService.getElement();
+      expect(activityIframe.parentNode === runtime.doc().getBody()).to.true;
```
henel677: Is it possible to use:
expect(...).to.equal(...) instead
?

src/runtime/analytics-service-test.js
```diff
@@ -287,6 +294,26 @@ describes.realWin('AnalyticsService', {}, env => {
...
+
+    it('Should work without waiting for the promise', async () => {
+      // This sends another event and waits for it to be sent
+      eventManagerCallback({
+        eventType: AnalyticsEvent.IMPRESSION_PAYWALL,
+        eventOriginator: EventOriginator.SWG_CLIENT,
+        isFromUserAction: true,
+        additionalParameters: {droppedData: true},
+      });
+      await analyticsService.lastAction_;
```
henel677: Does it make sense to verify that that lastAction_ is not null before waiting for it (apparently waiting for non-promises is valid and returns the non-promise)?

src/runtime/pay-flow.js
```diff
@@ -143,6 +143,7 @@ export class PayStartFlow {
       true,
       getEventParams(swgPaymentRequest['skuId'])
     );
+    PayCompleteFlow.waitingForGpay_ = true;
```
alin04: To be pedantic, this variable is being set a bit earlier than actual gpay invocation. But I suppose it's inconsequential.

mborof: Should I change the flag to "waitingForPayClient_"?

alin04: sure

src/runtime/analytics-service-test.js
```diff
@@ -119,6 +119,13 @@ describes.realWin('AnalyticsService', {}, env => {
       analyticsService.setTransactionId(txId);
       expect(analyticsService.getTransactionId()).to.equal(txId);
     });
+
+    it('should close', () => {
+      const activityIframe = analyticsService.getElement();
+      expect(activityIframe.parentNode).to.equal(runtime.doc().getBody());
+      analyticsService.close();
+      expect(activityIframe.parentNode === null).to.true;
```
henel677: I think this could also use "....to.not.be.null;"

PR #956, opened 2020-01-16
SwG Server to Client Side Logging

mborof:
Currently, when a boq iframe wants to log this request is sent directly to the SwG analytics server.  Events communicated in this manner will not get sent to Propensity.  In order to resolve this issue, the SwG client will listen for AnalyticsRequests messaged to it from inside the boq iframes.  It will then redirect those events to the event manager which can send them out to both SwG analytics and Propensity.

Old path:
  boq iframe -> analytics service -> analytics server

New path:
  boq iframe -> swg client -> event manager -> { analyticsServer, propensity }

This only allows the client to listen for these events and does not actually implement any logging of this nature.  That will come in future development.

### DIFF
--- a/src/components/activities.js
+++ b/src/components/activities.js
@@ -282,6 +282,7 @@ export class ActivityPorts {
         'publicationId': pageConfig.getPublicationId(),
         'productId': pageConfig.getProductId(),
         '_client': 'SwG $internalRuntimeVersion$',
+        'supportsClientLogging': true,
       },
       args || {}
     );
--- a/src/components/activities.js
+++ b/src/components/activities.js
@@ -235,7 +235,7 @@ export class ActivityIframePort {
     try {
       label = getLabel(message);
     } catch (ex) {
-      // Thrown is message is not a proto object
+      // Thrown if message is not a proto object and has no label
       label = null;
     }
     if (!label) {
--- a/src/components/activities.js
+++ b/src/components/activities.js
@@ -14,10 +14,15 @@
  * limitations under the License.
  */
 import {
+  AnalyticsRequest,
+  EventOriginator,
+  deserialize,
+  getLabel,
+} from '../proto/api_messages';
+import {
   ActivityIframePort as WebActivityIframePort,
   ActivityPorts as WebActivityPorts,
 } from 'web-activities/activity-ports';
-import {deserialize, getLabel} from '../proto/api_messages';
 
 /**
  * @interface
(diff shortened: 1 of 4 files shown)


src/components/activities.js
```diff
@@ -216,7 +231,10 @@ export class ActivityIframePort {
    * @template T
    */
   on(message, callback) {
-    const label = getLabel(message);
+    let label = null;
+    try {
+      label = getLabel(message);
+    } catch (ex) {}
```
ChrisAntaki: Could you add a comment describing why this `catch` block is empty?

ChrisAntaki: I understand why it's empty, but if we could make that really obvious for future editors who might be in a rush, that would be awesome :)

mborof: done

src/components/activities.js
```diff
@@ -216,7 +231,13 @@ export class ActivityIframePort {
    * @template T
    */
   on(message, callback) {
-    const label = getLabel(message);
+    let label = null;
+    try {
+      label = getLabel(message);
+    } catch (ex) {
+      // Thrown is message is not a proto object
```
ChrisAntaki: I'm having trouble parsing this sentence. Could you try rewriting for a dummy like me? :)

mborof:
A note that I added a flag so the server can intelligently decide whether to send logs to the client or to log itself directly.

src/components/activities.js
```diff
@@ -282,6 +282,7 @@ export class ActivityPorts {
         'publicationId': pageConfig.getPublicationId(),
         'productId': pageConfig.getProductId(),
         '_client': 'SwG $internalRuntimeVersion$',
+        'supportsClientLogging': true,
```
ChrisAntaki: Which client is `supportsClientLogging` referring to?

mborof: It is a flag to tell the server that the server can message log requests back to the client and that the client has the wiring to handle that logging.  This way when I re-write the server logging to message back to the client it can:

if (supports) -> message to client -> client messages to {clearcut, propensity}
else -> message to clearcut

You can actually see most of that change in a CL now (just sent to you), I just can't get the tests to run due to contribution iframe issues unrelated to my dev.

This is necessary because:
  1) On mondays we released swg-js which is immediately sent out to everyone not on AMP
  2) It takes 1 weeks to get promoted from the amp canary to amp prod
  3) There are additional references to old versions of swg in the AMP cache that can take weeks to actually update
  4) There are random old clients out there that I have yet to explain.  Probably because a publisher copied our code instead of referencing through Google properly.

ChrisAntaki: So I consider any JavaScript code running in the browser being on the "client side". Could we pick a more descriptive name for this flag? Something like, "supportsExternalLogging"?

mborof: supportsEventManager

PR #958, opened 2020-01-16
Set logging context after it is available

mborof:
The createLogRequest function relies on a context object that isn't fully set up until after the analytics service is wired up and ready to go.  The code was reading the object too soon resulting in some lost data in our logging system.


mborof:
A note that I did not update test expectations here - expectations haven't changed.  This kind of issue could only have been caught by an integration test since it relies on multiple parts of the system interacting for us to catch.

mborof:
The context is passed by reference so it is automatically updated everywhere when it does finally update.  This was never a real problem.

PR #959, opened 2020-01-17
Resolve Issues with Analytics Context

mborof:
The problems:
1) The core problem is that "openIframe" (called in analytics-service before it sets up context) relies on the analytics-service context.  
    a) This means that some logs will be missing context information (such as client version).
2) The secondary problem is that one of the context properties (experiments) can change dynamically or might not all be set when analytics service is constructing at load.

This development:
  1) Sets up context during load for all properties that can be set.
  2) Waits a little longer to set experiments and does so before calling openIframe
  3) Fixes them again once the port is ready.
  3) Ensures runtime forwards experiment changes to analytics so it has the current state.

There is still a "hole" here in that openIframe may not have some experiment data in it (if it is set late in the code life cycle.  This is a minor problem and eventually the analytics context is going to be removed from openIframe so the problem will go away.

### DIFF
--- a/src/runtime/analytics-service.js
+++ b/src/runtime/analytics-service.js
@@ -95,7 +95,7 @@ export class AnalyticsService {
      * @private @const {!AnalyticsContext}
      */
     this.context_ = new AnalyticsContext();
-    this.context_.setTransactionId(getUuid());
+    this.setStaticContext_();
 
     /** @private {?Promise<!web-activities/activity-ports.ActivityIframePort>} */
     this.serviceReady_ = null;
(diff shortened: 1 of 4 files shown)


src/runtime/analytics-service.js
```diff
@@ -196,39 +196,50 @@ export class AnalyticsService {
...
         .openIframe(this.iframe_, feUrl('/serviceiframe'), null, true)
         .then(
           port => {
             // Register a listener for the logging to code indicate it is
             // finished logging.
             port.on(FinishedLoggingResponse, this.afterLogging_.bind(this));
-            this.setContext_();
-            return port.whenReady().then(() => port);
+            return port.whenReady().then(() => {
+              // The publisher should be done setting experiments but runtime
```
ChrisAntaki: Thanks for adding these descriptive comments throughout this PR

PR #962, opened 2020-01-21
Set Canonical URL in Analytics Service

mborof:


src/runtime/analytics-service.js
```diff
@@ -155,6 +155,13 @@ export class AnalyticsService {
     this.context_.setSku(sku);
   }
 
+  /**
+   * @param {string} url
+   */
+  setUrl(url) {
+    this.context_.setUrl(url);
```
ChrisAntaki: Could you give this method test coverage?

mborof: Ok

PR #967, opened 2020-01-23
✔️ Refactors 'client-event-manager' tests and brings coverage to 100%

ChrisAntaki:


src/runtime/client-event-manager-test.js
```diff
@@ -34,191 +35,202 @@ const DEFAULT_EVENT = {
...
-        event.eventOriginator = null;
-        expected = 'Event has an invalid eventOriginator(' + null + ')';
-        tryIt();
-        expect(errorCount).to.equal(2);
-        expect(matchedExpected).to.equal(2);
-        event.eventOriginator = OTHER_ORIGIN;
-        tryIt();
-        expect(errorCount).to.equal(2);
-        expect(matchedExpected).to.equal(2);
-        event.eventOriginator = DEFAULT_ORIGIN;
```
ChrisAntaki: It looks like this case wasn't being tested

src/runtime/client-event-manager-test.js
```diff
@@ -34,191 +35,202 @@ const DEFAULT_EVENT = {
...
-        expect(matchedExpected).to.equal(1);
-        event.isFromUserAction = true;
-        tryIt();
-        expect(errorCount).to.equal(1);
-        expect(matchedExpected).to.equal(1);
-        event.isFromUserAction = false;
-        tryIt();
-        expect(errorCount).to.equal(1);
-        expect(matchedExpected).to.equal(1);
-        event.isFromUserAction = null;
```
ChrisAntaki: It looks like this case wasn't being tested

src/runtime/client-event-manager-test.js
```diff
@@ -34,191 +35,202 @@ const DEFAULT_EVENT = {
...
 
-      it('should validate additionalParameters', () => {
-        event.additionalParameters = BAD_VALUE;
-        expected =
-          'Event has an invalid additionalParameters(' + BAD_VALUE + ')';
-        tryIt();
-        expect(errorCount).to.equal(1);
-        expect(matchedExpected).to.equal(1);
-        event.additionalParameters = null;
-        expected = 'Event has an invalid additionalParameters(' + null + ')';
```
ChrisAntaki: Was `null` expected to be an invalid value? It doesn't cause a throw

src/runtime/client-event-manager-test.js
```diff
@@ -34,272 +35,310 @@ const DEFAULT_EVENT = {
...
+    let eventManager;
 
     beforeEach(() => {
-      eventMan = new ClientEventManager(RESOLVED_PROMISE);
+      eventManager = new ClientEventManager(RESOLVED_PROMISE);
     });
 
     it('should identify publisher events', () => {
-      const testIsPublisherEvent = function(originator, isPublisherEvent) {
-        DEFAULT_EVENT.eventOriginator = originator;
```
ChrisAntaki: Looks like this was mutating a `const` shared across most of the tests

src/runtime/client-event-manager-test.js
```diff
@@ -34,272 +35,310 @@ const DEFAULT_EVENT = {
...
           : EventManagerApi.FilterResult.PROCESS_EVENT
       );
 
-      //ensure the default origin is filtered out
-      eventMan.logEvent(DEFAULT_EVENT);
-      await eventMan.lastAction_;
-      expect(counter).to.equal(0);
-
-      //ensure the other origin is not filtered out
-      DEFAULT_EVENT.eventOriginator = OTHER_ORIGIN;
```
ChrisAntaki: Looks like this was mutating a `const` shared across most the tests

src/runtime/client-event-manager-test.js
mborof: Null is allowed, it just means there are no additional parameters and is the most common logging case.

src/runtime/client-event-manager-test.js
```diff
@@ -34,272 +35,310 @@ const DEFAULT_EVENT = {
...
+      expect(events).to.deep.equal([]);
     });
 
-    it('should not log events until its promise is resolved', async () => {
-      let counter2 = 0;
+    it('should log events after configuration succeeds', async () => {
+      // Request event log.
+      eventManager.logEvent(DEFAULT_EVENT);
+
+      indicateConfigurationSucceeded();
```
mborof: We should do an await tick before indicating configuration success.  Then check to ensure no log was sent.  Then indicate configuration success.  Then await the logging promise (instead of tick which might be flaky).  Then check to ensure events === [DEFAULT_EVENT]

src/runtime/client-event-manager-test.js
```diff
@@ -34,272 +35,310 @@ const DEFAULT_EVENT = {
...
     });
 
-    it('should not log events if promise rejected', async () => {
-      rejector();
-      try {
-        await eventMan.lastAction_;
-      } catch (err) {}
-      expect(counter).to.equal(0);
+    it('returns ready promise', () => {
+      expect(eventManager.getReadyPromise()).to.equal(configurationPromise);
```
mborof: It surprises me that this test passes.  The configuration promise (passed in by the caller) is different from the ready promise (awaiting iframe load).

src/runtime/client-event-manager-test.js
ChrisAntaki: This is covered in the test above: 'should not log events before configuration succeeds'

src/runtime/client-event-manager-test.js
ChrisAntaki: In `client-event-manager.js` I'm seeing `this.isReadyPromise_ = configuredPromise;`

src/runtime/client-event-manager-test.js
mborof: It was mutating then unmutating a sub value of the const which is valid but potentially slightly confusing.

mborof: e.g. 
const x = {a: 1};
x.a = 5;

is valid but potentially confusing

ChrisAntaki: I don't see the `DEFAULT_EVENT` object being mutated again in this function

ChrisAntaki: I think we should avoid mutating shared state across tests (if possible), since that can cause flakiness

PR #980, opened 2020-02-01
DO NOT SUBMIT: Removed experiment flags from runtime methods for Update Subscription

Kestrad:


src/runtime/runtime.js
```diff
@@ -775,10 +775,6 @@ export class ConfiguredRuntime {
 
   /** @override */
   showUpdateOffers(options) {
-    assert(
```
alin04: Should we ramp the experiment to 100% first? Or instruct the publisher to override with query param (if all they're doing is testing)?

Kestrad: @jpettitt I think?

Kestrad: After talking to Angie and John we've decided to ramp the experiment instead. Keeping this PR open for the time being, but not submitting.

ChrisAntaki:
Thanks for your work on this, @Kestrad !

Issue #966, opened 2020-01-22
`reset()` and `clear()` do not refresh the page config


ed-bird:
When operating a Single Page Application (SPA), you will need to reload the configuration to faciliate switching between different types of webpage. For example:

- Go from the homepage to a basic article (no markup to markup with `"productID: norcaltribune.com:basic"`)
- Go from a basic article to a premium article (markup with `productId: "norcaltribune.com:basic"` to markup with `productId: "norcaltribune.com:premium"`

At the moment, the existing methods `reset()` and `clear()` only clear the entitlements and open dialogs but do not reset the page config which means that the flow described above cannot currently be supported.

ed-bird:
A workaround for this is as follows:
- [Initialise swg.js manually](https://github.com/subscriptions-project/swg-js/blob/master/docs/embed-client.md#manual-initialization) when the application loads
- For each article page:
   - Do *not* re-call `subscriptions.init()`, this will error out saying that the library is already configured
   - Call `subscriptions.getEntitlements()` and check the full entitlements response to see if any of the entitlements are valid
   - Do *not* use `entitlements.enablesThis()` as it will be based on the configuration on the original page, not necessarily the configuration on the current page

It would still be good for `reset()` and `clear()` to be updated to support this flow though.

PR #985, opened 2020-02-07
Create Dual Logging Experiment

mborof:
Adds an experiment which causes the system to log to both the old place and the new place so we can ensure everything works properly.

Pre-defending the serialization code: 
1) The format output by our protos includes a "message type" label as the first part of the array.  The BOQ server does not like that and expects the array to not include the message type label.  Hence the strange shifting during serialization.
2) I reduced the footprint by changing true/false to 1/0 which seems to work as expected
3) It would be a boon if AMP could somehow share that serialization code since it will currently need to re-write the exact same code in its custom fetcher.  I'm not sure about sharing it between the two though since it would need to be exposed in an API at that point.  This is kind of consistent with the existing fetcher in AMP which is just calling a slightly different path to do the same thing already.


mborof:
Note that the server currently "does nothing".  For now we will verify communications work only.

src/utils/types.js
```diff
@@ -58,3 +58,11 @@ export function isFunction(value) {
 export function isBoolean(value) {
   return typeof value === 'boolean';
 }
+
+/**
+ * @param {?*} data
+ * @return {boolean}
+ */
+export function isMessage(data) {
```
ChrisAntaki: `Message` is a kind of general name, could we call this `isProtoMessage` to be more specific?

src/runtime/fetcher.js
```diff
@@ -60,4 +68,19 @@ export class XhrFetcher {
   fetch(url, init) {
     return this.xhr_.fetch(url, init);
   }
+
+  /** @override */
+  sendBeacon(url, data) {
+    url = addQueryParam(url, 'f.req', serializeMessageForUrl(data));
+    if (navigator.sendBeacon) {
+      navigator.sendBeacon(url);
```
ChrisAntaki: Why not send the `data` as the second parameter to `sendBeacon`, which includes it in the POST body instead of the URL, to avoid serialization and potentially running out of space in the URL? https://developer.mozilla.org/en-US/docs/Web/API/Navigator/sendBeacon

src/runtime/fetcher.js
```diff
@@ -60,4 +68,19 @@ export class XhrFetcher {
...
+      return;
+    }
+    const init = /** @type {!../utils/xhr.FetchInitDef} */ ({
```
ChrisAntaki: Could you add a comment saying it's falling back to a normal XHR here?

src/utils/url.js
```diff
@@ -195,6 +197,47 @@ export function addQueryParam(url, param, value) {
...
+ * @return {*}
+ */
+function prepForSerialization(data) {
+  if (data === null || data === undefined) {
+    return undefined;
+  }
+  let arr = null;
+  switch (typeof data) {
+    case 'boolean':
+      return data ? 1 : 0;
```
ChrisAntaki: `true` and `false` are valid values in JSON, right? If so, why convert to numbers here?

src/utils/url.js
```diff
@@ -195,6 +197,47 @@ export function addQueryParam(url, param, value) {
...
+    arr[i] = prepForSerialization(arr[i]);
+  }
+  return arr;
+}
+
+/**
+ * @param {!../proto/api_messages.Message} message
+ * @return {string}
+ */
+export function serializeMessageForUrl(message) {
```
ChrisAntaki: Could you rename this to `serializeProtoMessageForUrl` just so it's clear what kind of message we're serializing?

jpettitt:
What's the issue with the AMP fetcher?

src/runtime/analytics-service-test.js
```diff
@@ -246,6 +253,39 @@ describes.realWin('AnalyticsService', {}, env => {
...
+
+        // These wait for analytics server to be ready to send data.
+        expect(analyticsService.lastAction_).to.not.be.null;
+        await analyticsService.lastAction_;
+        await activityIframePort.whenReady();
+
+        expectOpenIframe = true;
+        expect(eventsLoggedToService.length).to.equal(expectedEventCount);
+      });
+      it('should not log to clearcut if experiment off', () => {});
```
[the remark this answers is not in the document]
mborof: I like having shared code when its identical - what do you think about the function I created instead?

src/runtime/analytics-service-test.js
```diff
@@ -478,6 +518,12 @@ describes.realWin('AnalyticsService', {}, env => {
     });
   });
 
+  describe('SwG Clearcut Logging Experiment', () => {
```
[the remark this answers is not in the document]
mborof: No TODO required, the code was just moved above.

src/utils/url.js
mborof: That's what the original clearcut logging JavaScript was doing and 1 is shorter than 'true' or 'false' making it a more compact representation.

I'm open to not making that flip though, it does make the testing harder.

mborof: Removed this

mborof:
> What's the issue with the AMP fetcher?

AMP Fetcher will need to implement the "sendBeacon" method I added to the common interface before we can release any SwG client with this code in it.  I plan to implement the sendBeacon method before this is submitted and then during the SwG release we can just add the @override tag to the new method.

Fortunately I am doing the release this Monday so I can just handle it myself.

src/runtime/fetcher.js
mborof: I spent 2 days trying every possible serialization format using the second parameter.  BOQ rejected them all.  I had to manually hack at BOQ until I got it to accept one of the serialization formats I randomly guessed (this odd query string thing is the only one that worked).  

I was not able to find any assistance in YAQS or moma defining the format it was expecting (they just tell you to use JSPB which we can't use).

I have at this point been able to confirm that the server is saving the proto I send to clearcut so deserialization is working right now.

src/runtime/analytics-service.js
```diff
@@ -65,8 +67,15 @@ function createErrorResponse(error) {
...
    * @param {!./deps.DepsDef} deps
+   * @param {!./fetcher.Fetcher} fetcher
    */
-  constructor(deps) {
+  constructor(deps, fetcher) {
+    /** @private @const {!./fetcher.Fetcher} */
+    this.fetcher_ = fetcher;
+
+    /** @private @const {string} */
+    this.publicationId_ = deps.pageConfig().getPublicationId();
```
alin04: In light of #966, maybe pubId shouldn't be cached/stored.

src/runtime/fetcher.js
alin04: The server should be able to handle binary protobuf wire format for the request body. Not sure if our lightweight js protobufs have a way to serialize to binary format.

mborof: Do you have any idea what the binary format looks like?  Because I couldn't find anything defining the proper format, I just guessed until I found something that worked.

mborof: The blob method you indicated I should use is rejected by Chrome currently because of a security issue.  The URL size in scenic is under 1000 - I do agree this is a slightly concern of the 2048 character URL limit though I think nothing should actually hit that.

src/runtime/analytics-service-test.js
```diff
@@ -246,6 +253,42 @@ describes.realWin('AnalyticsService', {}, env => {
...
+        expect(eventsLoggedToService.length).to.equal(expectedEventCount);
+      }
+
+      beforeEach(() => {
+        // This ensure nothing gets sent to the server.
+        sandbox.stub(activityIframePort, 'execute').callsFake(() => {});
+      });
+
+      afterEach(async () => {});
+      it('should not log to clearcut if experiment off', async () =>
```
ChrisAntaki: https://google.github.io/styleguide/jsguide.html#features-functions-arrow-functions
> By default the body is a block statement (zero or more statements surrounded by curly braces). The body may also be an implicitly returned single expression if either: the program logic requires returning a value, or the void operator precedes a single function or method call (using void ensures undefined is returned, prevents leaking values, and communicates intent).

ChrisAntaki: Could you also add a line between the two `it`s?

src/runtime/fetcher.js
```diff
@@ -46,18 +54,44 @@ export class XhrFetcher {
...
+      'f.req',
+      serializeProtoMessageForUrl(data)
+    );
+    if (navigator.sendBeacon) {
+      navigator.sendBeacon(newUrl);
+      return;
+    }
+    // Only newer browsers support beacon.  Fallback to standard XHR POST.
+    this.fetch(newUrl, this.getCredentialedInit_('POST'));
+    return;
```
ChrisAntaki: This `return` could be removed

src/runtime/analytics-service-test.js
```diff
@@ -246,6 +253,42 @@ describes.realWin('AnalyticsService', {}, env => {
...
+        expectOpenIframe = true;
+        expect(eventsLoggedToService.length).to.equal(expectedEventCount);
+      }
+
+      beforeEach(() => {
+        // This ensure nothing gets sent to the server.
+        sandbox.stub(activityIframePort, 'execute').callsFake(() => {});
+      });
+
+      afterEach(async () => {});
```
ChrisAntaki: Looks like this line could get removed

src/runtime/analytics-service-test.js
```diff
@@ -246,6 +253,42 @@ describes.realWin('AnalyticsService', {}, env => {
...
+        expect(analyticsService.lastAction_).to.not.be.null;
+        await analyticsService.lastAction_;
+        await activityIframePort.whenReady();
+
+        expectOpenIframe = true;
+        expect(eventsLoggedToService.length).to.equal(expectedEventCount);
+      }
+
+      beforeEach(() => {
+        // This ensure nothing gets sent to the server.
```
ChrisAntaki: Should "ensure" be plural?

src/runtime/analytics-service-test.js
```diff
@@ -246,6 +253,42 @@ describes.realWin('AnalyticsService', {}, env => {
...
+          additionalParameters: null,
+        });
+
+        // These wait for analytics server to be ready to send data.
+        expect(analyticsService.lastAction_).to.not.be.null;
+        await analyticsService.lastAction_;
+        await activityIframePort.whenReady();
+
+        expectOpenIframe = true;
+        expect(eventsLoggedToService.length).to.equal(expectedEventCount);
```
ChrisAntaki: > I like having shared code when its identical - what do you think about the function I created instead?

I think this TotT makes some good points for readability over uniqueness in tests specifically
go/tott/598

src/utils/url-test.js
```diff
@@ -258,3 +260,77 @@ describe('addQueryParam', () => {
     expect(addQueryParam('file?d=e#f', 'a', 'b')).to.equal('file?d=e&a=b#f');
   });
 });
+
+describe('serializeProtoMessageForUrl', () => {
```
ChrisAntaki: Could we simplify both of these tests to just verify `JSON.stringify(inputArray.slice(1))` matches the result of `serializeProtoMessageForUrl`?

src/runtime/fetcher.js
```diff
@@ -46,18 +54,44 @@ export class XhrFetcher {
...
   }
 
   /** @override */
   fetch(url, init) {
     return this.xhr_.fetch(url, init);
   }
+
+  /** @override */
+  sendBeacon(url, data) {
+    const newUrl = addQueryParam(
```
ChrisAntaki: Could we just modify `url` here? That way later on in the method, there doesn't have to be the concept of a `newUrl`

src/runtime/fetcher.js
```diff
@@ -46,18 +54,44 @@ export class XhrFetcher {
...
+      url,
+      'f.req',
+      serializeProtoMessageForUrl(data)
+    );
+    if (navigator.sendBeacon) {
+      navigator.sendBeacon(newUrl);
```
ChrisAntaki: > The blob method you indicated I should use is rejected by Chrome currently because of a security issue. The URL size in scenic is under 1000 - I do agree this is a slightly concern of the 2048 character URL limit though I think nothing should actually hit that.

Could you describe the security issue in Chrome?

src/utils/url-test.js
mborof: There are sub objects and we need to shift each of the sub arrays EXCEPT for a single sub array that is actually a list of strings and not a sub object.

src/runtime/analytics-service-test.js
mborof: I mostly disagree with the advice.  I find it significantly more time consuming to figure out what dozens of tests do when they all do the same thing over and over again.  I think you are much more likely to miss the 1 random test that does something subtly different because of all the repeated code.  But this is just 2 tests and its not that important to me,

src/runtime/fetcher.js
mborof: Firstly, noting that even if I get Chrome to accept the attempt I'd still need to figure out the right format and that it isn't defined anywhere.

This stack overflow goes through everything:
https://stackoverflow.com/questions/40523469/navigator-sendbeacon-to-pass-header-information

But ultimately when you take the advice on "how to use blob" you get this:
      const header = {
        type: 'text/plain, application/json',
      };
      const blob = new Blob([serializeProtoMessageForUrl(data)], header);
      navigator.sendBeacon(url, blob);

Browser throws:
Uncaught (in promise) DOMException: Failed to execute 'sendBeacon' on 'Navigator': sendBeacon() with a Blob whose type is not any of the CORS-safelisted values for the Content-Type request header is disabled temporarily. See http://crbug.com/490015 for details.

This post is from 2015 so I don't think that's going away anytime soon.

mborof: const header = {
        type: 'application/json',
      };

Also throws the same error.  Not passing headers does too.

src/utils/url-test.js
mborof: Looking at this again, I'm testing an object without sub objects and perhaps I should also test one with sub objects but your suggestion would probably work for these simpler non complicated messages.  

What I'm doing validates it can be deserialized though and that everything can get fed in, serialized, fed out, deserialized and that all the possible outputs are expected.

I do think you could argue some of this is already tested in the api_messages-test.js file.

We may want to have a conversation about this (and the sendBeacon thing).

mborof:
At the moment I am leaning towards releasing this so we can test the basic server actions in dev (and do it concurrently with some broader context logging testing we are outsourcing to gTech soon).  

It will remain behind the experiment while I work on a better serialization / communications protocol.  We will need to finalize that piece before removing it from the experiment.

src/runtime/analytics-service-test.js
```diff
@@ -246,6 +253,51 @@ describes.realWin('AnalyticsService', {}, env => {
       await analyticsService.getLoggingPromise();
       expect(loggedErrors.length).to.equal(0);
     });
+
+    describe('SwG Clearcut Service Experiment', () => {
+      beforeEach(() => {
+        // This ensures nothing gets sent to the server.
+        void sandbox.stub(activityIframePort, 'execute').callsFake(() => {});
```
ChrisAntaki: This `void` keyword seems unnecessary

src/runtime/analytics-service-test.js
```diff
@@ -246,6 +253,51 @@ describes.realWin('AnalyticsService', {}, env => {
...
+        // These wait for analytics server to be ready to send data.
+        expect(analyticsService.lastAction_).to.not.be.null;
+        await analyticsService.lastAction_;
+        await activityIframePort.whenReady();
+
+        expectOpenIframe = true;
+        expect(eventsLoggedToService.length).to.equal(0);
+      });
+
+      it('should not log to clearcut if experiment on', async () => {
```
ChrisAntaki: Should the title of this test change?

src/utils/url-test.js
```diff
@@ -258,3 +260,77 @@ describe('addQueryParam', () => {
...
+  it('should serialize message', () => {
+    const inputArray = [
+      null,
```
ChrisAntaki: Instead of `null` could you add something that looks like a `label` for both tests?

src/runtime/fetcher.js
alin04: FormData is more appropriate, because of the f.req param.

https://developer.mozilla.org/en-US/docs/Web/API/FormData

mborof:
I've actually just gotten a standard POST to work using body (as long as I
disable CSRF protection) but am still struggling with Beacon which seems to
always send the wrong content-type.  FormData sends the wrong kind of
Content-Type header.  Blob I think is the one we want but Chrome has it
disabled due to a security glitch.  It seems like I will need to
implement my own custom deserializer on the BOQ-end.

ChrisAntaki:
Thanks for the update! Too bad about that security glitch

src/runtime/fetcher.js
mborof: It sends the wrong content-type.  Blob is the one we need but Chrome has it disabled.  I am looking into work arounds.

PR #993, opened 2020-02-25
sendBeacon now uses request body

mborof:
Modifies the new sendBeacon interface so it sends logging data via the request body / form submission rather than putting everything into the URL string.


mborof:
Just noting that I very specifically needed to figure out that Blob required a content type of "'application/x-www-form-urlencoded;charset=UTF-8';" before Chrome stopped erroring out when I tried to use blob.

ChrisAntaki:
Nice! Which browsers have you tested at this point?

ChrisAntaki:
Could you verify this works in iOS 12, iOS 13, Chrome, and Firefox?

mborof:
I did just verify it works in "Firefox ESR" which is what we can install on our glinux machines...

PR #994, opened 2020-02-25
Adds old_transaction_id to analytics proto

henel677:
Adds proto definitions for keeping track of old and new transaction ID whenever the ID changes.


src/utils/url-test.js
```diff
@@ -279,7 +279,7 @@ describe('serializeProtoMessageForUrl', () => {
       'baseUrl',
     ];
     const analyticsEventMetaArray = ['AnalyticsEventMeta', 1, true];
-    const eventParamsArray = ['EventParams', 'smartbox', 'gpay', false, 'sku'];
+    const eventParamsArray = ['EventParams', 'smartbox', 'gpay', false, 'sku', ''];
```
mborof: Set the last string to 'othertxid'

PR #997, opened 2020-02-26
Adds logging when transaction ID changes.

henel677:


src/runtime/analytics-service.js
```diff
@@ -137,7 +137,14 @@ export class AnalyticsService {
    * @param {string} transactionId
    */
   setTransactionId(transactionId) {
+    const oldTransactionId = this.context_.getTransactionId();
     this.context_.setTransactionId(transactionId);
+    if (oldTransactionId != null) {
```
mborof: exit early if someone is setting it to the same value.

henel677: Done

PR #999, opened 2020-02-26
Propensity GDPR Support - Incomplete / Experimental Code

mborof:
DO NOT SUBMIT - this is not ready yet.

This is my initial take on GDPR support in SwG.  There are some things remaining TODO but before I do those things I want to ensure the basic idea is correct.

TODO:
  1) Tests
  2) Redoing event registration (so it uses existing SwG code instead of copied google3 code)
  3) Validation
  4) Others?  This is just an initial take on the concept / project.

### DIFF
--- a/src/runtime/propensity-server.js
+++ b/src/runtime/propensity-server.js
@@ -18,10 +18,12 @@ import {
   EventOriginator,
   EventParams,
 } from '../proto/api_messages';
+import {TC_STRING_UNAVAILABLE} from './gdpr-tcfv2-constants';
+import {TcfV2} from './gdpr-tcfv2';
 import {addQueryParam} from '../utils/url';
 import {adsUrl} from './services';
 import {analyticsEventToPublisherEvent} from './event-type-mapping';
-import {isBoolean, isObject} from '../utils/types';
+import {doArraysMatch, isBoolean, isObject} from '../utils/types';
 
 /**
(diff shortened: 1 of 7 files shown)


src/runtime/propensity-server.js
```diff
@@ -238,18 +239,71 @@ export class PropensityServer {
...
+      return;
+    }
+    this.vendors_ = vendors;
+    this.gdpr_ = new TcfV2(this.win_);
+    let resolver;
+    this.consentStringPromise_ = new Promise(
+      resolve => void (resolver = resolve)
+    );
+    const callback = TCData => {
+      if (TCData.tcString === TC_STRING_UNAVAILABLE) {
```
mborof: Note here that I am assuming "unavailable" is temporary and not permanent.  So it loops until it works.  If this is not a valid assumption let me know.

src/runtime/propensity-server.js
```diff
@@ -238,18 +239,71 @@ export class PropensityServer {
       method: 'GET',
       credentials: 'include',
     });
-    const url =
```
mborof: Should getPropensity also go through GDPR?

mborof:
Note that I had to create a new PR due to a reinstall of my operating system.  The latest code is here:

https://github.com/subscriptions-project/swg-js/pull/1002

PR #1002, opened 2020-02-28
Propensity gdpr

mborof:
Not Ready for Review

### DIFF
--- a/src/runtime/gdpr-tcfv2.js
+++ b/src/runtime/gdpr-tcfv2.js
@@ -1,0 +1,427 @@
+/**
+ * Copyright 2020 The Subscribe with Google Authors. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS-IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {TC_STRING_UNAVAILABLE} from './gdpr-tcfv2-constants.js';
+import {isFunction} from '../utils/types.js';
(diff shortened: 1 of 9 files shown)


mborof:
A note that I had to create a new PR due to reinstalling windows.  This cleans up the initial PR (999) a bit.

DO NOT SUBMIT - this is not ready yet.

This is my initial take on GDPR support in SwG. There are some things remaining TODO but before I do those things I want to ensure the basic idea is correct.

TODO:

Tests
Redoing event registration (so it uses existing SwG code instead of copied google3 code)
Validation
Others? This is just an initial take on the concept / project.

src/runtime/gdpr-tcfv2.js
```diff
@@ -0,0 +1,427 @@
...
+  }
+
+  /**
+   * Invokes the callback with whatever is available in the current state of the
+   * CMP. If CMP is on the same domain, it is expected to be synchronous.
+   * @param {function(!./gdpr-tcfv2-constants.TCData): undefined} clientCallback
+   * @param {!Array<string>|undefined} vendorIds Which vendors to lookup, or
+   *     undefined for all vendors.
+   */
+  getTCData(clientCallback, vendorIds) {
```
mborof: I modified this function, I am considered about the possibility of an infinitely loop.  Is TC_STRING_UNAVAILABLE something you just wait until its is ready for or is it a permanent state?

src/runtime/propensity-server.js
```diff
@@ -175,6 +176,53 @@ export class PropensityServer {
...
   }
 
+  /**
+   * @param {string} url
+   * @private
+   */
+  sendBeacon_(url) {
+    this.setupGDPR_();
+    this.tcDataPromise_.then(tcdata => {
+      if (tcdata) {
```
mborof: If TCData never becomes available (and GDPR is setup) then the promise result will never get called.  Is this OK?

src/runtime/propensity-server.js
```diff
@@ -175,6 +176,53 @@ export class PropensityServer {
     );
   }
 
+  /**
+   * @param {string} url
+   * @private
+   */
+  sendBeacon_(url) {
+    this.setupGDPR_();
+    this.tcDataPromise_.then(tcdata => {
```
mborof: TCData, once established - is it static or do I need to check again?

Issue #1016, opened 2020-03-13
Better error handling


jpettitt:
Right now swg.js tends to just `throw new Error(...)` when something fails.  This is causing a problem with amp integration as it doesn't fit the AMP error handling model.  [AMP issue 27223](https://github.com/ampproject/amphtml/issues/27223).

We should use a standard error handler that can be replaced with the AMP one in the similarly to the way we handle `fetch`

rcebulko:
If you'd like, I could port a simplified version `user().error(args)`, `user().expectedError(args)`, `dev().error(args)`, and `dev().expectedError(args)` into SwG that just adds the sentinel values/`expected` property and throws

jpettitt:
That would be great - I'd rather not have the sentinels unless it's in an amp context so maybe detect the AMP global before adding them?

rcebulko:
I can't request your review, but the above PR adds an error logger interface

PR #1017, opened 2020-03-13
Introduce amphtml-style error logging conventions/helpers

rcebulko:
Addresses https://github.com/subscriptions-project/swg-js/issues/1016

Introduces a logging convention similar to, and compatible with, ampproject/amphtml. Also updates one user error to use the error logger. The module is used like this:

```javascript
import {user, dev} from '../utils/error-logger';

user().expectedError('Illegal configuration option');
try {
  doWebRequest();
} catch (e) {
  dev().error(e);
}
```

### DIFF
--- a/src/model/page-config.js
+++ b/src/model/page-config.js
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+import {user} from '../utils/error-logger';
+
 /**
  */
 export class PageConfig {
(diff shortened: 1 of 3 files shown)


rcebulko:
@jpettitt PTAL, let me know if you think a different calling convention would be more useful in this repo, for now I made it mirror amphtml fairly closely

jpettitt:
The amp convention LGTM

PR #1040, opened 2020-04-03
Update random.js for ie11 compatibility

elijahsoria:
Changes self.crypto to self.msCrypto when in IE.


ChrisAntaki:
Verified the method broke before in IE11 and is fixed now 👍

PR #1092, opened 2020-05-12
Pay complete flow now provides one-time status

Kestrad:
Updated pay completion flow to forward information about whether a contribution was one-time or not to the payConfirmIframe. Added a test to check this new behavior, and modified old tests to accommodate for the new response.

### DIFF
--- a/src/runtime/pay-flow.js
+++ b/src/runtime/pay-flow.js
@@ -269,10 +269,12 @@ export class PayCompleteFlow {
     );
     this.deps_.entitlementsManager().reset(true);
     this.response_ = response;
+    // TODO(dianajing): future-proof isOneTime flag
     const args = {
       'publicationId': this.deps_.pageConfig().getPublicationId(),
       'productType': this.response_['productType'],
       'isSubscriptionUpdate': !!this.response_['oldSku'],
+      'isOneTime': !!this.response_['paymentRecurrence'],
     };
     // TODO(dvoytenko, #400): cleanup once entitlements is launched everywhere.
     if (response.userData && response.entitlements) {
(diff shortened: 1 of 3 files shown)


src/runtime/pay-flow.js
```diff
@@ -269,10 +269,12 @@ export class PayCompleteFlow {
     );
     this.deps_.entitlementsManager().reset(true);
     this.response_ = response;
+    // TODO(dianajing): future-proof isOneTime flag
     const args = {
       'publicationId': this.deps_.pageConfig().getPublicationId(),
       'productType': this.response_['productType'],
       'isSubscriptionUpdate': !!this.response_['oldSku'],
+      'isOneTime': !!this.response_['paymentRecurrence'],
```
alin04, line 277: Do you recall the reason for mapping oneTime-> paymentRecurrence in PayStartFlow and then mapping it back from paymentRecurrence -> oneTime in PayCompleteFlow ?

Also minor nit to keep the json fields the same. It's 'oneTime' in PayStartFlow, but 'isOneTime' here.

Kestrad: I believe we had to change oneTime to paymentRecurrence specifically because the Pay side needed the one-time enum. Currently we're only even passing the information in case the contribution is a one-time contribution, so converting it back to a boolean this way is expedient, but not very future-proof.

I can change 'isOneTime' to 'oneTime' though I was trying to be consistent with 'isSubscriptionUpdate.'

alin04: Ah, that's right. paymentRecurrence flows all the way through to buy flow into backend.

PR #1125, opened 2020-06-04
♻️ Uses the term allowlist instead of whitelist

ChrisAntaki:
Also includes some Prettier formatting changes that VS Code automatically applied


kristoferbaxter:
Nice work!

PR #1154, opened 2020-06-29
♻  Moves Payments Client Initialization into start() from ctor

elijahsoria:
-Inside of PayClient, moves PaymentsAscynClient initialization into start method instead of initialization upon PayClient ctor.
-Removes payments prefetch call from Runtime and puts it in PayClient.start().
-Fixes and Adds tests.

### DIFF
--- a/src/runtime/runtime.js
+++ b/src/runtime/runtime.js
@@ -578,13 +578,12 @@ export class ConfiguredRuntime {
     const preconnect = new Preconnect(this.win_.document);
 
     preconnect.prefetch('$assets$/loader.svg');
+    preconnect.preconnect('https://www.gstatic.com/');
+    preconnect.preconnect('https://fonts.googleapis.com/');
+    preconnect.preconnect('https://www.google.com/');
     LinkCompleteFlow.configurePending(this);
     PayCompleteFlow.configurePending(this);
-    if (isExperimentOn(this.win_, ExperimentFlags.PAY_CLIENT_LAZYLOAD)) {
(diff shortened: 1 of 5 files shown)


src/runtime/runtime.js
```diff
@@ -580,7 +580,6 @@ export class ConfiguredRuntime {
     preconnect.prefetch('$assets$/loader.svg');
     LinkCompleteFlow.configurePending(this);
     PayCompleteFlow.configurePending(this);
-    this.payClient_.preconnect(preconnect);
```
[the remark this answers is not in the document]
elijahsoria: I think preconnecting is safe but wary of putting any prefetch code since this issue was about cookies being dropped by pay. Adding in preconnecting within Runtime directly.

jpettitt:
Will this work in the redirect case - specifically where user is redirected out of a webview to a browser so the return from redirect is in a different local storage / cookie context and relies on the url fragment?

alin04:
As discussed separately, let's protect this behind an experiment...

src/runtime/runtime.js
```diff
@@ -580,7 +580,13 @@ export class ConfiguredRuntime {
...
+    if (isExperimentOn(this.win_, ExperimentFlags.PAY_CLIENT_LAZYLOAD)) {
```
alin04, line 583: These 3 preconnects should always happen regardless of pay. They are independent/unrelated.

elijahsoria: Good point. I'll just move these preconnects directly into Runtime now.

PR #1160, opened 2020-07-07
Support for launching publisher login flows from Google mobile apps.

kprevas:


src/api/subscriptions.js
```diff
@@ -271,6 +272,40 @@ export class Subscriptions {
...
   getPropensityModule() {}
 
+  /**
+   * Returns true if the Google app login flow is currently
+   * active, i.e., if googleAppLoginFlowComplete should be
+   * called when the user logs in.
+   *
+   * @return {boolean}
+   */
+  isGoogleAppLoginFlowActive() {}
```
jpettitt, line 282: Nit: active or available?

src/runtime/app-login-manager.js
```diff
@@ -0,0 +1,113 @@
...
+ */
+
+import {tryParseJson} from "../utils/json";
+import {UserState} from '../model/user-state';
+
+const LOGIN_REDIRECT_PARAM = 'googleLoginRedirect';
+
+/**
+ */
+export class AppLoginManager {
```
jpettitt, line 24: Can we add some logging to this to help with debugging?

src/api/subscriptions.js
kprevas: "Active", in that this is the login page running in a CCT that was launched by the app login flow.

src/runtime/app-login-manager.js
kprevas: Done.

ChrisAntaki:
Thanks for your formative work on the GSI flows in SwG, @kprevas !

PR #1164, opened 2020-07-09
Expose event manager to publishers

### LINKED ISSUE #1163: [Feature Request] Allow publishers to register an event listener using the `subscriptions` object
Using the existing `registerEventListener` or `registerEventFilterer` methods, publishers should be able to register their own event listener to track SwG events.

For example:
```js
subscriptions.registerEventListener(function (event) { 
  // Send event to publisher server
});
```

I think exposing the entire `eventManager` object might be too much as that could be used to push new events instead of just registering the listener.

jpettitt:
Exposes event manager to publisher to allow them to see all SwG analytics events.  Fixes #1163

### DIFF
--- a/src/api/subscriptions.js
+++ b/src/api/subscriptions.js
@@ -23,7 +23,6 @@ import {
 import {SubscribeResponse} from './subscribe-response';
 import {PropensityApi} from './propensity-api';
 import {LoggerApi} from './logger-api';
-import {ClientEventManager} from './client-event-manager'
 
 /**
  * @interface
--- a/src/api/subscriptions.js
+++ b/src/api/subscriptions.js
@@ -23,6 +23,7 @@ import {
 import {SubscribeResponse} from './subscribe-response';
 import {PropensityApi} from './propensity-api';
 import {LoggerApi} from './logger-api';
+import {ClientEventManager} from './client-event-manager'
 
 /**
  * @interface
@@ -273,6 +274,9 @@ export class Subscriptions {
 
   /** @return {!Promise<LoggerApi>} */
   getLogger() {}
+
+  /** @return {!ClientEventManager} } */
+  getEventManager() {}
 }
 
 /** @enum {string} */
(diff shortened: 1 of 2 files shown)


src/api/subscriptions.js
```diff
@@ -273,6 +273,9 @@ export class Subscriptions {
 
   /** @return {!Promise<LoggerApi>} */
   getLogger() {}
+
+  /** @return {!ClientEventManager} } */
+  getEventManager() {}
```
mborof, line 278: Should we be concerned about them trying to become filterers and squashing out logging?

mborof: We had initially created logger as their version of event manager (it was only working in propensity events granted).  We could extend listeners through that if we are concerned about filterers.  Not sure we care.  Other than that this seems fine, as long as they have access to AnalyticsEvent already also

src/runtime/runtime.js
```diff
@@ -474,6 +474,11 @@ export class Runtime {
   getLogger() {
     return this.configured_(true).then((runtime) => runtime.getLogger());
   }
+
+  /** @override */
+  eventManager() {
+    return this.configured_(true).then((runtime) => runtime.eventManager());
```
mborof, line 480: Isn't this returning a promise to an event manager?

mborof: It's kind of odd to change the name also isn't it?

src/api/subscriptions.js
jpettitt: It's their data and given the limitless ways publisher can screw up the data filtering is the least of our worries.

src/runtime/runtime.js
jpettitt: Not sure what you mean by changing the name?

mborof: getEventManager vs eventManager

mborof: There is also an existing eventManager function in ConfiguredRuntime that returns event manager directly.

jpettitt: ah, yes I named them differently because eventManager returns the event manager directly but getEventManager returns a Promise.

mborof: So what is this new function overriding then?  Should it be called getEventManager and override that?  Sorry if I'm slightly confused

PR #1166, opened 2020-07-10
Entitlements Pingback Framework

mborof:
Sends a metered entitlement to the entitlements pingback endpoint.


src/proto/api_messages.js
```diff
@@ -72,6 +72,9 @@ const AnalyticsEvent = {
   ACTION_SWG_BUTTON_SELECT_OFFER_CLICK: 1018,
   ACTION_SWG_BUTTON_SHOW_CONTRIBUTIONS_CLICK: 1019,
   ACTION_SWG_BUTTON_SELECT_CONTRIBUTION_CLICK: 1020,
+  ACTION_USER_CONSENT_DEFERRED_ACCOUNT: 1021,
```
ChrisAntaki, line 75: This file seems to have a lot of unrelated changes. Could you make a separate PR to regenerate this file without the metering changes?

src/utils/date-utils.js
```diff
@@ -0,0 +1,26 @@
...
+
+import {Timestamp} from "../proto/api_messages";
+
+/**
+ * @param {!Date} date
+ * @return {!Timestamp}
+ */
+export function toTimestamp(date) {
+    const millis = date.getTime();
+    return new Timestamp([millis / 1000, Math.floor ((millis % 1000) * 1000000)]);
```
ChrisAntaki, line 25: Could you add a test for this file? I think the first parameter of Timestamp needs to be an integer, and it'd be good to have that checked in an automated test

src/runtime/fetcher.js
```diff
@@ -64,28 +72,30 @@ export class XhrFetcher {
     return this.fetch(url, init).then((response) => response.json());
   }
 
+  sendMessage(url, message) {
```
ChrisAntaki, line 75: Does this need `@override`?

src/runtime/fetcher.js
```diff
@@ -40,6 +40,14 @@ export class Fetcher {
...
    */
   sendBeacon(unusedUrl, unusedData) {}
+
+  /**
+   * POST data to a URL endpoint, get a Promise for a response
+   * @param {!string} unusedUrl
+   * @param {!string|!Object} unusedMessage
+   * @return {!Promise<!../utils/xhr.FetchResponse>}
+   */
+  sendMessage(unusedUrl, unusedMessage) {}
```
ChrisAntaki, line 50: I think `sendPost` would be a more descriptive name, what do you think? Also, would it be possible to add a test for this and create a new PR for it?

src/runtime/entitlements-manager.js
```diff
@@ -356,30 +371,59 @@ export class EntitlementsManager {
...
+  markMeterAsUsed(signedMeter) {
+    if (isExperimentOn(this.win_, ExperimentFlags.METERING)) {
+      const url =
+        '/publication/' +
+        encodeURIComponent(this.publicationId_) +
+        '/entitlements';
+      const message = new EntitlementsPingbackRequest();
+      message.setSignedMeter(signedMeter);
+      message.setClientEventTime(toTimestamp(new Date()));
+      message.setHashedCanonicalUrl(this.url_);
```
ChrisAntaki, line 405: For now, I would suggest just sending a dummy string for the hashed URL and publisher user ID. We don't know where these values will come from yet, but I'm working on designing a solution :)

src/runtime/fetcher.js
mborof: I'm only just getting to the tests.  I can break things up a bit into separate PRs though

ChrisAntaki: Sweet!

PR #1170, opened 2020-07-14
Add public non-beacon post

mborof:


src/runtime/fetcher.js
```diff
@@ -40,6 +40,14 @@ export class Fetcher {
...
    */
   sendBeacon(unusedUrl, unusedData) {}
+
+  /**
+   * POST data to a URL endpoint, get a Promise for a response
+   * @param {!string} unusedUrl
+   * @param {!string|!Object} unusedMessage
+   * @return {!Promise<!../utils/xhr.FetchResponse>}
+   */
+  postMessage(unusedUrl, unusedMessage) {}
```
ChrisAntaki, line 50: How about the name `sendPost`? I like `postMessage`, but it's already being used by an important JS API https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage

mborof: Sure

PR #1171, opened 2020-07-14
Create toTimestamp

mborof:


src/utils/date-utils-test.js
```diff
@@ -0,0 +1,43 @@
...
+ * @param {*} val
+ * @return {boolean}
+ */
+function isInteger(val) {
+  return Math.floor(val) === Math.ceil(val);
+}
+
+describe('toTimestamp', () => {
+  it('should create a timestamp', () => {
+    const date = new Date();
```
ChrisAntaki, line 30: To make this test more readable, could you use a consistent seed for the Date? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date

This would allow for simple values in the expectations instead of needing to repeat equations in the test

ChrisAntaki:
Thanks for making this small PR, so easy to review

src/utils/date-utils-test.js
mborof: sure

ChrisAntaki, line 30:
```suggestion
    const date = new Date(819199441032);
```

You can pass the timestamp directly to the `Date` constructor

src/utils/date-utils-test.js
```diff
@@ -0,0 +1,43 @@
...
+ * @return {boolean}
+ */
+function isInteger(val) {
+  return Math.floor(val) === Math.ceil(val);
+}
+
+describe('toTimestamp', () => {
+  it('should create a timestamp', () => {
+    const date = new Date(819199441032);
+    date.setTime(819199441032);
```
mborof, line 31:
```suggestion
```

PR #1174, opened 2020-07-17
Move PayFrameHelper.load() to PaymentsAsyncClient's ctor

elijahsoria:
PayFrameHelper.load() is being called before the pay client is instantiated, which means that it drops cookies before pay is in use. Moving it to the ctor of PaymentsAsyncClient should solve the issue.

The problem is that we can't test this change fully since we can't use experiments and Pay doesn't work for local environments, so we will have to test this change in Qual before creating a prod swg-js release.


elijahsoria:
Create an issue to update third_party/gpay and get the sync between the main codebase and swg-js to work again: https://github.com/subscriptions-project/swg-js/issues/1177

PR #1186, opened 2020-07-28
Removes Pay Client Experiment Guards

elijahsoria:
Moves the code from the PAY_CLIENT_LAZYLOAD experiment into production. Removes unnecessary tests. This is essentially putting https://github.com/subscriptions-project/swg-js/pull/1154 into production, meaning that payments prefetch calls will now happen upon start instead of in PayClient's constructor.


jpettitt:
Let's hold off on this, set the experiment to 100% in the deploy code and after it's been there for a week without breaking anything then remove the experiment code.

ChrisAntaki (approved):

PR #1193, opened 2020-07-31
Upgrade Closure Compiler so it runs in Java 11

mborof:
This gets the compiler to basically run but does not fix some syntax errors.


mborof:
The actual compiler is now being invoked successfully but something has changed in referencing when doing imports.  The system can't find imports that are not under the "src" directory

PR #1195, opened 2020-08-01
SwG Builds in Java 11

mborof:
This PR fixes our compiler so it can run in Java 11. Unfortunately, there is some kind of difference between a local build and a travis build. This fix passes Travis. I will follow up on how to fix the local build system.


mborof:
This PR fixes our compiler so it can run in Java 11.  Unfortunately, there is some kind of difference between a local build and a travis build.  This fix passes Travis.  I will follow up on how to fix the local build system.

package.json
```diff
@@ -56,6 +56,7 @@
     "fancy-log": "1.3.3",
     "fs-extra": "9.0.1",
     "glob": "7.1.6",
+    "google-closure-compiler": "^20200719.0.0",
     "gulp": "4.0.2",
     "gulp-closure-compiler": "0.4.0",
```
qidonna, line 61: remove this dep?

PR #1200, opened 2020-08-03
Create a hashing utility

mborof:


src/utils/string.js
```diff
@@ -165,3 +166,37 @@ export function getUuid() {
...
+  for (let i = 0; i < view.byteLength; i += 4) {
+    // toString(16) will give the hex representation of the number without padding
+    const stringValue = view.getUint32(i).toString(16);
+    hexCodes.push(padString(stringValue, PADDING));
+  }
+  return hexCodes.join('');
+}
+
+/**
+ * @param {string} stringToHash
```
ChrisAntaki, line 193: Please add a comment for this method

src/utils/string.js
```diff
@@ -165,3 +166,37 @@ export function getUuid() {
 export function getSwgTransactionId() {
   return getUuid() + '.swg';
 }
+
+/**
+ * Returns a string who's length matches the length of format.
```
ChrisAntaki, line 171: Please update "who's" to "whose"

src/utils/string-test.js
```diff
@@ -162,3 +163,19 @@ describe('swgTransactionId', () => {
     }
   });
 });
+
+describe('hash', async () => {
+  it('should create 128 character hexadecimal hashes', async () => {
+    let result = null;
```
ChrisAntaki, line 169: Please combine this line and the line below it. Ex: `const result = await ...`

src/utils/string-test.js
```diff
@@ -162,3 +163,19 @@ describe('swgTransactionId', () => {
...
+    await hash('string1').then((h) => (result = h));
+    expect(result.length).to.equal(128);
+  });
+
+  it('should create unique hashes', async () => {
+    let hash1 = null;
+    let hash2 = null;
```
ChrisAntaki, line 176: Please use the values of the `await` expressions to define these variables more succinctly. Ex: `const hash1 = await ...`

src/utils/string-test.js
```diff
@@ -162,3 +163,19 @@ describe('swgTransactionId', () => {
     }
   });
 });
+
+describe('hash', async () => {
+  it('should create 128 character hexadecimal hashes', async () => {
+    let result = null;
+    await hash('string1').then((h) => (result = h));
+    expect(result.length).to.equal(128);
```
ChrisAntaki, line 171: Please check against an actual hash value instead of length

src/utils/string.js
mborof: To be honest, I'm not sure what a good comment is for this function.

ChrisAntaki: Could you describe what it does, and what if might be used for?

src/utils/string-test.js
mborof: I'm not sure I agree in this case, we can discuss over chat if you'd like.  I did improve the tests some.

PR #1205, opened 2020-08-10
DRAFT Metering

ChrisAntaki:


examples/sample-pub/publisher.js
```diff
@@ -144,6 +144,36 @@ function startFlowAuto() {
...
+      subsciptions
+        .getEntitlements({
+          metering: {
+            userState: {
+              id:
+                'user5901e3f7a7fc5767b6acbbbaa927d36f5901e3f7a7fc5767b6acbbbaa927',
+              attributes: [
+                {
+                  name: 'registered_user',
+                  timestamp: (Date.now() / 1000) | 0,
```
jpettitt, line 158: What's with the  `| 0`

examples/sample-pub/publisher.js
```diff
@@ -144,6 +144,36 @@ function startFlowAuto() {
...
+              ],
+            },
+          },
+        })
+        .then((entitlements) => {
+          if (entitlements.enablesThisWithGoogleMetering()) {
+            console.log('Hey Google, thanks for the metering entitlement.');
+          } else {
+            console.log(
+              'I was promised metering entitlements... (╯°□°)╯︵ ┻━┻'
```
jpettitt, line 169: Change to "Sorry no meter entitlements today"

jpettitt:
Looks good so far. LMK when you've done pingback and I'll add the amp-subscriptions-google support.

examples/sample-pub/publisher.js
ChrisAntaki: It's a bit faster than using `Math.floor`. I updated the code to use `Math.floor` instead for readability. This will also help the example be more future proof, since the bitwise OR 0 trick only works up to values of 2^31−1. :hourglass_flowing_sand:

examples/sample-pub/publisher.js
ChrisAntaki: The message has been updated

ChrisAntaki:
Sounds good, I'll borrow from Matt's PR #1166 for the pingback

examples/sample-pub/publisher.js
```diff
@@ -144,6 +144,41 @@ function startFlowAuto() {
...
+      subscriptions
+        .getEntitlements({
+          metering: {
+            state: {
+              id:
+                'user5901e3f7a7fc5767b6acbbbaa927d36f5901e3f7a7fc5767b6acbbbaa927',
+              standardAttributes: [
+                {
+                  name: 'registered_user', // standard_registered_user
+                  timestamp: Math.floor(Date.now() / 1000),
```
jpettitt, line 158: This time must not change (unless the date the user registered changed :-).  Using `Date.now()` makes it appear that user data has changed on every page.   The timestamp is the time the attribute last changed.

examples/sample-pub/publisher.js
```diff
@@ -144,6 +144,41 @@ function startFlowAuto() {
...
+                },
+              ],
+              customAttributes: [
+                {
+                  name: 'newsletter_subscriber', // custom_newsletter_subscriber
+                  timestamp: Math.floor(Date.now() / 1000),
```
jpettitt, line 164: As above

src/runtime/entitlements-manager.js
```diff
@@ -108,12 +119,25 @@ export class EntitlementsManager {
...
    * @return {!Promise<!Entitlements>}
    */
-  getEntitlements(encryptedDocumentKey) {
+  getEntitlements(params) {
+    // Remain backwards compatible by accepting
+    // `encryptedDocumentKey` string as a first param.
+    if (typeof params === 'string') {
+      // TODO: Delete the fallback if nobody needs it. Use a log to verify.
+      warn(
+        `[swg.js:getEntitlements]: The first param of getEntitlements() should be an object of type GetEntitlementsParams.`
```
jpettitt, line 131: ... if present ...

jpettitt: Maybe comment out the warn until we have the matching AMP code deployed so as not to confuse people..

src/runtime/entitlements-manager.js
```diff
@@ -356,75 +402,157 @@ export class EntitlementsManager {
...
+              encodeURIComponent(params.encryption.encryptedDocumentKey)
+          );
+          delete params.encryption;
+        }
+
+        // Add metering params.
+        const productId = this.pageConfig_.getProductId();
+        if (productId && params && params.metering && params.metering.state) {
+          // Populate fields.
+          params.metering.clientTypes = [1];
```
jpettitt, line 501: ?  ENUM?

src/runtime/entitlements-manager.js
```diff
@@ -356,75 +402,157 @@ export class EntitlementsManager {
...
-      encodeURIComponent(this.publicationId_) +
-      '/entitlements';
-    if (encryptedDocumentKey) {
-      url += '?crypt=' + encodeURIComponent(encryptedDocumentKey);
-    }
-    return this.fetcher_
-      .fetchCredentialedJson(serviceUrl(url))
-      .then((json) => this.parseEntitlements(json));
+  fetch_(params) {
+    return hash(getCanonicalUrl(this.deps_.doc()))
```
jpettitt, line 484: Better to build a new params object than mess the the one passed - a) the caller might be surprised their object got trashed and b) they may have crap in there we don't care about.  If you make a new object you know exactly what's in it.

ChrisAntaki: Good call, the code is now creating a new object

src/runtime/entitlements-manager.js
ChrisAntaki: I created a new enum (`MeterClientTypes`) with a `LICENSED_BY_GOOGLE` (`1`) value

src/runtime/entitlements-manager.js
ChrisAntaki: I added a `Date.now()` which will block this warning for 2 weeks

examples/sample-pub/publisher.js
ChrisAntaki: I added a variable that defines a specific time.
`const timestamp = 1597686771`

examples/sample-pub/publisher.js
```diff
@@ -144,6 +144,43 @@ function startFlowAuto() {
...
+                  timestamp,
+                },
+              },
+            },
+          },
+        })
+        .then((entitlements) => {
+          if (entitlements.enablesThisWithGoogleMetering()) {
+            console.log(
+              'Consuming Google metering entitlement from Publisher JS'
```
chenshay, line 174: Can you please add more detail / comments / links to docs / etc? If I'm a developer on the publisher's side I won't understand what this does and why I want to put it in my code.

src/api/entitlements.js
```diff
@@ -82,6 +90,26 @@ export class Entitlements {
     };
   }
 
+  /**
+   * Returns true if the entitlements enable the current product
+   * with a cacheable entitlement.
```
chenshay, line 95: what's a cacheable entitlement? what does it mean to enable the current product with a cacheable entitlement?

src/api/entitlements.js
```diff
@@ -82,6 +90,26 @@ export class Entitlements {
...
+   * @return {boolean}
+   */
+  enablesThisWithCacheableEntitlements() {
+    const entitlement = this.getEntitlementForThis();
+    return !!entitlement && entitlement.source !== GOOGLE_METERING_SOURCE;
+  }
+
+  /**
+   * Returns true if the entitlements enable the current product
+   * with a Google metering entitlement.
```
chenshay, line 105: ditto - comment is not clear

src/api/entitlements.js
```diff
@@ -133,21 +161,46 @@ export class Entitlements {
...
-          (!source || source == this.entitlements[i].source)
-        ) {
-          return this.entitlements[i];
-        }
+    if (!product) {
+      return null;
+    }
+
+    // Sort non-metering entitlements first, so they are returned if possible,
+    // to avoid using metered reads unnecessarily.
```
chenshay, line 178: ditto. why do they need to be sorted? is there a thing that will go through them in a specific order? what is that thing?

src/api/entitlements.js
```diff
@@ -133,21 +161,46 @@ export class Entitlements {
...
+        return -1;
+      }
+
+      return 0;
+    });
+
+    for (let i = 0; i < entitlements.length; i++) {
+      if (
+        entitlements[i].enables(product) &&
+        (!source || source == entitlements[i].source)
```
chenshay, line 198: add comment to explain what you're checking for here

src/api/entitlements.js
```diff
@@ -178,6 +231,13 @@ export class Entitlements {
   ack() {
     this.ackHandler_(this);
   }
+
+  /**
+   * A 3p site should call this method to consume a Google metering entitlement.
```
chenshay, line 236: still not sure what it means to consume a GME

src/api/subscriptions.js
```diff
@@ -60,10 +60,10 @@ export class Subscriptions {
   clear() {}
 
   /**
-   * @param {?string=} encryptedDocumentKey
+   * @param {!GetEntitlementsParamsExternal=} params
    * @return {!Promise<!Entitlements>}
    */
-  getEntitlements(encryptedDocumentKey) {}
+  getEntitlements(params) {}
```
chenshay, line 66: Did you check that changing these params won't break anything? Check w/ @elijahsoria  to make sure

examples/sample-pub/publisher.js
```diff
@@ -144,6 +144,43 @@ function startFlowAuto() {
...
+    whenReady((subscriptions) => {
+      subscriptions
+        .getEntitlements({
+          metering: {
+            state: {
+              id:
+                'user5901e3f7a7fc5767b6acbbbaa927d36f5901e3f7a7fc5767b6acbbbaa927',
+              standardAttributes: {
+                registered_user: {
+                  timestamp,
```
mborof, line 160: Could we provide some explanation of this structure or an object typedef somewhere?  Or just declare it explicitly so someone can more easily look it up.

src/api/subscriptions.js
```diff
@@ -317,6 +317,72 @@ export const SubscriptionFlows = {
  */
 export let Config;
 
+/**
+ * Internal params for GetEntitlements requests.
```
chenshay, line 321: What does this mean? How will it be used. More details in the comments please

src/runtime/entitlements-manager.js
```diff
@@ -108,12 +124,29 @@ export class EntitlementsManager {
   }
 
   /**
-   * @param {?string=} encryptedDocumentKey
+   * @param {!GetEntitlementsParamsExternal=} params
```
chenshay, line 127: ditto. Check w/ @elijahsoria that this won't break anything

src/runtime/entitlements-manager.js
```diff
@@ -142,33 +175,54 @@ export class EntitlementsManager {
...
    * @private
    */
-  getEntitlementsFlow_(encryptedDocumentKey) {
-    return this.fetchEntitlementsWithCaching_(encryptedDocumentKey).then(
-      (entitlements) => {
-        this.onEntitlementsFetched_(entitlements);
-        return entitlements;
-      }
-    );
+  getEntitlementsFlow_(params) {
```
chenshay, line 204: ditto

src/runtime/entitlements-manager.js
```diff
@@ -142,33 +175,54 @@ export class EntitlementsManager {
...
   }
 
   /**
-   * @param {?string=} encryptedDocumentKey
+   * @param {!GetEntitlementsParamsExternal=} params
    * @return {!Promise<!Entitlements>}
    * @private
    */
-  fetchEntitlementsWithCaching_(encryptedDocumentKey) {
+  fetchEntitlementsWithCaching_(params) {
```
chenshay, line 216: ditto

src/runtime/entitlements-manager.js
```diff
@@ -356,75 +411,159 @@ export class EntitlementsManager {
...
-    if (!entitlement) {
-      return Promise.resolve();
-    }
+  maybeShowToast_(entitlement) {
     // Check if storage bit is set. It's only set by the `Entitlements.ack`
     // method.
     return this.storage_.get(TOAST_STORAGE_KEY).then((value) => {
-      if (value == '1') {
-        // Already shown;
+      const toastWasShown = value === '1';
```
chenshay, line 431: nice fix!

src/runtime/entitlements-manager.js
```diff
@@ -356,75 +411,159 @@ export class EntitlementsManager {
...
+      activityIframeView_.onCancel(() => {
+        this.sendPingback_(entitlements);
+      });
+      return this.deps_.dialogManager().openView(activityIframeView_);
     }
   }
 
   /**
-   * @param {?string=} encryptedDocumentKey
+   * @param {!GetEntitlementsParamsExternal=} params
```
chenshay, line 488: ditto

src/runtime/entitlements-manager.js
```diff
@@ -356,75 +411,159 @@ export class EntitlementsManager {
...
-  ack_(entitlements) {
-    if (entitlements.getEntitlementForThis()) {
-      this.setToastShown(true);
+  consume_(entitlements) {
+    if (entitlements.enablesThisWithGoogleMetering()) {
+      // TODO: Open Metering Prompt.
+      const activityIframeView_ = new ActivityIframeView(
+        this.win_,
+        this.deps_.activities(),
+        feUrl('/contributionsiframe'),
```
chenshay, line 469: contributionsiframe for consume + google metering?
or is this a placeholder for something?
I don't get it

src/runtime/runtime.js
```diff
@@ -276,9 +276,9 @@ export class Runtime {
   }
 
   /** @override */
-  getEntitlements(encryptedDocumentKey) {
+  getEntitlements(params) {
```
chenshay, line 279: ditto

src/runtime/entitlements-manager.js
```diff
@@ -108,12 +124,29 @@ export class EntitlementsManager {
   }
 
   /**
-   * @param {?string=} encryptedDocumentKey
+   * @param {!GetEntitlementsParamsExternal=} params
    * @return {!Promise<!Entitlements>}
    */
-  getEntitlements(encryptedDocumentKey) {
+  getEntitlements(params) {
```
elijahsoria, line 130: The backend expects a encryptedDocumentKey and a param at the moment, so it would be good to put it separately for now otherwise I think it will break encryption items.

src/runtime/runtime.js
elijahsoria: Same as above. I think for backwards compatibility we need to have two arguments here, where the publisher can pass in the encryptedDocumentKey directly or as a part of the params.

src/api/subscriptions.js
```diff
@@ -317,6 +317,72 @@ export const SubscriptionFlows = {
...
+ * }}
+ */
+export let GetEntitlementsEncryptionParams;
+
+/**
+ * Internal metering params for GetEntitlements requests.
+ * @typedef {{
+ *   clientTypes: !Array<number>,
+ *   owner: string,
+ *   state: {
```
mborof, line 342: Out of date?

src/api/subscriptions.js
```diff
@@ -317,6 +317,72 @@ export const SubscriptionFlows = {
...
+ *     customAttributes: !Object<string, {
+ *       timestamp: number,
+ *     }>,
+ *   },
+ *   resource: {
+ *     hashedCanonicalUrl: string,
+ *   },
+ * }}
+ */
+export let GetEntitlementsMeteringParamsExternal;
```
mborof, line 384: I'd rather avoid having the word "External" in a public facing typedef.  From their perspective it is just GetEntitlementsMeteringParams.  Perhaps we can just have Internal for ours or come up with a better name so we don't need this descritpor?

src/runtime/entitlements-manager.js
ChrisAntaki: Don't worry, the JavaScript is still sending the encryption parameter to the backend just as it used to. You can search this file for "crypt=" to verify.

src/runtime/entitlements-manager.js
ChrisAntaki: Thanks!

src/runtime/entitlements-manager.js
ChrisAntaki: This is just a placeholder. I will add a comment mentioning this

src/api/subscriptions.js
ChrisAntaki: Yes, I designed these changes to be backwards compatible. Elijah is double checking as well

src/runtime/entitlements-manager.js
elijahsoria: Awesome work! Disregard my initial comment.

src/runtime/runtime.js
elijahsoria: Disregard! You addressed the concern already.

examples/sample-pub/publisher.js
ChrisAntaki: I added descriptions in `publisher.js` and `entitlements-flow.md`

src/api/subscriptions.js
ChrisAntaki: I would expect publishers would start by checking out the `publisher.js` example and/or a markdown file like `entitlements-flow.md`. Publishers won't need to reference this typedef in their code either. If they arrive at this typedef, it's because they were digging in the source code, at which point I think a term like `External` would be fairly descriptive. Please feel free to suggest clearer alternatives though?

examples/sample-pub/publisher.js
ChrisAntaki: For sure, I added comments in `publisher.js` and `entitlements-flow.md`

src/api/entitlements.js
ChrisAntaki: I updated this comment and added more details

src/api/entitlements.js
ChrisAntaki: I updated this comment and added more details

src/api/entitlements.js
ChrisAntaki: I rewrote this comment and refactored the code below so it's (hopefully) more descriptive

src/api/entitlements.js
ChrisAntaki: I refactored this code to (hopefully) be more descriptive

src/api/entitlements.js
ChrisAntaki: I expanded this comment

src/api/subscriptions.js
ChrisAntaki: I added details to these comments. Thanks for letting me know they needed more work

PR #1226, opened 2020-08-27
Adds `onDialogClose` callback to entitlements.consume() method, and improves Scenic metering demo

ChrisAntaki:


examples/sample-pub/metering.js
```diff
@@ -0,0 +1,41 @@
+/** Helps demonstrate metering functionality. */
+const MeteringDemo = {
+  /** Sets up controls for the metering demo. */
+  setupControls: () => {
+    // Wire up reset button.
+    document
+      .querySelector('#metering-controls button')
+      .addEventListener('click', () => {
+        // Forget the existing PPID.
```
elijahsoria, line 9: Could you spell out what PPID is in one of these comments?

src/api/entitlements.js
```diff
@@ -235,11 +235,13 @@ export class Entitlements {
 
   /**
    * A 3p site should call this method to consume a Google metering entitlement.
-   * When a metering entitlement is consumed, that depletes the user's
-   * remaining "free reads".
+   * When a metering entitlement is consumed, SwG shows the user a metering dialog.
+   * When the user closes the dialog, SwG depletes one of the user's remaining
+   * "free reads".
+   * @param {!Function=} onCloseDialog Called after the user closes the dialog.
```
elijahsoria, line 241: Can we let this be null? Do we expect something to be called after every dialog flow?

examples/sample-pub/metering.js
ChrisAntaki: Definitely, I spelled it out in a couple of comments below

src/api/entitlements.js
ChrisAntaki: Good call, the function is now optional

elijahsoria: Please update the comment above and elsewhere to be "?Function=" instead of"!Function="

elijahsoria (dismissed):

src/api/entitlements.js
ChrisAntaki: I think `!Function=` is ideal because it tells the Closure compiler "Either a Function or undefined are OK". Whereas `?Function=` would tell Closure "Either a null, or a function, or undefined are OK". In this case, I think just expecting a Function or undefined would make more sense

ChrisAntaki: Then again, allowing `null` wouldn't hurt. I'll update the code :)

PR #1238, opened 2020-09-01
Create Example Google Sign-in iframe

elijahsoria:
Creates an example Google sign-in iframe using One Tap that publishers will use to handle their google sign-in configuration.

https://developers.google.com/identity/gsi/web/reference/js-reference#google.accounts.id.renderButton


examples/google-signin/google-signin-iframe.html
```diff
@@ -0,0 +1,100 @@
...
+      }
+      domainVerifiedCallback = done;
+      window.addEventListener('message', (event) => {
+          if (event.source != window.parent || !event.data) {
+              return;
+          }
+          if (!pendingNonce || !domainVerifiedCallback) {
+              return;
+          }
+          if (event.data['sentinel'] != 'onetap_google' || event.data['command'] != 'parent_frame_ready') {
```
jpettitt, line 35: Let use a different sentinel value

examples/google-signin/google-signin-iframe.html
```diff
@@ -0,0 +1,100 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <script src="https://accounts.google.com/gsi/client"></script>
+  <script>
```
jpettitt, line 6: We'll want to move most the JS into a .js file we host so allow us to update it

examples/google-signin/google-signin-iframe.html
```diff
@@ -0,0 +1,100 @@
...
+          form.appendChild(input);
+        });
+      }
+      form.submit();
+    }
+
+    function handleLoginResponse(response) {
+      if (!response || !response['credential']) return;
+      responseReturned = true;
+      notifyParent(response);
```
jpettitt, line 80: need to call formPost here and get a user data back to pass to parent.

jpettitt: use fetch not form so we can get json back.

examples/google-signin/google-signin-iframe.html
elijahsoria: Used "google_signin" instead.

examples/google-signin/google-signin-iframe.html
elijahsoria: Modified the code to use fetch (with a POST) through the SwG XHR class.

examples/google-signin/google-signin-iframe.html
elijahsoria: Cool! I put a dummy file in there for now. I'll have to update it once it actually is ready.

examples/google-signin/google-signin-iframe.html
elijahsoria: Decided on leaving this up to the publisher (but recommended fetch).

src/utils/google-signin-utils.js
```diff
@@ -0,0 +1,127 @@
...
+/** Command name for when the parent frame is ready for iframe start. */
+export const PARENT_READY_COMMAND = 'parent_frame_ready';
+/** Command name for when the Google Sign-in iframe is ready. */
+export const INTERMEDIATE_IFRAME_READY_COMMAND = 'intermediate_iframe_ready';
+/** Command name for when the user's entitlements are returned from the publisher. */
+export const ENTITLEMENTS_READY_COMMAND = 'entitlements_ready';
+
+/**
+ * Modifies the input call back to chain parent notification of response.
+ * @param {!Window} win
```
ChrisAntaki, line 28: I noticed several methods accept a `win` param. Could they use the global `self` reference instead?

src/utils/google-signin-utils.js
```diff
@@ -0,0 +1,127 @@
...
+ * @param {!function()} callback
+ */
+export function createGoogleSigninCallback(win, callback) {
```
ChrisAntaki, line 31: Could you rename this function to "createGoogleSignInCallback"? I typically see the "i" from "in" capitalized, when people refer to Google Sign In (GSI)

src/utils/google-signin-utils.js
```diff
@@ -0,0 +1,127 @@
...
+  };
+}
+
+/**
+ * Sends a post message to the window's parent.
+ * @param {!Window} win
+ * @param {!Object} message
+ */
+export function notifyParent(win, message) {
+  win.parent && win.parent.postMessage(message, '*');
```
ChrisAntaki, line 48: Allowing any domain ("*") here might open a security vulnerability

src/utils/google-signin-utils.js
```diff
@@ -0,0 +1,127 @@
...
+
+  /**
+   * Registers an event listener that calls the signinCallback to display the Sign-in button after the event is verified.
+   */
+  registerDomainVerifier_() {
+    if (!this.win_.parent) {
+      return;
+    }
+    this.win_.addEventListener('message', (event) => {
+      if (event.source != this.win_.parent || !event.data) {
```
ChrisAntaki, line 88: Please check strict equality (`===` or `!==`) instead of loose equality (`==` or `!=`) in JavaScript, when possible
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness

ChrisAntaki: Please add comments that describe why each of these if statements are necessary

src/utils/google-signin-utils.js
```diff
@@ -0,0 +1,127 @@
...
+export const PARENT_READY_COMMAND = 'parent_frame_ready';
+/** Command name for when the Google Sign-in iframe is ready. */
+export const INTERMEDIATE_IFRAME_READY_COMMAND = 'intermediate_iframe_ready';
+/** Command name for when the user's entitlements are returned from the publisher. */
+export const ENTITLEMENTS_READY_COMMAND = 'entitlements_ready';
+
+/**
+ * Modifies the input call back to chain parent notification of response.
+ * @param {!Window} win
+ * @param {!function()} callback
```
ChrisAntaki, line 29: Could this param get a more descriptive type, name, and comment?

src/utils/google-signin-utils.js
```diff
@@ -0,0 +1,127 @@
...
+    });
+  };
+}
+
+/**
+ * Sends a post message to the window's parent.
+ * @param {!Window} win
+ * @param {!Object} message
+ */
+export function notifyParent(win, message) {
```
ChrisAntaki, line 47: Could the `createGoogleSigninCallback` and `notifyParent` methods be private methods of `SwgGoogleSigninCreator`? I noticed they're only used within that class

src/utils/google-signin-utils.js
```diff
@@ -0,0 +1,127 @@
...
+    this.registerDomainVerifier_();
+    this.requestDomainVerification_();
+  }
+
+  /**
+   * Registers an event listener that calls the signinCallback to display the Sign-in button after the event is verified.
+   */
+  registerDomainVerifier_() {
+    if (!this.win_.parent) {
+      return;
```
ChrisAntaki, line 85: Could you add tests that cover all the lines?

ChrisAntaki (changes requested):

src/utils/google-signin-utils.js
ChrisAntaki: I'm worried an attacker could open the iframe on their own website, and then get the publisher's response postMessage.

src/utils/google-signin-utils.js
ChrisAntaki: I just noticed `createGoogleSigninCallback` is used by the iframe. Could we simplify the API for publishers? Perhaps the `SwgGoogleSigninCreator` constructor could accept a `CLIENT_ID` and `publisherGoogleSignInCallback` directly

src/utils/google-signin-utils.js
ChrisAntaki: If the `registerDomainVerifier_` method takes care of this, please add a comment mentioning this

examples/google-signin/google-signin-iframe.html
```diff
@@ -0,0 +1,60 @@
...
+      const formData = new FormData();
+      Object.keys(response).map(function (name) {
+        formData.append(name, response[name]);
+      });
+      const init = /** @type {!./xhr.FetchInitDef} */ ({
+        method: 'POST',
+        credentials: 'include',
+        body: formData,
+      });
+      return window.fetch(LOGIN_ENDPOINT, init);
```
ChrisAntaki, line 39: IE11 doesn't support fetch

src/utils/google-signin-utils.js
ChrisAntaki: My guess would be the callback accepts a GSI response, and then returns a `Promise` containing a `Response`?

src/utils/google-signin-utils.js
```diff
@@ -0,0 +1,127 @@
...
+ * @param {!Window} win
+ * @param {!function()} callback
+ */
+export function createGoogleSigninCallback(win, callback) {
+  return (signinResponse) => {
+    const response = callback(signinResponse);
+    notifyParent(win, {
+      sentinel: SENTINEL,
+      command: ENTITLEMENTS_READY_COMMAND,
+      response: (response && JSON.parse(response)) || {},
```
ChrisAntaki, line 37: If the `response` is a `Promise` containing a `Response`, I don't think line 37 would correctly parse it
https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch

src/utils/google-signin-utils.js
elijahsoria: Not here since this function isn't a class method. I thought about adding it as a class method but realized that this function would have to be run before the class is instantiated.

src/utils/google-signin-utils.js
elijahsoria: Done.

src/utils/google-signin-utils.js
ChrisAntaki: The `self` variable is available globally, inside or outside of classes. It refers to the current `Window` https://developer.mozilla.org/en-US/docs/Web/API/Window/self

src/utils/google-signin-utils.js
elijahsoria: Added a comment about why it is safe (we verified the origin in registerDomainVerifier_ as you expected).

src/utils/google-signin-utils.js
elijahsoria: Done.

src/utils/google-signin-utils.js
elijahsoria: TODO!

src/utils/google-signin-utils.js
elijahsoria: Decided on using Promise.resolve(response) so that we can accept either a promise or the value itself.

examples/google-signin/google-signin-iframe.html
elijahsoria: Added in a note that IE doesn't support fetch. We are okay with publishers not supporting it if they don't want to.

src/utils/google-signin-utils.js
elijahsoria: Modified the function to handle promises.

ChrisAntaki (approved):

ChrisAntaki (approved):
Thanks for adding those tests!

examples/google-signin/google-signin-iframe.html
jpettitt: Can you expand the notes to explain that pub should persist the GIS response data, make an account if needed or return state for one that already exists, and set a cookie.  Regarding IE11 expand the not to indicate that if a pub wants to support IE11 they will need to use xhr

jpettitt (dismissed):

ChrisAntaki (approved):

PR #1247, opened 2020-09-03
The demo now validates Google Article Access (GAA) URL params

ChrisAntaki:


examples/sample-pub/publisher.js
```diff
@@ -309,6 +310,62 @@ function whenDemoReady(callback) {
...
+ * @return {boolean}
+ */
+function isGaa() {
+  // Validate GAA params.
+  const params = getQueryParams();
+  if (!params.gaa_at) {
+    return false;
+  }
+  if (!params.gaa_n) {
+    console.error('SwG Entitlements: The `gaa_n` URL param is missing.');
```
elijahsoria, line 324: Could you add a comment that points to where the URL parameters are defined?

examples/sample-pub/publisher.js
```diff
@@ -309,6 +310,62 @@ function whenDemoReady(callback) {
...
+  }
+  if (parseInt(params.gaa_ts, 16) < Date.now() / 1000) {
+    console.error(
+      'SwG Entitlements: The `gaa_ts` URL param should contain a hex string timestamp which points to the future.'
+    );
+    return false;
+  }
+
+  // Validate referrer.
+  const GOOGLE_DOMAIN_RE = /(^|\.)google\.(com?|[a-z]{2}|com?\.[a-z]{2}|cat)$/;
```
elijahsoria, line 343: Did you come up with this regular expression? If not, could you link to the original location? RE matching google makes me nervous 😟

ChrisAntaki: I copied this regex from SwG's AMP extension: https://github.com/ampproject/amphtml/blob/c23bf281f817a2ee5df73f6fd45e9f4b71bb68b6/extensions/amp-subscriptions-google/0.1/amp-subscriptions-google.js#L56

I added a comment mentioning this

examples/sample-pub/publisher.js
ChrisAntaki: We don't have a public document describing these params yet. I added a TODO to link to such a document though!

elijahsoria (approved):

PR #1273, opened 2020-09-17
GSI: Renders GSI iframe

ChrisAntaki:


examples/google-signin/google-signin-iframe.html
```diff
@@ -2,7 +2,7 @@
 <html lang="en">
 <head>
   <meta charset="UTF-8">
-  <script async src="https://news.google.com/swg/js/v1/google_signin.js"></script>
+  <script src="<% swgGsiJsUrl %>"></script>
```
elijahsoria, line 5: I like the dynamic changing but I worry that it will cause confusion for publishers. This would be okay to do if we have a comment describing how to inject this URL and where the production one lives.

examples/google-signin/google-signin-iframe.html
```diff
@@ -43,10 +43,10 @@
     };
 
     window.onload = () => {
-        const creator = new SwgGoogleSigninCreator(
+        const iframe = new SwgGsiIframe(
```
elijahsoria, line 46: I prefer the original name because the class doesn't create an iframe, it only will create and handle the buttons on the iframe. 

Also I prefer to spell out Google Sign In instead of GSI since it isn't commonly used.

examples/sample-pub/sample-pub-app.js
```diff
@@ -46,6 +41,14 @@ const SWG_JS_URLS = {
   tt: 'https://news.google.com/swg/js/v1/swg-tt.js',
 };
 
+const SWG_GSI_JS_URLS = {
+  local: '/dist/subscriptions-gsi.max.js',
+  local_min: '/dist/subscriptions-gsi.js',
+  prod: 'https://news.google.com/swg/js/v1/swg-gsi.js',
+  autopush: 'https://news.google.com/swg/js/v1/swg-gsi-autopush.js',
```
elijahsoria, line 48: Do publishers need these? Wouldn't they only use the prod version and not even the local version? The shouldn't be messing with anything in the helper file.

examples/google-signin/google-signin-iframe.html
ChrisAntaki: I added a comment above the script

examples/sample-pub/sample-pub-app.js
ChrisAntaki: SwG developers need these

examples/google-signin/google-signin-iframe.html
ChrisAntaki: Good call, I named it `SwgGoogleSignInButtonCreator`

elijahsoria (approved):

examples/google-signin/google-signin-iframe.html
```diff
@@ -43,10 +47,10 @@
     };
 
     window.onload = () => {
-        const creator = new SwgGoogleSigninCreator(
-          ALLOWED_ORIGINS, CLIENT_ID, publisherGoogleSignInCallback, window
```
elijahsoria, line 47: Could you inform Martin about this change?

src/utils/google-signin-utils.js
```diff
@@ -14,162 +14,230 @@
...
  */
 
-/** Sentinel used to tell parent that we are communicating with it. */
-export const SENTINEL = 'google_signin';
-/** Command name for when the parent frame is ready for iframe start. */
-export const PARENT_READY_COMMAND = 'parent_frame_ready';
-/** Command name for when the Google Sign-in iframe is ready. */
-export const INTERMEDIATE_IFRAME_READY_COMMAND = 'intermediate_iframe_ready';
-/** Command name for when the user's metering parameters are returned from the publisher. */
+import {GetEntitlementsParamsExternalDef} from '../../src/api/subscriptions';
```
elijahsoria, line 17: Is there a way to avoid this import? It looks like it only used for type checks. I would like it if this file doesn't have any imports.

ChrisAntaki: The closure compiler uses this type defintion import. This import shouldn't add to the filesize of the compiled JS though, since it's just used by the compiler to verify types. Was your concern related to filesize?

elijahsoria: Yep that was my concern. Thanks for explaining it to me.

examples/google-signin/google-signin-iframe.html
ChrisAntaki: For sure

elijahsoria (approved):

src/utils/google-signin-utils.js
ChrisAntaki: Np!

elijahsoria (approved):

PR #1305, opened 2020-10-09
Allow clicking outside of meter toast iframe to close the iframe.

elijahsoria:
Imports the new ToastCloseRequest message and allows for a click event occurring outside of the meter toast iframe to close the iframe. Does this by sending a ToastCloseRequest to the backend on a click event. We then remove the click event during the onCancel callback so future clicks don't send a message to the iframe backend.


src/runtime/meter-toast-api.js
```diff
@@ -36,6 +39,13 @@ export class MeterToastApi {
     /** @private @const {!../components/dialog-manager.DialogManager} */
     this.dialogManager_ = deps.dialogManager();
 
+    /** @private @const {!function()} */
+    this.sendCloseRequestFunction_ = () => {
+      const closeRequest = new ToastCloseRequest();
+      closeRequest.setClose(true);
```
ChrisAntaki, line 45: This field seems unnecessary at first glance

ChrisAntaki (approved):

PR #1333, opened 2020-10-27
Regwall: Switches to GSI's "popup" mode (without iframe shim for now)

ChrisAntaki:
This PR will help me verify the flow E2E in the GNews app

### DIFF
--- a/examples/sample-pub/metering.js
+++ b/examples/sample-pub/metering.js
@@ -44,7 +44,14 @@ const MeteringDemo = {
     GaaMeteringRegwall.signOut().then(() => void location.reload());
   },
 
-  /** Mocks registration of a user, given data from Google Sign-In. */
+  /**
+   * Mocks registration of a user, given data from Google Sign-In.
+   *
+   * This is suitable for a demo. For real integrations, we
+   * advise setting a 1st party, secure, HTTP-only cookie,
+   * so it lives past 7 days in Safari.
+   * https://webkit.org/blog/10218/full-third-party-cookie-blocking-and-more/
+   */
   registerUser: (googleSignInUser) => {
     // Record the registration timestamp in seconds (not milliseconds).
     localStorage.meteringRegistrationTimestamp = Math.floor(Date.now() / 1000);
--- a/examples/sample-pub/metering.js
+++ b/examples/sample-pub/metering.js
@@ -44,6 +44,17 @@ const MeteringDemo = {
     GaaMeteringRegwall.signOut().then(() => void location.reload());
   },
 
+  /** Mocks registration of a user, given data from Google Sign-In. */
+  registerUser: (googleSignInUser) => {
+    // Record the registration timestamp in seconds (not milliseconds).
+    localStorage.meteringRegistrationTimestamp = Math.floor(Date.now() / 1000);
+
+    // Record the user's name, for the metering demo.
+    localStorage.meteringUsername = googleSignInUser
+      .getBasicProfile()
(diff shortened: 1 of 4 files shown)


examples/sample-pub/metering.js
```diff
@@ -44,6 +44,17 @@ const MeteringDemo = {
     GaaMeteringRegwall.signOut().then(() => void location.reload());
   },
 
+  /** Mocks registration of a user, given data from Google Sign-In. */
```
elijahsoria, line 47: Could you add a comment about what pubs should do in production?

src/utils/gaa.js
```diff
@@ -187,19 +187,35 @@ const REGWALL_HTML = `
...
+ *   getBasicProfile: function(): {
+ *     getName: function(): string,
+ *     getGivenName: function(): string,
+ *     getFamilyName: function(): string,
+ *     getImageUrl: function(): string,
+ *     getEmail: function(): string,
+ *   },
+ * }} GoogleSignInUserDef
+ */
+let GoogleSignInUserDef;
```
elijahsoria, line 204: nit: Maybe rename to GoogleUserBasicProfile (or similar) to match the name used on the Google developer site.

src/utils/gaa.js
```diff
@@ -251,11 +266,7 @@ export class GaaMeteringRegwall {
         .then(
           () =>
             // Only initialize Google Sign-In once.
-            self.gapi.auth2.getAuthInstance() ||
-            self.gapi.auth2.init({
-              'ux_mode': 'redirect',
-              'redirect_uri': redirectUri,
-            })
+            self.gapi.auth2.getAuthInstance() || self.gapi.auth2.init()
```
elijahsoria, line 269: From the dev docs, it seems that client_id is a required parameter to init. Are you using something else?

https://developers.google.com/identity/sign-in/web/reference#gapiauth2initparams

src/utils/gaa.js
```diff
@@ -231,10 +247,9 @@ export class GaaMeteringRegwall {
   /**
    * Configures Google Sign-In.
    * @nocollapse
-   * @param {{ redirectUri: string }=} params
    * @return {!Promise}
    */
-  static configureGoogleSignIn({redirectUri} = {redirectUri: ''}) {
+  static configureGoogleSignIn() {
```
elijahsoria, line 252: Should this be made private? I see that it is getting called within renderGoogleSignInButton_(). Should users be able to call this function or only allow show()?

src/utils/gaa.js
ChrisAntaki: I updated this to GoogleUser, to match the docs. The basic profile is something the GoogleUser contains

examples/sample-pub/metering.js
ChrisAntaki: I updated the docs to suggest setting a 1p http-only cookie so it lives past 7 days in Safari

src/utils/gaa.js
ChrisAntaki: I'm defining the Client ID in a meta tag. This is the simple method GSI tells developers about first in their docs https://developers.google.com/identity/sign-in/web/sign-in#specify_your_apps_client_id

elijahsoria (approved):

src/utils/gaa.js
ChrisAntaki: I just privatized it

elijahsoria (approved):

PR #1349, opened 2020-11-05
E2E: Removes dependency on Google account

ChrisAntaki:
- Updates `checkPayment` to verify gPay window is opened and was redirected to a Google Sign-In prompt
- Runs `gulp check` job in parallel with unit and e2e test jobs, to speed up Travis runs


qidonna (approved):
Thank you for fixing and enabling e2e tests!!!!

PR #1361, opened 2020-11-13
Regwall: i18n

ChrisAntaki:


assets/i18n/strings/pt_BR.xlb
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<localizationbundle locale="pt-BR">
+<messages>
+
+
+
+
+<msg desc="This title text introduces people to the idea that Google helps them get more (free articles)." hidden="0" name="GAA_REGWALL_TITLE" xml:space="default">Veja mais com o Google</msg>
+<msg desc="This description text tells people they are out of free articles from a given publication. Then it tells people they can unlock more free articles if they create a publisher account with their Google account." hidden="0" name="GAA_REGWALL_DESCRIPTION" xml:space="default"><strong></strong>This article usually requires payment, but you can keep reading, compliments of Google, when you register with <ph name="PUBLICATION"><ex>AP News</ex>{publication}</ph> using your Google account.</msg>
```
elijahsoria, line 9: I see that this msg is in English whereas the other two are in Portuguese. Is that expected? The same thing is true with fr.xlb.

assets/i18n/strings/compile.js
```diff
@@ -0,0 +1,67 @@
...
+
+async function main() {
+  const files = await fs.readdir(__dirname);
+  const xlbFiles = files.filter((name) => /\.xlb$/.test(name));
+  const localesPerMessage = {};
+  for (const xlbFile of xlbFiles) {
+    const xml = await fs.readFile(__dirname + '/' + xlbFile, 'utf8');
+    const $ = cheerio.load(xml);
+    let locale = $('localizationbundle').attr('locale').toLocaleLowerCase();
+    if (locale === 'en-gb') {
```
elijahsoria, line 30: What about if locale.startsWith('en-')? Will we have to do the same thing for 'en-us'?

ChrisAntaki: We're only expecting this one English variant

assets/i18n/strings/pt_BR.xlb
ChrisAntaki: Yep! This message is newer and hasn't been translated yet

assets/i18n/strings/compile.js
ChrisAntaki: Checking for all "en-" locales wouldn't hurt though. I just updated the code to do that :)

elijahsoria (approved):

PR #1390, opened 2020-12-03
Regwall: Only render if URL has fresh GAA URL params

ChrisAntaki:


src/utils/gaa.js
```diff
@@ -479,8 +491,45 @@ export class GaaMeteringRegwall {
       regwallContainer.remove();
     }
   }
+
+  /**
+   * Returns true if the URL contains fresh Google Article Access (GAA) params.
```
elijahsoria, line 496: nit: "...contains fresh and valid Google..."

src/utils/gaa.js
```diff
@@ -479,8 +491,45 @@ export class GaaMeteringRegwall {
...
+   * @private
+   * @return {boolean}
+   */
+  static urlContainsFreshGaaParams_() {
+    const params = parseQueryString(
+      GaaMeteringRegwall.location_.search.split('?')[1]
```
elijahsoria, line 502: We might want to check for only the first '?' in the string and consider everything after that valid: https://tools.ietf.org/html/rfc3986#section-3.4

elijahsoria: It looks like the parseQueryString function already does some splitting for you: https://github.com/subscriptions-project/swg-js/blob/ad4118da72087b1b6002e8fc2740945bc281fcfe/src/utils/url.js#L163

elijahsoria (approved):

src/utils/gaa.js
ChrisAntaki: I wanted to avoid the term "valid" since this function isn't validating the params, other than checking the timestamp

src/utils/gaa.js
ChrisAntaki: Nice catch!

PR #1395, opened 2020-12-07
SwG basic runtime and API skeleton.

stellachui:
The implementation will follow in separate changelists.


src/runtime/runtime-basic.js
```diff
@@ -0,0 +1,200 @@
...
+  /**
+   * @param {!Window|!Document|!Doc} winOrDoc
+   * @param {!../model/page-config.PageConfig} pageConfig
+   */
+  constructor(winOrDoc, pageConfig) {
+    this.configuredClassicRuntime_ = new ConfiguredRuntime(
+      winOrDoc,
+      pageConfig,
+      /* integr= */ null,
+      /* config= */ null
```
qidonna, line 166: null is not needed here, but aren't integr and config needed?

src/api/basic-subscriptions.js
```diff
@@ -0,0 +1,51 @@
...
+    type,
+    isAccessibleForFree,
+    isPartOfType,
+    isPartOfProductId,
+    autoPromptType = AutoPromptType.NONE} = {}) {}
+}
+
+/** @enum {string} */
+export const AutoPromptType = {
+  NONE: 'none',
```
qidonna, line 48: none means button?

src/runtime/runtime-basic.js
```diff
@@ -0,0 +1,200 @@
...
+ */
+let basicRuntimeInstance_;
+
+/**
+ * Returns runtime for testing if available. Throws if the runtime is not
+ * initialized yet.
+ * @visibleForTesting
+ * @return {!BasicRuntime}
+ */
+export function getBasicRuntime() {
```
qidonna, line 34: rename this file as basic-runtime.js so it will be consistent with basicRuntime, basicSubscription, etc.

src/api/basic-subscriptions.js
stellachui: No, none just means no auto prompt. Buttons are added to the html of the page, and we automatically search for the attribute swg-standard-button:subscription|contribution and set those up.

src/runtime/runtime-basic.js
stellachui: They're used for the config the config promise, and if unspecified, still be set to default. I am following up with Shams re:whether we need to make certain things configurable, so this value may change in the future.

src/runtime/runtime-basic.js
stellachui: Done.

qidonna (approved):

src/runtime/basic-runtime.js
```diff
@@ -0,0 +1,203 @@
...
+ * distributed under the License is distributed on an "AS-IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {AutoPromptType} from '../api/basic-subscriptions';
+import {ConfiguredRuntime} from './runtime';
+
+const RUNTIME_BASIC_PROP = 'SWG_BASIC';
```
ChrisAntaki, line 20: Could you rename this constant to "BASIC_RUNTIME_PROP"? I think that would be more consistent with the other names in this script

ChrisAntaki (approved):

src/runtime/basic-runtime.js
stellachui: Good catch, done.

PR #1408, opened 2020-12-14
Adds logging and tests to meter toast iframe

elijahsoria:
Adds the ability to log from MeterToastApi and adds log events for ACTION_METER_TOAST_CLOSED_BY_ARTICLE_INTERACTION, IMPRESSION_METER_TOAST, and EVENT_OFFERED_METER. Adds tests to check that logging is being done correctly.


src/runtime/meter-toast-api.js
```diff
@@ -43,21 +44,31 @@ export class MeterToastApi {
     /** @private @const {!../components/dialog-manager.DialogManager} */
     this.dialogManager_ = deps.dialogManager();
 
+    const iframeArgs = this.activityPorts_.addDefaultArguments(
+      feArgs({
+        isClosable: true,
+        publicationId: deps.pageConfig().getPublicationId(),
```
mborof, line 50: Default arguments populates some of these other values for you.

src/runtime/meter-toast-api.js
```diff
@@ -43,21 +44,31 @@ export class MeterToastApi {
...
     this.activityIframeView_ = new ActivityIframeView(
       this.win_,
       this.activityPorts_,
       feUrl('/metertoastiframe'),
-      feArgs({
-        publicationId: deps.pageConfig().getPublicationId(),
-        productId: deps.pageConfig().getProductId(),
-        hasSubscriptionCallback: deps.callbacks().hasSubscribeRequestCallback(),
-      }),
+      iframeArgs,
```
mborof, line 60: this.activityPorts_.addDefaultArguments({
  isClosable: true,
  hasSubscriptionCallback: deps.callbacks().hasSubscribeRequestCallback(),
})

src/runtime/meter-toast-api.js
elijahsoria: Good catch!

src/runtime/meter-toast-api.js
elijahsoria: Added this with feArgs in between.

mborof: drop feArgs.  Default args takes care of it

elijahsoria: Done!

src/runtime/meter-toast-api.js
```diff
@@ -43,21 +44,27 @@ export class MeterToastApi {
     /** @private @const {!../components/dialog-manager.DialogManager} */
     this.dialogManager_ = deps.dialogManager();
 
+    const iframeArgs = this.activityPorts_.addDefaultArguments({
```
mborof, line 47: As a followup on this, there is probably some central guidance that was followed when creating this initially?  Can we update that guidance?

mborof (approved):

src/runtime/meter-toast-api.js
elijahsoria: I think Chen copied a previous class that was using feArgs without addDefaultArguments. Here is an example: https://github.com/subscriptions-project/swg-js/blob/c51bdb5af989317b08922d649441f77e91be49e8/src/runtime/login-notification-api.js

PR #1430, opened 2021-01-06
Log Regwall Impressions

mborof:


src/runtime/meter-regwall-api.js
```diff
@@ -59,6 +60,11 @@ export class MeterRegwallApi {
     this.deps_
       .callbacks()
       .triggerFlowStarted(SubscriptionFlows.SHOW_METER_REGWALL);
-    return this.dialogManager_.openView(this.activityIframeView_);
+    return this.dialogManager_.openView(this.activityIframeView_).then(() => {
+      this.deps_.eventManager().logSwgEvent(AnalyticsEvent.IMPRESSION_REGWALL);
```
ChrisAntaki, line 64: Are these two events always sent together? If so, is that redundant?

mborof: In the future if there is something else that is a regwall it can trigger IMPRESSION_{OtherRegwallKeyname}_REGWALL and IMPRESSION_REGWALL.  If someone wants to parse on "any" regwall they use the generic and if they care specific about showcase regwall they use that one.

ChrisAntaki: That makes sense! Could you add a comment explaining this?

mborof: Done

src/runtime/meter-regwall-api.js
```diff
@@ -59,6 +60,12 @@ export class MeterRegwallApi {
     this.deps_
       .callbacks()
       .triggerFlowStarted(SubscriptionFlows.SHOW_METER_REGWALL);
-    return this.dialogManager_.openView(this.activityIframeView_);
+    return this.dialogManager_.openView(this.activityIframeView_).then(() => {
+      // Log a generic and specific regwall event
```
ChrisAntaki, line 64: Could you update this comment so it describes why there are two events which could seem redundant?

mborof: We do this double logging in quite a few places, don't have multiple sentence comments explaining why we do double logging for them anywhere else.  It's just kind of standard procedure.  I'm not exactly clear what or why something needs further clarification so I'm not sure what you're looking for.  If you have a specific thing you want to suggest, I'm happy to consider it.

ChrisAntaki: I'm pretty familiar with the swg-js codebase, and I found the code confusing. I was hoping you could add a comment that basically answered the question, "How is this not redundant?". I think this would help other engineers working on swg-js too

mborof: I'm just not exactly sure how to clarify what was confusing you and thought you might have a suggestion.  Did that do it?

ChrisAntaki: This is getting better! Could you also mention the publisher would log their own event too? Re: "In the future if there is something else that is a regwall it can trigger IMPRESSION_{OtherRegwallKeyname}_REGWALL and IMPRESSION_REGWALL"

mborof: I've added additional clarification and details

ChrisAntaki: Thanks! The comment still confuses me a bit...

I'm not sure what this part means: "other regwalls could also log this event if they have a regwall" 

Also, I'm curious about the "screen" part of the "Specific screen impression" comment. I think that the "Specific screen impression" is the same concept as the "specific impression" mentioned in the earlier comment. Is "screen" referring to a specific dialog?

ChrisAntaki: Also, feel free to reach out via chat if that's easier!

ChrisAntaki (approved):

PR #1434, opened 2021-01-11
Fill in Showcase Entitlement Events

mborof:
Initial take, tests not yet implemented

PR #1437, opened 2021-01-12
Create URL utilities

mborof:


ChrisAntaki:
Will `urlContainsFreshGaaParams` only be used by the `GaaMeteringRegwall` class?

ChrisAntaki:
I'm not sure I see the benefit of creating this new `GaaUtil` class currently

mborof:
https://github.com/subscriptions-project/swg-js/pull/1434/files#diff-9ca79101e09937bb7155a923ee939f378e908de75b855103591091bb85d485c2R1052

mborof:
I called it isGaa here but the intention was the same.  We only send these events in a gaa context so we need to validate the context is correct.

mborof:
https://docs.google.com/document/d/1btBfkEnjgRDZq6bBLSYry9KyIrchDmGgqY_cEhxTj4s/edit?resourcekey=0-AgUnbhZk5PEEUagMnnHh4A#heading=h.bsq5f1rmvy1n

ChrisAntaki:
I'd prefer to see that standalone method, as opposed to this new static class

ChrisAntaki:
I think that'd be a simpler approach

mborof:
isSecure(GaaUtil.location_) &&
      isSecure(GaaUtil.referrer_) &&
      isGoogleDomain(GaaUtil.referrer_) &&

I did it as a static method to keep it consistent with the way you had other code in the file.  

The reason you had done it this way when you wrote this file (I'm fairly certain) is that you can't override the location or referrer methods (as far as I can tell) when testing so the static object gave you a test.

The other way I had was to have optional parameters that are only there for testing.  This complicated the user interface of the function so I wasn't as thrilled by it.

I suppose I could make the static object private, but it doesn't seem to matter much either way...

ChrisAntaki:
What if you added the referrer property to `GaaMeteringRegwall`, and then renamed the `isValidGaaContext` method to `isAnalyticsEnabled` and moved it to `GaaMeteringRegwall` as well?

mborof:
Because then the function would look like it had something to do with analytics and/or the gaa regwall when it doesn't really directly connotate with either.

ChrisAntaki:
I must not understand the function's purpose. Could you explain to me what it's doing?

mborof:
This function is intended to indicate the user is in a valid gaa / showcase context.

The function that will call this one is a function the publisher calls with information about their entitlements.  We can only use that entitlement information if the browser is in a valid gaa context.

ChrisAntaki:
> This function is intended to indicate the user is in a valid gaa / showcase context.

Valid in what sense? The user can still go through the Regwall flow, see the Metering Toast, get a free read, etc., right?

> The function that will call this one is a function the publisher calls with information about their entitlements. We can only use that entitlement information if the browser is in a valid gaa context.

What does the other function do?

mborof:
What do you think about this change instead?  I'll ultimately use the newly public function in 2 places and the new function in those places as well.

src/utils/url-test.js
```diff
@@ -368,19 +368,21 @@ describe('isSecure', () => {
...
-  it('first parameter should default to current page', () => {
-    const URL = parseUrl(self.window.location.href);
-    expect(isGoogleDomain(URL)).to.equal(isGoogleDomain());
+describe('wasReferredByGoogle', () => {
+  it("first parameter should default to current page's referrer", () => {
+    expect(wasReferredByGoogle(parseUrl(self.document.referrer))).to.equal(
+      wasReferredByGoogle()
+    );
+  });
+  it('should accept a secure Google referrer', () => {
```
ChrisAntaki, line 377: Could you add a newline above this?

ChrisAntaki:
This looks great!

ChrisAntaki (approved):

PR #1442, opened 2021-01-14
Improve test so it doesn't require updates

mborof:


ChrisAntaki:
I see this branch has conflicts. Please resolve those and let me know when it's ready for a review. Thanks!

ChrisAntaki (changes requested):
I see the `check` is failing. Could you fix that and then request a review?

src/utils/url-test.js
```diff
@@ -292,46 +290,51 @@ describe('serializeProtoMessageForUrl', () => {
...
+
+    // Clean up the message by internally converting it to an array and back into a message
+    // this standardizes different ways values could have been set.
+    const cleanupInput = new constructor(testMessage.toArray(false), false);
+    // Ensure both values are equal
+    expect(deserializedMessage).to.deep.equal(cleanupInput);
+  }
+
+  it('should serialize message', () =>
+    testSerialization(
```
ChrisAntaki, line 317: I think these tests are too "DRY". Could you make them "DAMP", to improve clarity? https://testing.googleblog.com/2019/12/testing-on-toilet-tests-too-dry-make.html

PR #1444, opened 2021-01-14
Don't close meter toast immediately on scroll.

elijahsoria:
Currently, the meter toast iframe closes immediately on any scroll on desktop. Now, we will only close the iframe on desktop if the scroll distance is greater than 100.

### DIFF
--- a/src/runtime/meter-toast-api.js
+++ b/src/runtime/meter-toast-api.js
@@ -128,11 +128,11 @@ export class MeterToastApi {
         } else {
           let start, scrollTimeout;
           this.scrollEventListener_ = () => {
-            start = start || this.win_./*REVIEW*/pageYOffset;
+            start = start || this.win_./*REVIEW*/ pageYOffset;
             this.win_.clearTimeout(scrollTimeout);
             scrollTimeout = this.win_.setTimeout(() => {
               // If the scroll is longer than 100, close the toast.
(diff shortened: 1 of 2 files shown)


src/runtime/meter-toast-api.js
```diff
@@ -117,11 +120,25 @@ export class MeterToastApi {
...
-        const $body = this.win_.document.body;
-        setStyle($body, 'overflow', 'hidden');
+        // while swiping on the iframe only on mobile.
+        if (this.isMobile_()) {
+          const $body = this.win_.document.body;
+          setStyle($body, 'overflow', 'hidden');
+        } else {
+          let start, scrollTimeout;
+          this.scrollEventListener_ = () => {
+            start = start || this.win_./*REVIEW*/pageYOffset;
```
elijahsoria, line 131: @ChrisAntaki I see that pageYOffset and scrollY are both banned but don't know of an alternative. Do you think this is okay?

https://github.com/subscriptions-project/swg-js/blob/7bd1026dbb17330c21ac108591eec804a064dd3c/build-system/tasks/check-rules.js#L137

elijahsoria (commented):
Ignoring the lint error since it conflicts with the advice from the banned usage file.

src/runtime/meter-toast-api.js
ChrisAntaki: Seems fine to me

ChrisAntaki (approved):

PR #1455, opened 2021-01-21
Add a PageConfigWriter that writes the metadata for users of BasicSubscription.

stellachui:
Add a PageConfigWriter that writes the metadata specified in BasicSubscriptions' init into the HTML of the page. Implements the configuration within basic-runtime.com.


src/model/page-config-writer-test.js
```diff
@@ -0,0 +1,255 @@
...
+            '@type': ['CreativeWork', 'Product'],
+          },
+        })
+      );
+      await expect(configWriter.writeConfigWhenReady(markupConfig)).to.be
+        .eventually.fulfilled;
+      expect(gd.getHead().childNodes).to.have.length(1);
+    });
+
+    it('should write a new node if there is an existing but invalid schema', async () => {
```
jpettitt, line 110: Type should be additive so `@type: "foobar"` would become `@type: [ "foobar", "CreativeWork", "Product" ]` in the json.   Needs to handle both string and array `@type: "foobar"` and `@type: ["foobar"]`

jpettitt (commented):
Added a comment in the test because it seemed to be the best place to capture the change needed to the writer.

src/model/page-config-writer-test.js
stellachui: Done, I've made both the @type and the isPartOf[@type] fields take either a string or an array of strings, and merged those fields in the PageConfigWriter.

stellachui:
Updated behavior based on feedback, and added more tests to cover the new behavior. PTAL, thanks!

jpettitt (approved):

PR #1469, opened 2021-01-29
Auto pingback if load fails

mborof:
Still need to fix/create tests

### DIFF
--- a/src/runtime/meter-toast-api-test.js
+++ b/src/runtime/meter-toast-api-test.js
@@ -55,7 +55,6 @@ describes.realWin('MeterToastApi', {}, (env) => {
       'matches': true,
       'addListener': (callback) => callback,
     });
-
     messageMap = {};
     pageConfig = new PageConfig(productId);
     runtime = new ConfiguredRuntime(win, pageConfig);
--- a/src/runtime/meter-toast-api-test.js
+++ b/src/runtime/meter-toast-api-test.js
@@ -233,10 +233,10 @@ describes.realWin('MeterToastApi', {}, (env) => {
       );
     await win.dispatchEvent(new Event('scroll'));
     expect(onConsumeCallbackFake).to.not.be.called;
-    win.scrollY = 10;
+    win.pageYOffset = 10;
     await win.dispatchEvent(new Event('scroll'));
     expect(onConsumeCallbackFake).to.not.be.called;
-    win.scrollY = 500;
+    win.pageYOffset = 500;
     await win.dispatchEvent(new Event('scroll'));
     expect(messageStub).to.be.calledOnce.calledWith(TOAST_CLOSE_REQUEST);
     expect(onConsumeCallbackFake).to.be.calledOnce;
--- a/src/runtime/meter-toast-api-test.js
+++ b/src/runtime/meter-toast-api-test.js
@@ -30,6 +30,10 @@ import {
 } from '../proto/api_messages';
 import {getStyle} from '../utils/style';
 
+const AUTO_PINGBACK_TIMEOUT = 10000;
+const TOAST_CLOSE_REQUEST = new ToastCloseRequest();
+TOAST_CLOSE_REQUEST.setClose(true);
+
 describes.realWin('MeterToastApi', {}, (env) => {
   let win;
   let runtime;
@@ -41,6 +45,8 @@ describes.realWin('MeterToastApi', {}, (env) => {
   let meterToastApi;
   let port;
   let dialogManagerMock;
+  let onConsumeCallbackFake;
+  let isMobile;
   const productId = 'pub1:label1';
 
   beforeEach(() => {
@@ -49,6 +55,7 @@ describes.realWin('MeterToastApi', {}, (env) => {
       'matches': true,
       'addListener': (callback) => callback,
     });
+
     messageMap = {};
     pageConfig = new PageConfig(productId);
     runtime = new ConfiguredRuntime(win, pageConfig);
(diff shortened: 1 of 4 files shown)


src/runtime/meter-toast-api.js
```diff
@@ -29,6 +29,8 @@ export const IFRAME_BOX_SHADOW =
   'rgba(60, 64, 67, 0.3) 0px -2px 5px, rgba(60, 64, 67, 0.15) 0px -5px 5px';
 export const MINIMIZED_IFRAME_SIZE = '420px';
 
+const AUTO_PINGBACK_TIMEOUT = 5000;
```
elijahsoria, line 32: I would prefer this be 10-15 seconds. 5 seconds seems like a somewhat normal loading time on slow connections.

IMO this should be the longest we should expect our users to wait before we bail on the toast request.

elijahsoria: Also add a comment please :)

src/runtime/meter-toast-api.js
```diff
@@ -65,6 +67,13 @@ export class MeterToastApi {
...
+    /**
+     * A timeout set while things are loading.  If the user dismisses the popup before
+     * loading is complete, this timeout ensures we consume the metering entitlement
+     * and cleanup the page properly.
+     * @private {?number} */
+    this.rapidCloseTimeout_ = null;
+
     /** @private @const {!function()} */
     this.sendCloseRequestFunction_ = () => {
       this.deps_
```
elijahsoria, line 79: I think this log would be best in the else statement below. Seems like the timeout should have its own log event.

elijahsoria (commented):
Seems like a good approach to me but I would definitely like to see this work in tests before submission. You might be able to add tests by mocking openDialog() to be a long delay so the timeout can complete.

src/runtime/meter-toast-api-test.js
```diff
@@ -49,6 +55,7 @@ describes.realWin('MeterToastApi', {}, (env) => {
       'matches': true,
       'addListener': (callback) => callback,
     });
+
```
elijahsoria, line 58: nit: remove line

src/runtime/meter-toast-api.js
```diff
@@ -65,18 +75,32 @@ export class MeterToastApi {
...
+      } else {
+        const closeRequest = new ToastCloseRequest();
+        closeRequest.setClose(true);
+        this.activityIframeView_.execute(closeRequest);
+        this.removeCloseEventListener();
+      }
+
       this.deps_
         .eventManager()
         .logSwgEvent(
```
elijahsoria, line 100: By the TODO, it seems like you meant for this to go inside of the else. Is that right? This will still get logged on both timeout and regular close.

src/runtime/meter-toast-api.js
mborof: I agree, I'd prefer to get the fix in now and worry about the event another time.  I'll create a buganizer to do so

src/runtime/meter-toast-api.js
```diff
@@ -29,6 +29,16 @@ export const IFRAME_BOX_SHADOW =
   'rgba(60, 64, 67, 0.3) 0px -2px 5px, rgba(60, 64, 67, 0.15) 0px -5px 5px';
 export const MINIMIZED_IFRAME_SIZE = '420px';
 
+// If the user is able to close the dialog before loading is complete,
+// this timeout ensures we still pingback a metering entitlement after X ms.
+// This timeout should represent the longest time it could reasonably take
+// to load a SwG BOQ iframe.
+const AUTO_PINGBACK_TIMEOUT = 10000;
+
+function getScrollY(win) {
```
elijahsoria, line 38: Why did you move move this from `this.win_. pageYOffset`? Is this for testing?

mborof: It didn't work on my computer until I did this.  I could not get it to pass tests

src/runtime/meter-toast-api.js
mborof: I'd prefer to log something still.  You can tell it's the timeout still because you won't have a meter toast impression or offer event.

elijahsoria: LGTM

src/runtime/meter-toast-api.js
mborof: https://b.corp.google.com/issues/178834552

src/runtime/meter-toast-api-test.js
```diff
@@ -231,21 +232,49 @@ describes.realWin('MeterToastApi', {}, (env) => {
...
+    win.pageYOffset = 500;
     await win.dispatchEvent(new Event('scroll'));
-    expect(messageStub).to.be.calledOnce.calledWith(toastCloseRequest);
+    expect(messageStub).to.be.calledOnce.calledWith(TOAST_CLOSE_REQUEST);
+    expect(onConsumeCallbackFake).to.be.calledOnce;
+  });
+
+  it('should timeout if the dialog closes very quickly', async () => {
+    meterToastApi.isMobile_.restore();
+    sandbox.stub(meterToastApi, 'isMobile_').returns(false);
```
elijahsoria, line 246: nit: remove this line since it should be independent of what is being tested.

elijahsoria (approved):

jpettitt:
The toast iframe has a an [`onCancel()`](https://github.com/subscriptions-project/swg-js/blob/b625522d23e6ad73e5bbe69b487fe058088a4e79/src/ui/activity-iframe-view.js#L210) method we should hook - it should fire if the iframe is closed prematurely.  @mborof Please do a new PR to use that.

jpettitt:
I think the simplest solution is to add a `.catch()` to https://github.com/subscriptions-project/swg-js/blob/5501f26bcc0702559abc6b1bc3433f61fd8d06c3/src/runtime/meter-toast-api.js#L132 

and in the catch trigger the pingback and cleanup the dialog.  That would allow us to revert this PR removing the timeout.  @elijahsoria can you do this since @mborof is out?

PR #1476, opened 2021-02-05
Replace timeout with cancellation catch in Meter Toast

elijahsoria:
Does a partial rollback of the timeout handling done in [this PR](https://github.com/subscriptions-project/swg-js/pull/1469), replacing the 10 second timeout on the dialog opening with a call to the onConsume callback on all cancellations. This protects us against the case of a user closing the dialog before it renders without the meter being decremented without using a timeout. Adds a boolean to protect against calling onConsume twice. Adds test.


jpettitt:
That's a great plan, Walter. That's f***in' ingenious, if I understand it correctly. It's a Swiss f***in' watch.

LGTM

jpettitt (approved):

PR #1479, opened 2021-02-10
Send metering pingback only on metering requests

elijahsoria:
Adds a check in the entitlements pingback methods that checks if the request URL has valid GAA params. Does this by modifying and using the urlContainsFreshGaaParams which checks if all necessary elements are present. Adds tests.


src/runtime/entitlements-manager.js
```diff
@@ -194,17 +201,25 @@ export class EntitlementsManager {
     const jwt = new EntitlementJwt();
     jwt.setSource(entitlement.source);
     jwt.setJwt(entitlement.subscriptionToken);
+    const params = parseQueryString(this.win_.location.search);
```
ChrisAntaki, line 204:
```suggestion
    const nonce = parseQueryString(this.win_.location.search)['gaa_n'];
```

ChrisAntaki: This would be a bit more descriptive

src/runtime/entitlements-manager.js
```diff
@@ -230,22 +245,29 @@ export class EntitlementsManager {
...
   }
 
   // Informs the Entitlements server about the entitlement used
   // to unlock the page.
   postEntitlementsRequest_(
     usedEntitlement,
     entitlementResult,
-    entitlementSource
+    entitlementSource,
+    opt_nonce = ''
```
ChrisAntaki, line 263: Could you rename this to `nonce` or `optionalNonce`? https://google.github.io/styleguide/jsguide.html#naming-parameter-names

src/runtime/entitlements-manager.js
ChrisAntaki: I also suggest this change on line 248

src/utils/gaa.js
```diff
@@ -302,10 +302,12 @@ let GoogleUserDef;
 
 /**
  * Returns true if the URL contains fresh Google Article Access (GAA) params.
+ * @param {string=} opt_urlParams
```
ChrisAntaki, line 305: Could you rename this to `optionalQueryString`? https://google.github.io/styleguide/jsguide.html#naming-parameter-names

src/utils/gaa.js
```diff
@@ -302,10 +302,12 @@ let GoogleUserDef;
...
  * @return {boolean}
  */
-export function urlContainsFreshGaaParams() {
-  const params = parseQueryString(GaaMeteringRegwall.location_.search);
+export function urlContainsFreshGaaParams(opt_urlParams = '') {
+  const urlParams = opt_urlParams || GaaMeteringRegwall.location_.search;
```
ChrisAntaki, line 309:
```suggestion
  const queryString = optionalQueryString || GaaMeteringRegwall.location_.search;
```

ChrisAntaki: Actually, I'm going to send you a PR that refactors this method a bit... please feel free to hold off on changing it

src/runtime/entitlements-manager.js
```diff
@@ -194,17 +201,25 @@ export class EntitlementsManager {
...
+      EntitlementSource.GOOGLE_SHOWCASE_METERING_SERVICE,
+      params['gaa_n']
     );
   }
 
   // Listens for events from the event manager and informs
   // the server about publisher entitlements and non-
   // consumable Google entitlements.
-  handleClientEvent_(event) {
+  possiblyPingbackOnClientEvent_(event) {
```
ChrisAntaki, line 216: This name is much more descriptive, thank you!

src/utils/gaa.js
ChrisAntaki: Here's the PR! https://github.com/subscriptions-project/swg-js/pull/1480

src/runtime/entitlements-manager.js
elijahsoria: optionalNonce it is.

src/utils/gaa.js
elijahsoria: Resolving since function is updated.

src/utils/gaa.js
elijahsoria: Thanks! Resolving since function is updated.

src/runtime/entitlements-manager-test.js
```diff
@@ -885,6 +894,28 @@ describes.realWin('EntitlementsManager', {}, (env) => {
       await manager.consumeMeter_(ents);
     });
 
+    it('should not send pingback with invalid GAA params', async () => {
+      nowStub.reset();
```
ChrisAntaki, line 898: I don't think this line is needed

src/runtime/entitlements-manager-test.js
```diff
@@ -1685,6 +1749,7 @@ describes.realWin('EntitlementsManager', {}, (env) => {
 
     it('should ignore malformed pushed entitlements', () => {
       storageMock.expects('set').withArgs('ents').never();
+      sandbox.stub(win, 'setTimeout').callsFake(() => {});
```
ChrisAntaki, line 1752: Just curious, why is this needed?

ChrisAntaki, line 1752: `setTimeout` should return a `Number`. Just curious though, is this line still needed?
```suggestion
      sandbox.stub(win, 'setTimeout').returns(1);
```

src/runtime/entitlements-manager-test.js
elijahsoria: You're right! Removing all calls to reset.

src/runtime/entitlements-manager-test.js
elijahsoria: Changed to returns(1). Without a stub, setTimeout will fail with a `TypeError: Illegal invocation`, I think because it hits L449 on entitlements-manager.js and doesn't know what to do without an implementation of setTimeout?

ChrisAntaki (approved):

mborof:
Can't our boq actions see the requesting browser tab's query string and pull out the nonce directly?

src/runtime/entitlements-manager.js
jpettitt: per Dan `nonce` is officially deprecated it's now `gaaToken` or in this case `optionalGaaToken`

jpettitt (approved):
LGTM modulo nonce name change

PR #1499, opened 2021-02-18
Showcase Regwall: Localizes Google Sign-In button

ChrisAntaki:
The Google Sign-In button won't localize itself :shrug:

### DIFF
--- a/src/utils/gaa.js
+++ b/src/utils/gaa.js
@@ -22,9 +22,9 @@
 // Thanks!
 
 import {I18N_STRINGS} from '../i18n/strings';
+import {getLanguageCodeFromElement} from './i18n';
 // eslint-disable-next-line no-unused-vars
 import {Subscriptions} from '../api/subscriptions';
-import {msg} from './i18n';
 import {parseJson} from './json';
 import {parseQueryString} from './url';
 import {setImportantStyles} from './style';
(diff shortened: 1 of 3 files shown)


ChrisAntaki:
image

ChrisAntaki:
CC @ed-bird

src/utils/gaa.js
```diff
@@ -259,7 +243,11 @@ body {
   display: none;
 }
 #${GOOGLE_SIGN_IN_BUTTON_ID} .abcRioButton.abcRioButtonBlue .abcRioButtonContents {
-  font-size: 15px !important;
+  font-size: 0 !important;
+}
+#${GOOGLE_SIGN_IN_BUTTON_ID} .abcRioButton.abcRioButtonBlue .abcRioButtonContents span[id^=not_signed_]:before {
```
elijahsoria, line 248: Could you add a comment about what this span does and why you need this class and the one above it?

src/utils/gaa.js
```diff
@@ -369,6 +357,16 @@ export class GaaMeteringRegwall {
    * @param {{ iframeUrl: string }} params
    */
   static render_({iframeUrl}) {
+    const languageCode = getLanguageCodeFromElement(self.document.body);
+
+    // Tell the iframe which language to render.
```
elijahsoria, line 362: Prefer to use `addQueryParam` here: https://github.com/subscriptions-project/swg-js/blob/41f226737a782a2afd56d312ff7f093b804cdb9f/src/utils/url.js#L190

ChrisAntaki: Good call, updated

src/utils/gaa.js
ChrisAntaki: I added a couple CSS comments, please let me know what you think

elijahsoria: I think I get it now. If we have localized text, we set the default class font size to 0px and add this class with 15px font?

What most confuses me most is what `span[id^=not_signed_]` looks for and how the two classes work together. If you could add in some stuff about that in the comment that would be great.

elijahsoria (approved):

ChrisAntaki:
> I think I get it now. If we have localized text, we set the default class font size to 0px and add this class with 15px font?

Just to clarify, the code in this PR is always hiding the default text and is then showing localized text instead, though it defaults to selecting "en" if nothing is specified

> What most confuses me most is what span[id^=not_signed_] looks for and how the two classes work together. If you could add in some stuff about that in the comment that would be great.

That's a good call out! I updated the code so both rules point to the "span[id^=not_signed_]" element. Does that simplify it to the point where the comments add enough clarity? Or would more clarity still be helpful?

ChrisAntaki:
Basically, the first rule hides the button's default "Sign in with Google" text using `font-size`. Then the second rule creates a pseudo-element using [`::before`](https://developer.mozilla.org/en-US/docs/Web/CSS/::before) and gives it localized text content and a non-zero `font-size`

elijahsoria (approved):

PR #1501, opened 2021-02-18
Revert Doc's getRootNode return type change to Node

stellachui:
And update the page config writer to pass in the Document accessed via Window..


stellachui:
@cvolzke4 FYI, for release 0.1.22.146.

jpettitt (approved):

PR #1503, opened 2021-02-19
Showcase: Exports swg-gaa.js to AMP

ChrisAntaki:
This PR makes Showcase classes and defs available to AMP extensions


exports/swg-gaa.js
```diff
@@ -0,0 +1,22 @@
...
+ * distributed under the License is distributed on an "AS-IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+export {
+  GaaGoogleSignInButton,
+  GaaMeteringRegwall,
+  GaaUserDef,
```
elijahsoria, line 20: I think this is unrelated to this PR but do GaaUserDef and GoogleUserDef need to be in gaa-main.js too?

elijahsoria (approved):

exports/swg-gaa.js
ChrisAntaki: swg-gaa.js doesn't need these defs at runtime, so it's best we leave them out of gaa-main.js

PR #1505, opened 2021-02-22
Add client config and auto prompt managers

stellachui:
Add client config and auto prompt managers which fetch publication specific client settings from the server. 

- Add new manager classes
- Add the ClientConfigManager to Deps (this will soon be needed for classic as well, to modify the pay flow depending on config)
- Also strips the safe JSON response prefix in fetcher, if present (used in the client config response)

Up next: Mini prompt display and logging of display to local storage.


src/api/basic-subscriptions.js
```diff
@@ -56,10 +56,16 @@ export class BasicSubscriptions {
...
-   * parameter is an option to force show the prompt, regardless of any display
-   * rules. This parameter is intended for preview purposes.
-   * @param {{alwaysShow: (boolean|undefined)}=} options
+   * function will be skipped, and the prompt will be displayed. The
+   * autoPromptType specifies which type of prompt should be displayed (see
+   * AutoPromptType below). The alwaysShow parameter is an option to force show
+   * the prompt, regardless of any display rules. This parameter is intended for
+   * preview purposes.
+   * @param {{
+   *   autoPromptType: (string|undefined),
```
ChrisAntaki, line 65:
```suggestion
   *   autoPromptType: (!AutoPromptType|undefined),
```

ChrisAntaki: Would it make sense to refer to the enum here?

src/runtime/fetcher-test.js
```diff
@@ -113,6 +113,23 @@ describes.realWin('XhrFetcher', {}, (env) => {
...
+        return Promise.resolve({
+          text: () => Promise.resolve(")]}'\n"),
+        });
+      });
+      sentInit = {
+        method: 'GET',
+        headers: {'Accept': 'text/plain, application/json'},
+        credentials: 'include',
+      };
+      fetcher.fetchCredentialedJson(sentUrl);
```
ChrisAntaki, line 130: Could you update this test to `expect` a certain return value from the method? It would be nice to verify it can remove the prefix and parse the JSON

src/runtime/client-config-manager-test.js
```diff
@@ -0,0 +1,98 @@
...
+    const expectedClientConfig = new ClientConfig(expectedAutoPromptConfig);
+    expect(clientConfig).to.deep.equal(expectedClientConfig);
+  });
+
+  it('getClientConfig should throw an error for undefined publication ID', async () => {
+    clientConfigManager = new ClientConfigManager(undefined, fetcher);
+    fetcherMock.expects('fetchCredentialedJson').never();
+
+    try {
+      clientConfigManager.getClientConfig();
```
ChrisAntaki, line 58: This test will pass if the error isn't thrown. Could you use the `expect(() => ...).to.throw` pattern instead?

src/runtime/auto-prompt-manager.js
```diff
@@ -0,0 +1,172 @@
...
+   *   - alwaysShow == true, used for demo purposes, OR
+   *   - There is no active entitlement found AND
+   *   - The user had not reached the maximum impressions allowed, as specified
+   *     by the publisher
+   * A prompt may not be displayed if the appropriate criteria are not met.
+   * @param {{
+   *   autoPromptType: (string|undefined),
+   *   alwaysShow: (boolean|undefined),
+   * }=} options
+   * @param {function()=} displayForLockedContentFn
```
ChrisAntaki, line 58: Could the `displayForLockedContentFn` param be part of the `options` object?

src/runtime/auto-prompt-manager.js
```diff
@@ -0,0 +1,172 @@
...
+      return false;
+    }
+
+    // The mini auto prompt is only for non-paygated content.
+    if (this.pageConfig_.isLocked()) {
+      return false;
+    }
+
+    // Don't cap subscription prompts.
+    if (options.autoPromptType == AutoPromptType.SUBSCRIPTION) {
```
ChrisAntaki, line 144: Could you use the triple equals whenever possible? It's more predictable
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness

src/runtime/fetcher-test.js
stellachui: Added a check for the parsed object.

src/runtime/client-config-manager-test.js
stellachui: Yup, done.

src/runtime/auto-prompt-manager.js
stellachui: It could, options were meant to contain the mini prompt configuration details, passed in from the BasicSubscriptions API, and displayForLockedContentFn kind of fell outside of that because it provides the alternative-to-mini-prompt display route. I don't feel too strongly about this and could see combining them to a more general params object though, so let me know if you'd like to see them combined. Just out of curiosity, is the way the parameters are currently structured against a design pattern?

src/runtime/auto-prompt-manager.js
stellachui: Done.

src/runtime/auto-prompt-manager.js
ChrisAntaki: I'd love to see them combined, I just think that would simplify things

src/runtime/runtime.js
```diff
@@ -687,6 +687,11 @@ export class ConfiguredRuntime {
     return this.storage_;
   }
 
+  /** @override */
+  clientConfigManager() {
```
ChrisAntaki, line 691: Could you add a test for this, just to keep the coverage at 100% for the PR?

ChrisAntaki (approved):

src/runtime/runtime.js
stellachui: Added.

src/runtime/auto-prompt-manager.js
stellachui: Done.

PR #1517, opened 2021-03-01
CI: Uses Java 8

ChrisAntaki:
# Changes
- Pins Java version to 8
- Pins Ubuntu version to 20.04

# Context
Our GitHub actions are using the `ubuntu-latest` virtual environment. GitHub updated this to Ubuntu 20.04 today, and it defaults to Java 11. Our Closure compiler still needs Java 8 though

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -4,7 +4,7 @@ on: [push, pull_request]
 
 jobs:
   check:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


qidonna (approved):

ChrisAntaki:
Thanks, @qidonna ! Ubuntu will support 20.04 until April 2025 (cc @jpettitt)

jpettitt:
Good call, two questions:

1. Would it be better to pin to lts whatever version that is?
2. Can we document this someplace other than this PR so that whoever is maintaining this has some record of what it's pinned to a specific version?

ChrisAntaki:
> Would it be better to pin to lts whatever version that is?

This PR pins the actions to 20.04 LTS :+1: 

> Can we document this someplace other than this PR so that whoever is maintaining this has some record of what it's pinned to a specific version?

I'll add comments to the `ci.yaml` in a follow-up PR

PR #1520, opened 2021-03-02
Improve fetch error messaging

jpettitt:
Catch the fetch error and hint to the publisher what might be the root cause of the spurious the "CORS" error.  Tracked internally b/175667625

The error display as `Access to fetch at 'https://news.google.com/swg/_/api/v1/publication/norcal-tribune.com/entitlements' from origin 'https://www.norcal-tribune.com' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.`

This is due to the server not finding the domain, returning 404 without the `Access-Control-Allow-Origin` header and the browser throwing an error on the result rather than returning a result from the `fetch()`.  The PR adds hit text as to the real cause.

### DIFF
--- a/src/utils/xhr.js
+++ b/src/utils/xhr.js
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import {assert} from './log';
+import {assert, log} from './log';
 import {parseJson} from './json';
 import {parseUrl} from './url';
 import {utf8EncodeSync} from './bytes';
@@ -103,7 +103,9 @@ export class Xhr {
             reason && reason.message
           );
         }
-      )
+      ).catch(e => {
+        log('[Subscriptions]', 'Fetch failed, a CORS error above may indicate that this domain is not registered for use with Subscribe with Google');
+      })
       .then((response) => assertSuccess(response));
   }
 }
--- a/src/utils/xhr.js
+++ b/src/utils/xhr.js
@@ -91,21 +91,22 @@ export class Xhr {
    * @return {!Promise<!FetchResponse>}
    */
   fetch(input, init) {
-    // TODO (avimehta): Figure out if CORS needs be handled the way AMP does it.
     init = setupInit(init);
     return this.fetch_(input, init)
-      .then(
-        (response) => response,
-        (reason) => {
+      .catch(reason => {
+        /*
+         * If the domain is not valid for SwG we return 404 without
+         * CORS headers and the browser throws a CORS an error.
+         * We include some helpful text in the mssage to point the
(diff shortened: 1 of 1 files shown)


src/utils/xhr.js
```diff
@@ -91,19 +91,21 @@ export class Xhr {
...
-          throw new Error(
-            `XHR Failed fetching (${targetOrigin}/...):`,
-            reason && reason.message
-          );
-        }
-      )
+      .catch((reason) => {
+        /*
+         * If the domain is not valid for SwG we return 404 without
+         * CORS headers and the browser throws a CORS an error.
```
ChrisAntaki, line 99:
```suggestion
         * CORS headers and the browser throws a CORS error.
```

src/utils/xhr.js
```diff
@@ -91,19 +91,21 @@ export class Xhr {
...
+         * We include some helpful text in the mssage to point the
```
ChrisAntaki, line 100:
```suggestion
         * We include some helpful text in the message to point the
```

ChrisAntaki (approved):

ChrisAntaki:
Seems helpful!

src/utils/xhr.js
```diff
@@ -91,19 +91,21 @@ export class Xhr {
...
-      )
+      .catch((reason) => {
+        /*
+         * If the domain is not valid for SwG we return 404 without
+         * CORS headers and the browser throws a CORS error.
+         * We include some helpful text in the message to point the
+         * publisher towards the real problem.
+         */
+        const targetOrigin = parseUrl(input).origin;
+        throw new Error(
```
ed-bird, line 104: Thanks for this!

Do you think we could add a `warn` like we do [here](https://github.com/subscriptions-project/swg-js/blob/main/src/runtime/meter-toast-api.js#L114-L120)?

We could say something like:
```
[swg.js] To fix the CORS error, please double check that ${productId} is configured correctly on https://publishercenter.google.com/ and that you have added ${currentOrigin} to your allow list.
```

jpettitt: The problem is at the `fetch()` level we have no context on what exactly the request was so a more specific message is not possible to do cleanly.    Also xhr is used to fetch from publisher endpoints (AMP) and send analytics events to propensity so it's not 100% that a cors error is a publisher center config, just highly likely.

ed-bird: Ok, makes sense - thanks!

PR #1536, opened 2021-03-09
Persist SwgUserToken in local storage after buyflow completes

qidonna:


src/runtime/pay-flow.js
```diff
@@ -282,6 +287,11 @@ export class PayCompleteFlow {
       this.deps_
         .entitlementsManager()
         .pushNextEntitlements(response.entitlements.raw);
+      // Persist swgUserToken in local storage
+      if (response.swgUserToken) {
+        this.storage_ = new Storage(this.win_);
```
stellachui, line 292: You can access storage from deps; no need to instantiate your own here.

src/runtime/pay-flow.js
```diff
@@ -282,6 +287,11 @@ export class PayCompleteFlow {
...
+        this.storage_.set(Constants.USER_TOKEN, response.swgUserToken, true);
```
stellachui, line 293: Up to you whether you want to do this now or later (when you add more sources for setting the user token, like from the entitlements response), but it might be good to have a shared helper file with functions that get and set the user token. As is, the logic for storing/retrieving the token from local storage using Constants.USER_TOKEN will be spread across different files; it would be nice if that logic was consolidated, even though it's relatively minimal.

src/runtime/pay-flow-test.js
```diff
@@ -611,6 +611,58 @@ describes.realWin('PayCompleteFlow', {}, (env) => {
...
+          idToken: 'ID_TOK',
+          productType: ProductType.SUBSCRIPTION,
+          isSubscriptionUpdate: false,
+          isOneTime: false,
+        }
+      )
+      .returns(Promise.resolve(port));
+    await flow.start(response);
+    expect(PayCompleteFlow.waitingForPayClient_).to.be.true;
+  });
```
stellachui, line 664: Could you add a check to verify the token was stored?

qidonna: Done

src/runtime/pay-flow.js
qidonna: I can add that later. User token will be get in one place (call getEntitelments) and set in two places (SubscribeResponse and EntitlementResponse).

stellachui (approved):

alin04 (approved):

PR #1538, opened 2021-03-10
Move ClientConfigManager instantiation to ConfiguredRuntime for access in Pay*Flow/other dependencies.

stellachui:


alin04 (commented):
Let's hold off on this for now so we don't add an unnecessary call for ConfiguredRuntime. I'll work around the null ClientConfigManager by falling back to a default value.

alin04 (approved):

PR #1542, opened 2021-03-11
Base64Url encode the metering params

elijahsoria:
Switches from using regular base64 encoding (`btoa`) for the metering params in the getEntitlements request to Base64 URL encoding. Also encodes the string with UTF8 safe encodings to support non-Latin characters. Updates tests.


ChrisAntaki (approved):
Thanks!

PR #1543, opened 2021-03-11
Base64Url: Trims `=` padding characters when encoding, and leaves them when decoding

ChrisAntaki:
Note: I think the decoding of `.` characters to `=` characters was just a mistake. The Base64Url spec doesn't call for it. The `base64UrlDecodeToBytes` is only used in one production file, `jwt.js`. That file splits a JWT string on `.` characters, then decodes two segments of the JWT with `base64UrlDecodeToBytes`. So in production, the `base64UrlDecodeToBytes` would never encounter `.` characters

https://github.com/subscriptions-project/swg-js/blob/18a0332258687a3e2684b148d1688918c97c7195/src/utils/jwt.js#L62-L67


elijahsoria:
Do you know why we were decoding a `.` to an `=` in the first place? Base64 doesn't encode to `.` I believe, so they shouldn't be in a proper base64 encoded object.

Could you update the PR description with the answer?

ChrisAntaki:
Sure, I just added a note to the description. Basically, the `base64UrlDecodeToBytes` wasn't ever encountering `.` characters in production

elijahsoria (approved):

PR #1552, opened 2021-03-16
Simplifies setup

ChrisAntaki:
- Installs latest Nodejs dependencies automatically (except on CI systems since it'd be redundant)
- Removes need to manually install/update Yarn (uses `npx yarn` instead)
- Removes instructions to install Java 1.8
- Updates comments


stellachui (approved):
LGTM. Does the internal g3doc also need to be updated?

ChrisAntaki:
> LGTM. Does the internal g3doc also need to be updated?

Yep! Just sent it your way

PR #1556, opened 2021-03-16
Let getEntitlements API take swgUserToken

qidonna:


stellachui (approved):

src/runtime/entitlements-manager.js
```diff
@@ -600,6 +600,14 @@ export class EntitlementsManager {
             'crypt=' +
               encodeURIComponent(params.encryption.encryptedDocumentKey)
           );
+
+          // Add swgUserToken param.
+          if (params.encryption.swgUserToken) {
+            urlParams.push(
+              'swgUserToken=' +
```
alin04, line 607: Is there any guidance on query parameter names like this (which hint at user identification)? Should it be more obfuscated?

qidonna: Discussed offline. We will use param 'sut'. BE will be updated too.

PR #1559, opened 2021-03-17
Persist swgUserToken in entitlements response

qidonna:


src/runtime/entitlements-manager.js
```diff
@@ -375,6 +376,7 @@ export class EntitlementsManager {
     }
     const signedData = json['signedEntitlements'];
     const decryptedDocumentKey = json['decryptedDocumentKey'];
+    const swgUserToken = json['swgUserToken'];
```
stellachui, line 379: Will this be sent back in the entitlement response as sut or swgUserToken? Should we keep the naming consistent across the board, even in the js code?

src/runtime/entitlements-manager.js
```diff
@@ -383,11 +385,13 @@ export class EntitlementsManager {
         decryptedDocumentKey
       );
       if (entitlements) {
+        this.saveSwgUserToken_(swgUserToken);
         return entitlements;
       }
     } else {
       const plainEntitlements = json['entitlements'];
       if (plainEntitlements) {
+        this.saveSwgUserToken_(json['swgUserToken']);
```
stellachui, line 394: You can use your local swgUserToken here instead of reading from the json object again.

src/runtime/entitlements-manager.js
```diff
@@ -400,6 +404,17 @@ export class EntitlementsManager {
     return this.createEntitlements_('', [], isReadyToPay);
   }
 
+  /**
+   * Persist swgUserToken in local storage if entitlements and swgUserToken exist
+   * @param {string} swgUserToken
```
stellachui, line 409: I think swgUserToken can also be undefined here.

src/runtime/entitlements-manager.js
qidonna: sut is the entitlement request url param. swgUserToken is in the entitlement response from boq.

src/runtime/entitlements-manager.js
qidonna: Good catch! I forgot to change it back after debugging

src/runtime/entitlements-manager.js
qidonna: Added ?

stellachui: I think you want |undefined. '?' indicates it's nullable.

stellachui (approved):

PR #1560, opened 2021-03-18
Route to the new subscriptionoffersiframe in OffersFlow if useUpdatedOfferFlows in the ClientConfig is true.

stellachui:


src/runtime/client-config-manager-test.js
```diff
@@ -184,4 +184,30 @@ describes.realWin('ClientConfigManager', {}, () => {
...
     expect(clientConfig.paySwgVersion).to.equal('2');
   });
+
+  it('getClientConfig should have useUpdatedOfferFlows after fetch', async () => {
+    const expectedUrl =
+      '$frontend$/swg/_/api/v1/publication/pubId/clientconfiguration';
+    fetcherMock
+      .expects('fetchCredentialedJson')
+      .withExactArgs(expectedUrl)
+      .returns(Promise.resolve({useUpdatedOfferFlows: true}))
```
ChrisAntaki, line 194: Instead of `.returns(Promise.resolve(` could you use `.resolves(` ?

src/runtime/offers-flow-test.js
```diff
@@ -99,6 +100,41 @@ describes.realWin('OffersFlow', {}, (env) => {
     await offersFlow.start();
   });
 
+  it('should have valid OffersFlow constructed, routed to the new offers iframe', async () => {
+    sandbox
+      .stub(runtime.clientConfigManager(), 'getClientConfig')
+      .callsFake(() => {
+        return Promise.resolve(
```
ChrisAntaki, line 107: Could `.callsFake(() => { return Promise.resolve(` be replaced with `.resolves(` ?

src/runtime/client-config-manager-test.js
stellachui: Thanks for the tip, updated.

src/runtime/offers-flow-test.js
stellachui: Ditto.

ChrisAntaki (approved):

qidonna (approved):

PR #1564, opened 2021-03-18
Page config parser: Improves typing

ChrisAntaki:
This PR improves the typing for an array of `JsonObject`s


src/model/page-config-resolver.js
```diff
@@ -266,17 +266,18 @@ class JsonLdParser {
...
+      if (!this.checkType_.checkValue(config['@type'], ALLOWED_TYPES)) {
         continue;
       }
 
       // Must have a isPartOf[@type=Product].
       let productId = null;
-      const partOfArray = this.valueArray_(possibleConfig, 'isPartOf');
+      const partOfArray = this.valueArray_(config, 'isPartOf');
       if (partOfArray) {
         for (let i = 0; i < partOfArray.length; i++) {
```
elijahsoria, line 282: Do you want to use a different for loop for the overload of the `i` variable here? Or maybe just use a `j` instead?

ChrisAntaki: Renaming to `j` sounds good. Just a note, the inner loop's `i` variable wasn't rewriting the outer loop's `i`

ChrisAntaki: The for-of loop would have rocked, but we currently can't use those in JS that gets compiled into binaries :sob: . The other loop suggestions seemed like downgrades to me, in terms of readability and/or efficiency

elijahsoria (approved):

PR #1571, opened 2021-03-22
Add Metering Params to Pingback

elijahsoria:


src/runtime/entitlements-manager-test.js
```diff
@@ -1003,6 +1008,33 @@ describes.realWin('EntitlementsManager', {}, (env) => {
...
+        EntitlementSource.GOOGLE_SHOWCASE_METERING_SERVICE,
+        EntitlementResult.UNLOCKED_METER,
+        'token1',
+        GOOGLE_METERING_SOURCE,
+        null,
+        ENTITLEMENTS_URL + '?encodedParams=3ncod3dM3t3ringParams'
+      );
+      expectLog(AnalyticsEvent.EVENT_UNLOCKED_BY_METER, false);
+      manager.encodedParams_ = '3ncod3dM3t3ringParams';
+      return manager.consumeMeter_(ents);
```
ChrisAntaki, line 1035:
```suggestion
      await manager.consumeMeter_(ents);
```

If this is equivalent, I think it'd be a bit clearer

src/runtime/entitlements-manager.js
```diff
@@ -244,26 +252,50 @@ export class EntitlementsManager {
...
+    optionalToken = '',
+    optionalIsUserRegistered = null
   ) {
     const message = new EntitlementsRequest();
     message.setUsedEntitlement(usedEntitlement);
     message.setClientEventTime(toTimestamp(Date.now()));
     message.setEntitlementResult(entitlementResult);
     message.setEntitlementSource(entitlementSource);
+    message.setToken(optionalToken);
+    if (optionalIsUserRegistered !== null) {
```
ChrisAntaki, line 288:
```suggestion
    if (typeof optionalIsUserRegistered === 'boolean') {
```

This might be better, it'd check that `optionalIsUserRegistered` is either `true` or `false`

src/runtime/entitlements-manager.js
```diff
@@ -244,26 +252,50 @@ export class EntitlementsManager {
...
-    this.postEntitlementsRequest_(new EntitlementJwt(), result, source);
+    const token = parseQueryString(this.win_.location.search)['gaa_n'];
+    let isUserRegistered = null;
+    if (
+      'additionalParameters' in event &&
+      event.additionalParameters !== null &&
+      event.additionalParameters.getIsUserRegistered() !== null
+    ) {
+      isUserRegistered = event.additionalParameters.getIsUserRegistered();
+    }
```
ChrisAntaki, line 263:
```suggestion
    const isUserRegistered = event?.additionalParameters?.getIsUserRegistered?.();
```

The optional chaining operator can simplify this
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

src/runtime/entitlements-manager.js
```diff
@@ -88,6 +88,9 @@ export class EntitlementsManager {
     /** @private {boolean} */
     this.blockNextNotification_ = false;
 
+    /** @private {?string} */
```
ChrisAntaki, line 91: Could you add a comment that mentions currently this is used for metering params but we might expand it in the future?

src/runtime/entitlements-manager.js
```diff
@@ -199,19 +202,24 @@ export class EntitlementsManager {
       .eventManager()
       .logSwgEvent(AnalyticsEvent.EVENT_UNLOCKED_BY_METER, false);
 
+    const token = parseQueryString(this.win_.location.search)['gaa_n'];
```
ChrisAntaki, line 205: I noticed this is repeated on line 255. Would it make sense to make this a private method, like `getToken_()`?

src/runtime/entitlements-manager.js
```diff
@@ -244,26 +252,50 @@ export class EntitlementsManager {
...
+      message.setIsUserRegistered(optionalIsUserRegistered);
+    }
 
-    const url =
+    let url =
       '/publication/' +
       encodeURIComponent(this.publicationId_) +
       '/entitlements';
+    if (this.encodedParams_) {
+      url += '?encodedParams=' + this.encodedParams_;
```
ChrisAntaki, line 297: Could we use the `addQueryParam()` method from `url.js` here?

src/runtime/entitlements-manager-test.js
```diff
@@ -1115,7 +1147,13 @@ describes.realWin('EntitlementsManager', {}, (env) => {
       params = getParams(null)
     ) {
       const result = analyticsEventToEntitlementResult(event);
-      expectEntitlementPingback(expectedSource, result);
+      expectEntitlementPingback(
+        expectedSource,
+        result,
+        null,
```
ChrisAntaki, line 1153: Could you add comments to describe what these `null` args are for?

src/runtime/entitlements-manager.js
ChrisAntaki: I just sent you a PR that updates the rest of this script to use `addQueryParam()` https://github.com/subscriptions-project/swg-js/pull/1573

src/runtime/entitlements-manager-test.js
elijahsoria: Done.

src/runtime/entitlements-manager.js
elijahsoria: Thanks for the PR!

src/runtime/entitlements-manager.js
elijahsoria: Added a getGaaToken_ function.

src/runtime/entitlements-manager.js
elijahsoria: Done.

ChrisAntaki (approved):

PR #1573, opened 2021-03-23
Entitlements Manager: Uses addQueryParam when possible

ChrisAntaki:
You suggested using this method in another PR, and it seems like it would work well here too


elijahsoria (approved):
Thanks for the clean-up!

qidonna (approved):

PR #1577, opened 2021-03-24
Bind PayClient.handleResponse_ in ctor

elijahsoria:
Binds PayClient.handleResponse_ in ctor so that redirect case has a function to call.


alin04 (changes requested):
Unit test?

We'll want to verify this works before releasing to prod.

elijahsoria:
> Unit test?
> 
> We'll want to verify this works before releasing to prod.

I added a small check for the bound-ness of the variable. The functionality gets tested in the different start() tests.

Since we can't complete the pay flow from local enviornment, I don't think we'll be able to confirm a fix without pushing this change to qual/prod. Since this change is small and there is a user facing bug associated to it, do you think we should release it today in our prod release and test in prod?

alin04:
> > Unit test?
> > We'll want to verify this works before releasing to prod.
> 
> I added a small check for the bound-ness of the variable. The functionality gets tested in the different start() tests.
> 
> Since we can't complete the pay flow from local enviornment, I don't think we'll be able to confirm a fix without pushing this change to qual/prod. Since this change is small and there is a user facing bug associated to it, do you think we should release it today in our prod release and test in prod?

Can we isolate to only qual? We don't want to introduce any regression to non-redirect flow.

elijahsoria:
> Can we isolate to only qual? We don't want to introduce any regression to non-redirect flow.

Yes we can by pushing this change to only the qual environment. Since it is release day, if we want to have time to test in qual we should submit this change after this week's release is done and then push a new release to qual between weekly releases (tomorrow). Does that sound good?

src/runtime/pay-client-test.js
```diff
@@ -110,6 +110,9 @@ describes.realWin('PayClient', {}, (env) => {
   it('should initialize correctly', () => {
     expect(payClient.getType()).to.equal('PAYJS');
     expect(redirectVerifierHelperStubs.prepare).to.be.calledOnce;
+    const expectedResponse = Promise.resolve(INTEGR_DATA_OBJ);
+    payClient.handleResponse_(expectedResponse);
```
alin04, line 114: Don't call this directly.  We want to test that this automatically gets called because it is bound correctly.
Possibly do something like on line 167?

elijahsoria: Discussed offline. Added tests to both experiment and non-experiment flow.

PR #1589, opened 2021-03-31
Initial Implementation of Google Analytics Event Listener

elijahsoria:
Initial Implementation of Google Analytics Event Listener. Includes...

- Adding a mapping of AnalyticsEvents to Google Analytics objects.
- GoogleAnalyticsEventListener class with an event listener that logs to GA if the ga function is available.
- Unit tests.


src/runtime/google-analytics-event-listener.js
```diff
@@ -0,0 +1,59 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS-IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {analyticsEventToGoogleAnalyticsEvent} from './event-type-mapping';
+import {warn} from '../utils/log';
+3;
```
stellachui, line 19: Is this here intentionally?

src/runtime/google-analytics-event-listener.js
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  /**
+   *  Listens for new events from the events manager and logs appropriate events to Google Analytics.
+   * @param {!../api/client-event-manager-api.ClientEvent} event
+   */
+  handleClientEvent_(event) {
+    // Bail immediately if ga function doesn't exist in Window.
+    if (typeof this.win_.ga != 'function') {
+      warn(
```
stellachui, line 49: I'm worried this will be noisy for folks who do have/want GA set up.

src/runtime/event-type-mapping-test.js
```diff
@@ -112,3 +113,23 @@ describes.realWin('analyticsEventToEntitlementResult', {}, () => {
...
+      const result = analyticsEventToGoogleAnalyticsEvent(
+        AnalyticsEvent[event]
+      );
+      // Not all analytics events are mapped
+      if (result === undefined) {
+        continue;
+      }
+      expect(typeof result).to.be.equal('object');
+      const resultString = JSON.stringify(result);
+      // Each EntitlementResult should only be mapped to once
```
stellachui, line 130: s/EntitlementResult/GoogleAnalyticsEvent/?

stellachui (approved):
Thanks, Elijah!

src/runtime/google-analytics-event-listener.js
elijahsoria: No, good catch!

src/runtime/event-type-mapping-test.js
elijahsoria: Done.

src/runtime/google-analytics-event-listener.js
elijahsoria: Good point. Removing the call to warn.

PR #1596, opened 2021-04-07
Create generic static logging functions

mborof:
Hey Chris, before I put too much effort into this solution I thought I would validate it will work; I suspect there's a lot more going on there than I realize but basically when that render/show call is made we need a way to get back to event manager which is in whatever frame is running swg-js.  Is that the frame this code is running in?

### DIFF
--- a/src/utils/log.js
+++ b/src/utils/log.js
@@ -134,14 +134,16 @@ export function logPublisherEvent(event, isFromUserAction, params) {
 }
 
 // Setup code for logging helpers
-let _eventManagerPromiseResolver = null;
-const _eventManagerPromise = new Promise(resolver => _eventManagerPromiseResolver = resolver);
+let eventManagerPromiseResolver = null;
+const eventManagerPromise = new Promise(resolver => eventManagerPromiseResolver = resolver);
 
 /**
+ * This function is automatically called by the swg-js runtime to set up 
+ * the analytics logging functions.
  * @param {!../api/client-event-manager-api.ClientEventManagerApi} eventManager
  */
 export function setStaticEventManager(eventManager) {
-  _eventManagerPromiseResolver(eventManager);
+  eventManagerPromiseResolver(eventManager);
 }
 
 /**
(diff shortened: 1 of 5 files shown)


ChrisAntaki:
Yep, it's the same frame, thanks for checking!

mborof:
I created standard logging functions so it should be very easy to add logging statements to the gaa code later when we're ready.

src/utils/log.js
```diff
@@ -107,3 +109,50 @@ function toString(val) {
...
+export function logPublisherEvent(event, isFromUserAction, params) {
+  return logAnalyticsEvent(event, EventOriginator.PUBLISHER_CLIENT, isFromUserAction, params);
+}
+
+// Setup code for logging helpers
+let _eventManagerPromiseResolver = null;
+const _eventManagerPromise = new Promise(resolver => _eventManagerPromiseResolver = resolver);
+
+/**
+ * @param {!../api/client-event-manager-api.ClientEventManagerApi} eventManager
```
ChrisAntaki, line 141: Could you describe this method?

src/utils/log.js
```diff
@@ -107,3 +109,50 @@ function toString(val) {
...
+/**
+ * Logs an analytics event
+ * @param {!../proto/api_messages.AnalyticsEvent} eventType
+ * @param {!EventOriginator} eventOriginator 
+ * @param {boolean=} isFromUserAction 
+ * @param {!../proto/api_messages.EventParams=} additionalParameters 
+ * @return {!Promise}
+ */
+async function logAnalyticsEvent(eventType, eventOriginator, isFromUserAction, additionalParameters) {
+  const manager = await _eventManagerPromise;
```
ChrisAntaki, line 156: We can't use async/await in production code yet, unfortunately!

src/utils/log.js
```diff
@@ -107,3 +109,50 @@ function toString(val) {
...
+ * @param {boolean=} isFromUserAction 
+ * @param {!../proto/api_messages.EventParams=} params 
+ * @return {!Promise}
+ */
+export function logPublisherEvent(event, isFromUserAction, params) {
+  return logAnalyticsEvent(event, EventOriginator.PUBLISHER_CLIENT, isFromUserAction, params);
+}
+
+// Setup code for logging helpers
+let _eventManagerPromiseResolver = null;
```
ChrisAntaki, line 137: Could you remove the leading "_" from these local variable names?

ChrisAntaki:
Just curious, were you planning to have swg-gaa.js import `logSwgEvent` and call it?

mborof:
Yes, I'm back out on leave though

ChrisAntaki:
Since `swg.js` and `swg-gaa.js` are different JavaScript binaries, if they both import `logSwgEvent`, they'll end up with different instances of it

ChrisAntaki:
Hey! I'm working on a different approach, will close this PR for now just to clean up the queue

PR #1620, opened 2021-04-20
Initialize Pay Client in ctor upon redirect.

elijahsoria:
Initializes Pay Client in ctor upon redirect. Hides changes behind an experiment to ensure it works in production. Adds unit tests.

### DIFF
--- a/src/runtime/pay-client.js
+++ b/src/runtime/pay-client.js
@@ -79,16 +79,21 @@ export class PayClient {
     /** @private @const {!Preconnect} */
     this.preconnect_ = new Preconnect(this.win_.document);
 
+    // If the page is started from a redirect, immediately initialize
+    // client to avoid dropping user state.
+    if (
+      isExperimentOn(this.win_, ExperimentFlags.PAY_CLIENT_REDIRECT) &&
+      this.pageIsInitializedFromPayRedirect_()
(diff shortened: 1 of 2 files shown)


src/runtime/pay-client.js
```diff
@@ -121,6 +126,41 @@ export class PayClient {
...
+    );
+  }
+
+  /**
+   * Detects if the window is started from a Pay redirect by
+   * checking window's hash for Web Activities information.
+   */
+  pageIsInitializedFromPayRedirect_() {
+    const hash = this.win_.location.hash;
+    const hasWebActivitiesResponse = /^#__WA_RES__=/.test(hash);
```
alin04, line 152: I am not a fan of referencing the internals of another library (namely web activities).  Philosophically,  this is brittle in case the library changes their implementation.  Presence of 'redirectEncryptedCallbackdata' should suffice.

ChrisAntaki (approved):

src/runtime/pay-client.js
elijahsoria: Removed the check of WA_RES. Relying on `redirectEncryptedCallbackdata` and `swgRequest` only.

alin04 (approved):

PR #1621, opened 2021-04-21
Handle AlreadySubscribedResponse in swg-basic.js. It will open check …

qidonna:
…entitlements view and let users to login Google


src/api/basic-subscriptions.js
```diff
@@ -54,6 +54,11 @@ export class BasicSubscriptions {
    */
   setOnPaymentResponse(callback) {}
 
+  /**
+   * @return {?}
```
ChrisAntaki, line 58: Could you add a comment for this method?

ChrisAntaki (approved):

PR #1636, opened 2021-04-29
Include publiationid as a http param in the url to open check entitle…

qidonna:
…ments view because CheckEntitlementsViewAction needs it


src/runtime/services.js
```diff
@@ -55,9 +55,10 @@ export function adsUrl(url) {
 /**
  * @param {string} url Relative URL, e.g. "/offersiframe".
  * @param {string=} prefix
+ * @param {string=} param Extra param to append to the URL.
```
ChrisAntaki, line 58: This is a cool idea. In the future, I could imagine people needing multiple params. Could you update this so it accepts an object with 1+ params?

ChrisAntaki (approved):

PR #1638, opened 2021-04-29
Create dummy API of AudienceActions.

hot2346:
Create an entrypoint of the API of RevenueActions.

### DIFF
--- a/src/runtime/basic-runtime-test.js
+++ b/src/runtime/basic-runtime-test.js
@@ -347,7 +347,7 @@ describes.realWin('BasicRuntime', {}, (env) => {
 
       expect(openStub).to.be.calledOnceWithExactly(
         'CHECK_ENTITLEMENTS',
-        'https://news.google.com/swg/_/ui/v1/checkentitlements?_=_',
+        'https://news.google.com/swg/_/ui/v1/checkentitlements?_=_&publicationId=pub1',
         '_blank',
         {publicationId: 'pub1', _client: 'SwG $internalRuntimeVersion$'},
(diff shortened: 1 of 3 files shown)


src/runtime/runtime.js
```diff
@@ -515,6 +515,13 @@ export class Runtime {
       )
     );
   }
+
+  /** @override */
+  showBestAudienceAction() {
+    console.error('Forbidden call, under implementation.');
```
hartmannr76, line 521: Maybe use `debugLog` for these for now. Linter doesn't seem to like the `console` messaging

src/api/subscriptions.js
```diff
@@ -325,6 +325,15 @@ export class Subscriptions {
...
    */
   consumeShowcaseEntitlementJwt(showcaseEntitlementJwt, onCloseDialog) {}
+
+  /**
+   * Intelligently returns the most interesting action to the
+   * reader based on different different user status. For
+   * instance, a new user may get free metering by simply
+   * clicking 'follow-publisher' action, and a frequently
+   * visiting user may be shown a 'creating an account' action.
+   */
```
hartmannr76, line 335: Include the return type in the comment `* @return {?}`

src/runtime/runtime.js
ChrisAntaki: `warn` (from "src/utils/log.js") might be even better

src/api/subscriptions.js
hot2346: Will fix in next commit.

PR #1646, opened 2021-05-03
onResult entitlements in basic-runtime init()

qidonna:
This is to prevent from not setting onResult if the port loses connection with host after opening checkentitlements view.


src/runtime/basic-runtime.js
```diff
@@ -244,6 +245,13 @@ export class BasicRuntime {
   setupButtons() {
     return this.configured_(false).then((runtime) => runtime.setupButtons());
   }
+
+  // Process result from checkentitlements view
```
ChrisAntaki, line 249: Could you make this a JSDoc comment? (/** ...)

src/runtime/basic-runtime.js
```diff
@@ -421,6 +424,36 @@ export class ConfiguredBasicRuntime {
     });
   }
 
+  processEntitlements() {
```
ChrisAntaki, line 427: Could you add a JSDoc comment for this method?

ChrisAntaki (approved):

src/runtime/basic-runtime.js
qidonna: Done

PR #1668, opened 2021-05-07
Showcase: Validate timestamps.

ChrisAntaki:


src/runtime/entitlements-manager.js
```diff
@@ -697,25 +692,41 @@ export class EntitlementsManager {
...
+
+            const attributeNames = Object.keys(attributes);
+            attributeNames.forEach((attributeName) => {
+              const name = `${category}_${attributeName}`;
+              const timestamp = attributes[attributeName].timestamp;
+
+              // Validate timestamp.
+              const timestampIsTooFarInTheFuture =
+                timestamp > (Date.now() / 1000) * 2;
+              if (!timestamp || timestampIsTooFarInTheFuture) {
```
ed-bird, line 709: Can we add another check if the timestamp is actually an integer instead of something else like "Friday, 7 May 2021 18:16:51"?

Could be a different PR if that's better.

Thanks!

ed-bird (approved):
Thanks @ChrisAntaki !

src/runtime/entitlements-manager.js
ChrisAntaki: Good call, updated!

PR #1677, opened 2021-05-10
Showcase: Bail if `gaa_at` = 'na'

ChrisAntaki:
Some publishers use GAA params for analytics. This lets a publisher include GAA params without fetching metering entitlements


ed-bird (commented):
Looks great - thanks!

I have just noticed that this function is only checked during the `GaaMeteringRegwall.show()` call. For full coverage, I imagine we'd want something in the [entitlements manager](https://github.com/subscriptions-project/swg-js/blob/main/src/runtime/entitlements-manager.js) too?

ChrisAntaki:
The entitlements manager imports the `queryStringHasFreshGaaParams` method, right?

ed-bird:
> The entitlements manager imports the `queryStringHasFreshGaaParams` method, right?

Awesome, thanks!

ed-bird (approved):

PR #1688, opened 2021-05-14
Do not trigger auto prompt if UI predicate is false

qidonna:


src/runtime/auto-prompt-manager.js
```diff
@@ -129,19 +129,19 @@ export class AutoPromptManager {
           autoPromptType: params.autoPromptType,
           callback: params.displayForLockedContentFn,
         });
-      }, (autoPromptConfig?.clientDisplayTrigger.displayDelaySeconds || 0) * SECOND_IN_MILLIS);
+      }, (clientConfig.autoPromptConfig?.clientDisplayTrigger.displayDelaySeconds || 0) * SECOND_IN_MILLIS);
```
ChrisAntaki, line 132: Could `clientConfig` be `undefined`? If so, the chaining operator should be placed right after it
```suggestion
      }, (clientConfig?.autoPromptConfig.clientDisplayTrigger.displayDelaySeconds || 0) * SECOND_IN_MILLIS);
```

ChrisAntaki:
Could you write tests that cover the new lines of code?

ChrisAntaki:
Looks like an existing line lost coverage
https://app.codecov.io/gh/subscriptions-project/swg-js/compare/1688/changes

image

ChrisAntaki (approved):

ChrisAntaki:
Looks good!

PR #1691, opened 2021-05-17
Close contribution offers iframe if no entitlement found for the sign…

qidonna:
…ed in user

### DIFF
--- a/src/runtime/basic-runtime.js
+++ b/src/runtime/basic-runtime.js
@@ -465,6 +465,7 @@ export class ConfiguredBasicRuntime {
           this.configuredClassicRuntime_.getLastOffersFlow();
         if (lastOffersFlow) {
           lastOffersFlow.showNoEntitlementFoundToast();
+          return;
         }
 
         const lastContributionsFlow =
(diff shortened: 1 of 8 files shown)


src/runtime/basic-runtime.js
```diff
@@ -466,6 +466,12 @@ export class ConfiguredBasicRuntime {
         if (lastOffersFlow) {
           lastOffersFlow.showNoEntitlementFoundToast();
```
ChrisAntaki, line 467:
```suggestion
          lastOffersFlow.showNoEntitlementFoundToast();
          return;
```

src/runtime/basic-runtime.js
```diff
@@ -466,6 +466,12 @@ export class ConfiguredBasicRuntime {
...
         }
+
+        const lastContributionsFlow =
+          this.configuredClassicRuntime_.getLastContributionsFlow();
+        if (lastContributionsFlow) {
+          lastContributionsFlow.showNoEntitlementFoundToast();
```
ChrisAntaki, line 473:
```suggestion
          lastContributionsFlow.showNoEntitlementFoundToast();
          return;
```

src/runtime/basic-runtime.js
ChrisAntaki: Just to prevent multiple calls to `showNoEntitlementFoundToast`

ChrisAntaki (approved):

PR #1693, opened 2021-05-18
UI predicates should be true by default

qidonna:


src/runtime/client-config-manager.js
```diff
@@ -168,8 +168,8 @@ export class ClientConfigManager {
     let uiPredicates = undefined;
     if (uiPredicatesJson) {
       uiPredicates = new UiPredicates(
-        uiPredicatesJson.canDisplayAutoPrompt,
-        uiPredicatesJson.canDisplayButton
+        uiPredicatesJson.canDisplayAutoPrompt || true,
```
ChrisAntaki, line 171:
```suggestion
        uiPredicatesJson.canDisplayAutoPrompt ?? true,
```
I think the `??` is better than `||` here (and on the next line)
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator

ChrisAntaki: > The nullish coalescing operator (??) is a logical operator that returns its right-hand side operand when its left-hand side operand is null or undefined, and otherwise returns its left-hand side operand.

> This can be contrasted with the logical OR (||) operator, which returns the right-hand side operand if the left operand is any falsy value, not only null or undefined. In other words, if you use || to provide some default value to another variable foo, you may encounter unexpected behaviors if you consider some falsy values as usable (e.g., '' or 0)

qidonna: thanks!

ChrisAntaki (approved):

ChrisAntaki:
Just to be safe, could you add a test to verifies these defaults take effect, and don't override `false` values?

qidonna:
close this PR. UI Predicates shouldn't be true by default on the client side. On the server side it is false in prod and true in qual by default.

PR #1695, opened 2021-05-19
Showcase: Creates logShowcaseEvent method

ChrisAntaki:
This PR also deprecates `setShowcaseEntitlement`

### DIFF
--- a/src/api/subscriptions.js
+++ b/src/api/subscriptions.js
@@ -311,7 +311,7 @@ export class Subscriptions {
 
   /**
    * Logs a Showcase event with optional metadata.
-   * Publishers log these events, when they participate in Showcase.
+   * Publishers log these events, as part of participating in Showcase.
    * @param {!ShowcaseEventWithMetadata} event
    */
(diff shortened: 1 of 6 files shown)


src/api/subscriptions.js
```diff
@@ -362,17 +370,24 @@ export const PublisherEntitlementEvent = {
...
+ * }}
+ */
+export let ShowcaseEventWithMetadata;
+
+/**
+ * DEPRECATED
+ * Showcase publishers log these events with optional metadata.
+ * @typedef {{
+ *   entitlement: !ShowcaseEvent,
+ *   isUserRegistered: (boolean|undefined),
```
elijahsoria, line 387: Please add a comment explaining what isUserRegistered and isFromUserAction means or point to other documentation about logging, if available. Not sure if there are other public docs describing these events.

ChrisAntaki: Good call, I added some comments! I couldn't find public docs describing these events

PR #1697, opened 2021-05-20
Showcase: Log event when Regwall renders

ChrisAntaki:
This PR updates the Showcase JS to log an event when the Regwall renders. This PR also renames some internal classes/methods for readability


src/utils/gaa.js
```diff
@@ -694,6 +701,30 @@ function callSwg(callback) {
...
+    // Get reference to event manager.
+    swg.getEventManager().then((eventManager) => {
+      // Get individual analytics events from Showcase event.
+      const eventTypes = showcaseEventToAnalyticsEvents(showcaseEvent);
+
+      // Log each analytics event.
+      eventTypes.forEach((eventType) => {
+        eventManager.logEvent({
+          eventType,
+          eventOriginator: EventOriginator.SWG_CLIENT,
```
ed-bird, line 719: Just double checking in case this should be `EventOriginator.SHOWCASE_CLIENT`?

ed-bird (approved):
One small comment about the event originator, otherwise looks great - thanks @ChrisAntaki !

src/utils/gaa.js
ChrisAntaki: Yep! `src/runtime/entitlements-manager.js` suggests `SHOWCASE_CLIENT` represents publishers

PR #1699, opened 2021-05-21
Async/await

ChrisAntaki:


src/utils/gaa.js
```diff
@@ -481,7 +480,7 @@ export class GaaMeteringRegwall {
    * Returns the GAA user, after the user signs in.
    * @private
    * @nocollapse
-   * @return {!Promise<!GoogleUserDef>}
+   * @return {!Promise<!GaaUserDef>}
```
ChrisAntaki, line 483: Closure compiler might be better at catching type issues when using async/await? Not sure why it missed this before

qidonna (approved):

ChrisAntaki:
Thanks for the review! I'm still working on supporting IE11

ChrisAntaki:
If we drop support for IE11, this will get a lot simpler. I'll go ahead and close this PR

PR #1707, opened 2021-05-24
Showcase: Updates the Regwall's Japanese translations

ChrisAntaki:
- Updates Regwall's Japanese translations
- Updates i18n script to use newer Cheerio API


assets/i18n/strings/de_AT.xlb
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jpettitt, line 1: Is Austria intentionally included in the Japan PR?

ChrisAntaki: Good catch, I'll create a new PR for that

elijahsoria (approved):

PR #1729, opened 2021-06-01
Don't Validate Access Type during Pingback Logging

elijahsoria:
We want to send the pingback on all logging events, regardless to if gaa_at == na.

### DIFF
--- a/src/runtime/entitlements-manager.js
+++ b/src/runtime/entitlements-manager.js
@@ -234,12 +234,11 @@ export class EntitlementsManager {
   possiblyPingbackOnClientEvent_(event) {
     // Verify GAA params are present, otherwise bail since the pingback
     // shouldn't happen on non-metering requests.
-    // We don't validate access type since not all publishers use it and
-    // we want to pingback on all access types.
+    // We don't validate access type since we want to pingback on all access types.
     if (
       !queryStringHasFreshGaaParams(
         this.win_.location.search,
-        /*allowAllAccessTypes=*/ false
+        /*allowAllAccessTypes=*/ true
       )
     ) {
       return;
--- a/src/runtime/entitlements-manager.js
+++ b/src/runtime/entitlements-manager.js
@@ -234,7 +234,13 @@ export class EntitlementsManager {
   possiblyPingbackOnClientEvent_(event) {
     // Verify GAA params are present, otherwise bail since the pingback
     // shouldn't happen on non-metering requests.
-    if (!queryStringHasFreshGaaParams(this.win_.location.search)) {
+    // We don't validate access type since not all publishers use it.
+    if (
+      !queryStringHasFreshGaaParams(
+        this.win_.location.search,
+        /*shouldValidateAccessType=*/ false
+      )
+    ) {
       return;
     }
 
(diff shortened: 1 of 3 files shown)


src/utils/gaa.js
```diff
@@ -292,9 +292,13 @@ export let GoogleUserDef;
 /**
  * Returns true if the query string contains fresh Google Article Access (GAA) params.
  * @param {string} queryString
+ * @param {boolean} shouldValidateAccessType
  * @return {boolean}
  */
-export function queryStringHasFreshGaaParams(queryString) {
+export function queryStringHasFreshGaaParams(
+  queryString,
+  shouldValidateAccessType = true
```
ChrisAntaki, line 300: Could we update this name to `allowAllAccessTypes`? I think that'd be a bit clearer

src/runtime/entitlements-manager.js
```diff
@@ -234,7 +234,13 @@ export class EntitlementsManager {
   possiblyPingbackOnClientEvent_(event) {
     // Verify GAA params are present, otherwise bail since the pingback
     // shouldn't happen on non-metering requests.
-    if (!queryStringHasFreshGaaParams(this.win_.location.search)) {
+    // We don't validate access type since not all publishers use it.
```
ChrisAntaki, line 237: Could this comment read "We always want to send pingbacks" ?

elijahsoria: Added some more info along these lines.

src/utils/gaa.js
elijahsoria: Done

src/utils/gaa.js
```diff
@@ -292,9 +292,13 @@ export let GoogleUserDef;
 /**
  * Returns true if the query string contains fresh Google Article Access (GAA) params.
  * @param {string} queryString
+ * @param {boolean} allowAllAccessTypes
  * @return {boolean}
  */
-export function queryStringHasFreshGaaParams(queryString) {
+export function queryStringHasFreshGaaParams(
+  queryString,
+  allowAllAccessTypes = true
```
ChrisAntaki, line 300: Could we make this default to false now?

src/utils/gaa-test.js
```diff
@@ -108,6 +108,11 @@ describes.realWin('queryStringHasFreshGaaParams', {}, () => {
     const queryString = '?gaa_at=na&gaa_n=n&gaa_sig=sig&gaa_ts=99999';
     expect(queryStringHasFreshGaaParams(queryString)).to.be.false;
   });
+
+  it('succeeds if gaa_at param specifies "no access" but shouldValidateAccessType is false', () => {
```
ChrisAntaki, line 112:
```suggestion
  it('succeeds if gaa_at param specifies "no access" but allowAllAccessTypes is true', () => {
```

elijahsoria: Updated

src/utils/gaa.js
elijahsoria: Yep!

ChrisAntaki:
Thanks!

ChrisAntaki (approved):

PR #1736, opened 2021-06-04
Logs error message when crypto API isn't available

ChrisAntaki:
This PR adds a new error message (`Swgjs only works on secure (HTTPS or localhost) pages.`). Hopefully this lets developers know that they should test Swgjs on either localhost or HTTPS pages

This PR also adds tests to improve coverage

### DIFF
--- a/src/utils/string-test.js
+++ b/src/utils/string-test.js
@@ -200,17 +200,20 @@ describes.realWin('hash', {}, async () => {
     // Spy on console.warn method.
     sandbox.stub(self.console, 'warn');
 
-    // Temporarily remove the crypto API.
+    // Temporarily remove the crypto APIs.
     const crypto = self.crypto;
+    const msCrypto = self.msCrypto;
     delete self.crypto;
+    delete self.msCrypto;
 
     // Verify behavior.
     const message = 'Swgjs only works on secure (HTTPS or localhost) pages.';
     await expect(hash('string1')).to.be.rejectedWith(message);
     expect(self.console.warn).to.have.been.calledWithExactly(message);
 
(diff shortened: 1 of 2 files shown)


src/utils/string.js
```diff
@@ -197,7 +198,14 @@ function toHex(buffer) {
  */
 export function hash(stringToHash) {
   const crypto = self.crypto || self.msCrypto;
-  const subtle = crypto.subtle;
+  const subtle = crypto?.subtle;
```
mmmicedcoffee, line 201: it would've been nice if self.crypto/self.msCrypto were typed properly so the previous code triggered a JS compiler error instead.

src/utils/string-test.js
```diff
@@ -192,4 +195,51 @@ describe('hash', async () => {
     const hash2 = await hash('string2');
     expect(hash1).to.not.equal(hash2);
   });
+
+  it('should throw if subtle crypto API is missing', async () => {
+    // Spy on console.warn method.
+    sandbox.stub(self.console, 'warn');
+
+    // Temporarily remove the crypto API.
```
mmmicedcoffee, line 203: for completeness, should you also remove msCrypto? then we don't have to make changes if the test runs on IE 11

mmmicedcoffee (approved):

src/utils/string-test.js
ChrisAntaki: Good call!

src/utils/string.js
ChrisAntaki: Yep, that's definitely a gap in the type system

PR #1749, opened 2021-06-10
Contributions: Allow for non-closable dialogs

ChrisAntaki:
- Fixes a bug where Contribution flow dialogs were always closable, regardless of the `isClosable` param's value
- Adds a test to catch regressions
- Updates the code inside of the Subscription flow that chooses a default value for the `isClosable` param's value. This code wasn't broken, but it was easy to simplify, so why not

### DIFF
--- a/src/runtime/contributions-flow-test.js
+++ b/src/runtime/contributions-flow-test.js
@@ -61,7 +61,7 @@ describes.realWin('ContributionsFlow', {}, (env) => {
     callbacksMock.verify();
   });
 
-  it('should have valid ContributionsFlow constructed with a list', async () => {
+  it('has valid ContributionsFlow constructed with a list', async () => {
     contributionsFlow = new ContributionsFlow(runtime, {list: 'other'});
     activitiesMock
       .expects('openIframe')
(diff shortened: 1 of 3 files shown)


src/runtime/contributions-flow-test.js
```diff
@@ -83,6 +83,32 @@ describes.realWin('ContributionsFlow', {}, (env) => {
     await contributionsFlow.start();
   });
 
+  it('should allow for non-closable dialogs', async () => {
```
mmmicedcoffee, line 86: optional: I have a slight preference for more concise test descriptions that start directly with the relevant verb rather than "should", so this test description would read "allows non-closable dialogs". 

I don't feel strongly about it though. I noticed the surrounding tests are all written in the "should..." style, so it seems fine to be consistent, too.

mmmicedcoffee (approved):

src/runtime/contributions-flow-test.js
ChrisAntaki: That's definitely clearer! I updated all the surrounding test names

PR #1768, opened 2021-06-22
Disable showing offers based on client config ui predicates

adolgachev:


src/runtime/offers-flow.js
```diff
@@ -225,19 +231,22 @@ export class OffersFlow {
     return Promise.resolve();
   }
 
+  shouldShow_(clientConfig) {
```
mmmicedcoffee, line 234: is it worth naming this method "shouldShowAutoPrompt" for additional clarity?

mmmicedcoffee: also, please add a JSDoc comment so there's type information for the argument.

src/runtime/offers-flow.js
```diff
@@ -225,19 +231,22 @@ export class OffersFlow {
...
+      clientConfig.uiPredicates.canDisplayAutoPrompt
+    );
+  }
+
   /**
    * Gets the URL that should be used for the activity iFrame view.
-   * @param {!Promise<../model/client-config.ClientConfig>} clientConfigPromise
-   * @return {!Promise<string>}
+   * @param {!../model/client-config.ClientConfig} clientConfig
+   * @return {!string}
```
mmmicedcoffee, line 244: primitive types are non-nullable by default, so this type annotation can be simplified to "@return {string}"

src/runtime/offers-flow.js
```diff
@@ -121,17 +121,19 @@ export class OffersFlow {
     this.skus_ = feArgsObj['skus'] || [ALL_SKUS];
 
     /** @private @const {!Promise<!ActivityIframeView>} */
```
mmmicedcoffee, line 123: the type needs to be updated to reflect that the Promise can return null (!Promise<?ActivityIframeView>)

src/runtime/offers-flow.js
adolgachev: Actually intentionally keeping it more generic (just tied to showing of offers in this context)  as just re-using the UI predicates which both return the same thing but eventually may be tweaked and expanded.

mmmicedcoffee: sounds good!

src/runtime/offers-flow.js
```diff
@@ -120,18 +120,20 @@ export class OffersFlow {
...
-    ).then((url) => {
-      return new ActivityIframeView(
-        this.win_,
-        this.activityPorts_,
-        feUrl(url),
-        feArgsObj,
-        /* shouldFadeBody */ true
-      );
-    });
+    /** @private @const {!Promise<!ActivityIframeView|null>} */
```
mmmicedcoffee, line 123: is "!ActivityIframeView|null" different from "?ActivityIframeView"? I thought the "?" indicates "|null".

mmmicedcoffee (approved):

src/runtime/offers-flow.js
adolgachev: I was intending for explicit null (v.s. undefined like typescript) but, it doesn't seem to make a difference here. The docs are more explicit for behavior for params but I'll move to the simpler syntax in any case.

fikus:
Issue reference: b/190849246

PR #1769, opened 2021-06-22
Creates Bash shortcuts

ChrisAntaki:
This PR creates some initial Bash shortcuts. Hopefully these will make Swgjs development easier, especially for engineers newer to Nodejs and/or Git(Hub)

Over time, I hope to improve these shortcuts, and update documentation to reference them

### DIFF
--- a/shortcuts.sh
+++ b/shortcuts.sh
@@ -1,11 +1,53 @@
+###############
 ## Constants ##
+###############
+
 
 AMPHTML_PATH=~/projects/amphtml
 SWGJS_PATH=~/projects/swgjs
 
 
-## Methods ##
 
+####################
+## Public methods ##
+####################
+
+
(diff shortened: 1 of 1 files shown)


mmmicedcoffee (commented):
are you planning to add any documentation to go with this file? I think it would be helpful to get context around when one might use each of these functions. for example, `swgjs_get_github_username` looks like it's an internal helper function for this file -- I don't think people are expected to invoke that function?

ChrisAntaki:
Good call! I added comments for every public method that describe when someone might want to call ​them. I also moved the private methods to their own section, and prefixed their names with `__`

In future PRs, I hope to update the documentation to suggest using these methods

mmmicedcoffee (approved):
nice, thank you!!

PR #1774, opened 2021-06-23
Changelog: Suggest a version number

ChrisAntaki:
This PR updates the changelog command to suggest a version number. Since the previous version is available, this PR just increments that version by 1

This makes the release easier, and will let us simplify our documentation a bit too


jpettitt (approved):

jpettitt:
Nice.

PR #1779, opened 2021-06-25
Demos: Add basic local demos

ChrisAntaki:
These demos will help developers demo their changes locally


ChrisAntaki:
Actually, this one needs some more work

demos/index.js
```diff
@@ -0,0 +1,33 @@
...
+ * Redirects to SwG Basic JS.
+ * The endpoint is compatible with the classic Swgjs local demos.
+ */
+app.get('/examples/sample-pub/redirect-to/swg-basic.js', (req, res) => {
+  res.redirect('https://news.google.com/swg/js/v1/swg-basic.js');
+});
+
+app.listen(port, () => {
+  console /*OK*/
+    .log(`Swgjs Demos are available at http://localhost:${port}`);
```
kristenwang, line 32: Should this line here differentiate SwG classic & SwG basic?

ChrisAntaki: Good call, updated

demos/public/google541407f935830527.html
```diff
@@ -0,0 +1 @@
+google-site-verification: google541407f935830527.html
```
kristenwang, line 1: I'm wondering how does the token here gets generated?

ChrisAntaki: The Google Search Console generates HTML files like these. It helps them verify someone owns a domain or URL path https://support.google.com/webmasters/answer/9008080?hl=en

ChrisAntaki: The HTML file upload method is the easiest verification method for AppEngine sites

kristenwang (approved):
LGTM.

PR #1785, opened 2021-06-29
Showcase: Optionally render CASL blurb

ChrisAntaki:
This PR adds an optional CASL blurb

Additional translations will come in a follow-up PR

### DIFF
--- a/examples/sample-pub/publisher.js
+++ b/examples/sample-pub/publisher.js
@@ -133,6 +133,12 @@ function startFlow(flow, var_args) {
 function startFlowAuto() {
   let flow = (window.location.search || '').split('?')[1] || 'demo';
 
+  // Provide a shortcut to metering params.
+  if (location.search === '?metering') {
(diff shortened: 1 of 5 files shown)


examples/sample-pub/publisher.js
```diff
@@ -133,6 +133,12 @@ function startFlow(flow, var_args) {
 function startFlowAuto() {
   let flow = (window.location.search || '').split('?')[1] || 'demo';
 
+  // Provide a shortcut to metering params.
+  if (location.search === '?metering') {
```
adolgachev, line 137: This isn't consistent with above. If `window.location` is prior convention and prevalent, probably should continue as such. Or change above to match new convention.

adolgachev (approved):

examples/sample-pub/publisher.js
ChrisAntaki: Good catch, fixed !

PR #1801, opened 2021-07-08
Allow scrolling within dialogs, hidden behind an experiment

ChrisAntaki:
This PR creates an experiment that allows users with small screens to scroll down and see important stuff like the "Already a contributor?" button
image

### DIFF
--- a/src/runtime/experiment-flags.js
+++ b/src/runtime/experiment-flags.js
@@ -60,4 +60,9 @@ export const ExperimentFlags = {
    * Experiment flag for guarding changes to fix PayClient redirect flow.
    */
   PAY_CLIENT_REDIRECT: 'pay-client-redirect',
+
+  /**
+   * Enables scrolling within dialogs
+   */
+  SCROLLING_WITHIN_DIALOGS: 'scrolling-within-dialogs',
 };
--- a/src/runtime/experiment-flags.js
+++ b/src/runtime/experiment-flags.js
@@ -62,7 +62,7 @@ export const ExperimentFlags = {
   PAY_CLIENT_REDIRECT: 'pay-client-redirect',
 
   /**
-   * Enables scrolling within dialogs
+   * Enables scrolling within dialogs.
    */
   SCROLLING_WITHIN_DIALOGS: 'scrolling-within-dialogs',
 };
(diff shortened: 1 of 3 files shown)


src/runtime/experiment-flags.js
```diff
@@ -60,4 +60,9 @@ export const ExperimentFlags = {
    * Experiment flag for guarding changes to fix PayClient redirect flow.
    */
   PAY_CLIENT_REDIRECT: 'pay-client-redirect',
+
+  /**
+   * Enables scrolling within dialogs
```
mmmicedcoffee, line 65: nit: missing period at the end of this comment. 

it doesn't seem to be consistently applied in this file, but I've always used punctuation, capitalization, etc, in my JSDoc comments.

src/ui/activity-iframe-view-test.js
```diff
@@ -97,6 +106,30 @@ describes.realWin('ActivityIframeView', {}, (env) => {
       expect(activityIframePort.whenReady).to.have.been.calledOnce;
     });
 
+    it('should disallow scrolling within dialogs', async () => {
```
mmmicedcoffee, line 109: nit: can omit "should" for brevity, so the test description would read something like "disallows scrolling within dialogs"

src/ui/activity-iframe-view-test.js
```diff
@@ -97,6 +106,30 @@ describes.realWin('ActivityIframeView', {}, (env) => {
...
+        win,
+        activityPorts,
+        src,
+        activityArgs
+      ).getElement();
+
+      expect(iframe.scrolling).to.equal('no');
+    });
+
+    it('should (optionally) allow scrolling within dialogs', async () => {
```
mmmicedcoffee, line 120: same here - this could be: "allows scrolling within dialogs behind an experiment"

mmmicedcoffee (approved):

src/runtime/experiment-flags.js
ChrisAntaki: Added!

src/ui/activity-iframe-view-test.js
ChrisAntaki: Good call

src/ui/activity-iframe-view-test.js
ChrisAntaki: That's clearer 👍

PR #1804, opened 2021-07-09
Dialog: Only continue the freshest animations

ChrisAntaki:
This PR fixes an issue with the Dialog, where sometimes its stale animations overwrote its freshest animation


mmmicedcoffee (commented):
is there a way we could test the new animation cancellation behavior?

ChrisAntaki:
> is there a way we could test the new animation cancellation behavior?

Yep, expansion and collapse animations now have tests!

src/components/dialog.js
```diff
@@ -342,41 +348,51 @@ export class Dialog {
...
+              },
+              300,
+              'ease-out'
+            ).then(() => {
+              if (animationNumber === this.animationNumber_) {
+                setImportantStyles(this.getElement(), {
+                  'height': `${newHeight}px`,
+                  'transform': 'translateY(0)',
+                });
+              }
```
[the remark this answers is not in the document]
ChrisAntaki: That's a good suggestion. I went ahead and updated the new code to use early returns where possible. It looks better, and it's helpful to see the `return` line's covered by tests

src/components/dialog-test.js
```diff
@@ -155,38 +155,94 @@ describes.realWin('Dialog', {}, (env) => {
...
+      immediate();
+      await dialog.open();
+      await dialog.openView(view);
+
+      const setStylePropertySpy = sandbox
+        .stub(dialog.getElement().style, 'setProperty')
+        .callThrough();
+
+      dialog.resizeView(view, newHeight1, ANIMATE);
+      await dialog.resizeView(view, newHeight2, ANIMATE);
```
mmmicedcoffee, line 223: should the two new tests also have an expect statement verifying that the height property is set as expected?

src/components/dialog.js
```diff
@@ -342,12 +348,23 @@ export class Dialog {
     }
     const newHeight = this.getMaxAllowedHeight_(height);
 
+    // Uniquely identify this animation.
+    // This lets callbacks abandon stale animations.
+    const animationNumber = ++this.animationNumber_;
+    const isStale = () => {
```
mmmicedcoffee, line 354: I like the addition of the new helper!

mmmicedcoffee (approved):

src/components/dialog-test.js
ChrisAntaki: For sure, added!

PR #1814, opened 2021-07-14
Toast: Send flavors

ChrisAntaki:
This PR updates Swgjs to send what flavor of toast it wants (ex: "classic" or "basic")


src/runtime/basic-runtime.js
```diff
@@ -296,6 +296,7 @@ export class ConfiguredBasicRuntime {
 
     // Do not show toast in swgz.
     this.entitlementsManager().blockNextNotification();
+    this.entitlementsManager().setToastFlavor('basic');
```
alin04, line 299: I think line 461 will ignore the flavor.

alin04 (approved):

PR #1818, opened 2021-07-14
Add AttributionParams to ClientConfig and update ClientConfigManager to populate it

li-cai:


src/runtime/client-config-manager.js
```diff
@@ -173,11 +174,21 @@ export class ClientConfigManager {
       );
     }
 
+    const attributionParamsJson = json['attributionParams'];
+    let attributionParams = undefined;
```
ChrisAntaki, line 178:
```suggestion
    let attributionParams;
```

ChrisAntaki (approved):

PR #1826, opened 2021-07-16
Toast: Updated styles

ChrisAntaki:
This PR updates the Toast's styles

image

image


mmmicedcoffee (approved):
out of curiosity, why are all the styles on .swg-toast marked with "!important"? what rules do they have to override?

ChrisAntaki:
It's just defensive coding, since this CSS applies to `<iframe>` elements on 3p sites

mmmicedcoffee:
ah, I see. that makes sense!

PR #1830, opened 2021-07-20
Improves coverage by adding tests

ChrisAntaki:


src/api/entitlements-test.js
```diff
@@ -84,4 +84,15 @@ describes.realWin('Entitlements', {}, () => {
       );
     });
   });
+
+  describe('consume', () => {
+    it('executes callback', () => {
+      const consumeSpy = sandbox.spy();
```
mmmicedcoffee, line 90: nit: do you want to rename "consumeSpy" to "callbackSpy"?

currently, it was confusing to me at an initial glance since the method under test is also called "consume".

mmmicedcoffee (approved):
Thanks for improving the tests!

src/api/entitlements-test.js
ChrisAntaki: Oh yeah, that was confusing :joy:  - I updated the naming to "consume handler" / "consumeHandlerSpy"

PR #1835, opened 2021-07-21
Explicitly set box-sizing of miniprompt close icon to content-box.

mmmicedcoffee:
The current icon styles were written assuming a content-box box-sizing model, but 3p sites might have a style rule that sets box-sizing to border-box throughout the site (like Scenic does!).

Before:
miniprompt close icon (before)

After:
miniprompt close icon (after)


ChrisAntaki:
Good catch!

ChrisAntaki (approved):

PR #1845, opened 2021-07-27
Enabled alwaysShow to use a passed param

everyplace:
This PR enables the `alwaysShow` parameter to be passed through from the init() function in swg-basic.js. To preserve existing init() snippets, the parameter defaults to `false` if not set, which is the current behavior.

No tests are included in this change, as the param's validity is already being tested.


fhinson:
+1, LGTM

jpettitt (approved):

everyplace:
@jpettitt thanks for the speedy review!

PR #1846, opened 2021-07-27
Showcase: Look for publisher name in Microdata

ChrisAntaki:


src/utils/gaa.js
```diff
@@ -565,8 +589,26 @@ export class GaaMeteringRegwall {
...
+   * @nocollapse
+   * @return {string}
+   */
+  static getPublisherNameFromMicrodataPageConfig_() {
+    const publisherNameElements = self.document.querySelectorAll(
+      '[itemscope][itemtype][itemprop="publisher"] [itemprop="name"]'
+    );
 
-    throw new Error('Article needs JSON-LD with a publisher name.');
+    for (let i = 0; i < publisherNameElements.length; i++) {
```
mmmicedcoffee, line 605: could we use a for...of loop here? it would save one line (line 606).

is there a preference for for...of vs vanilla for loops in the swgjs codebase?

mmmicedcoffee (approved):

src/utils/gaa.js
ChrisAntaki: We've blocked `for...of` loops with a [custom ESLint rule](https://github.com/subscriptions-project/swg-js/blob/main/build-system/eslint-rules/no-for-of-statement.js) since I joined. There's a chance we can remove this rule though. I'll look into it, and send a follow-up PR if it's possible

PR #1859, opened 2021-08-05
Showcase: Log click events

ChrisAntaki:


src/api/subscriptions.js
```diff
@@ -359,6 +359,11 @@ export const ShowcaseEvent = {
   EVENT_SHOWCASE_INELIGIBLE_PAYWALL: 'EVENT_SHOWCASE_INELIGIBLE_PAYWALL', // When the user is not eligible for showcase entitlements
   EVENT_SHOWCASE_NO_ENTITLEMENTS_PAYWALL:
     'EVENT_SHOWCASE_NO_ENTITLEMENTS_PAYWALL', // When the user has no remaining showcase entitlements
+
+  // Events indicating the user interacted with the Showcase Regwall
```
ed-bird, line 363: Just checking if these need to be here?

All of the events related to the meter toast are just in the API messages: https://github.com/subscriptions-project/swg-js/blob/c10f2949cefdc476fef5faa49a348cb75475568a/src/proto/api_messages.js#L85-L89

src/runtime/event-type-mapping.js
```diff
@@ -89,6 +89,16 @@ const ShowcaseEvents = {
     // TODO(b/181690059): Create showcase ineligible AnalyticsEvent
     AnalyticsEvent.IMPRESSION_PAYWALL,
   ],
+
+  // User clicks "Sign in with Google" on Showcase Regwall.
```
ed-bird, line 93: As per previous comment, I don't think these need to be here either.

ed-bird (approved):
LGTM - just not sure about those ShowcaseEvents and if they need to be there or if those are primarily for the entitlement events.

src/api/subscriptions.js
ChrisAntaki: Good call! It's simpler now

PR #1862, opened 2021-08-09
IE11 Deprecation: Log a warning

ChrisAntaki:


jpettitt (approved):

jpettitt:
Approved but don't pull support from the build until we're communicated to publishers.

ChrisAntaki:
Sounds good !

PR #1863, opened 2021-08-09
E2E: Wait for local server to boot up

ChrisAntaki:
Fixes #1838


hartmannr76 (approved):
Nice! If this doesnt work, we could always test the endpoint once we know its started to see if it'll accept the traffic too, but this is easier to see if it'll work for now

ChrisAntaki:
That is a great idea, will definitely keep that in mind

PR #1867, opened 2021-08-13
SwG Buttons: Fix disabled state for elements other than buttons

ChrisAntaki:
Now that elements like `<span>`s can be SwG buttons, the HTML/CSS needs to change a bit


ChrisAntaki:
image

hartmannr76 (approved):

PR #1871, opened 2021-08-19
Set client timestamp on outgoing analytics events

dvirsky:
This PR adds a timestamp of the current client machine to every outgoing analytics request.

### DIFF
--- a/src/runtime/analytics-service.js
+++ b/src/runtime/analytics-service.js
@@ -132,10 +132,9 @@ export class AnalyticsService {
     /** @private {?number} */
     this.timeout_ = null;
 
-    // Time source for setting the client timestamp before sending requests.
-    // Injectable for tests.
+    // A callback for setting the client timestamp before sending requests.
     /** @private {function():!../proto/api_messages.Timestamp} */
-    this.timeSource_ = () => {
+    this.getTimestamp_ = () => {
       return toTimestamp(Date.now());
     };
   }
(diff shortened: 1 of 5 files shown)


dvirsky:
The checks are failing from something unrelated to my PR :/

src/runtime/analytics-service.js
```diff
@@ -130,6 +131,13 @@ export class AnalyticsService {
     // the user wait too long.
     /** @private {?number} */
     this.timeout_ = null;
+
+    // Time source for setting the client timestamp before sending requests.
+    // Injectable for tests.
+    /** @private {function():!../proto/api_messages.Timestamp} */
+    this.timeSource_ = () => {
```
ChrisAntaki, line 138: Could you call this `getTimestamp_` ? I think that's clearer than `timeSource_`

src/runtime/analytics-service.js
```diff
@@ -150,6 +158,14 @@ export class AnalyticsService {
     }
   }
 
+  /**
```
ChrisAntaki, line 161: It's simpler for tests to just override the `getTimestamp_` property

src/runtime/analytics-service.js
```diff
@@ -315,7 +331,7 @@ export class AnalyticsService {
     const meta = new AnalyticsEventMeta();
     meta.setEventOriginator(event.eventOriginator);
     meta.setIsFromUserAction(!!event.isFromUserAction);
-
+    this.updateContextTimestamp_();
```
ChrisAntaki, line 334: Since it's called just once, and it's a one liner, could you inline it for simplicity ?

dvirsky: Done.

src/runtime/analytics-service.js
dvirsky: Done.

src/runtime/analytics-service.js
dvirsky: Done. Old habbits from C++ :)

ChrisAntaki (approved):

PR #1894, opened 2021-09-08
Allow Dialog to be contained in a given element and iframe CSS class

li-cai:
- Add new method to open the Dialog in the specified container
- Add iframeCssClass to DialogConfig to specify the iframe CSS class to use instead of swg-dialog

### DIFF
--- a/src/components/dialog-test.js
+++ b/src/components/dialog-test.js
@@ -272,6 +272,21 @@ describes.realWin('Dialog', {}, (env) => {
       expect(container.nodeName).to.equal('SWG-CONTAINER');
     });
 
+    it('opens the iframe inside the container element', async () => {
+      const containerEl = doc.createElement('div');
+      doc.body.appendChild(containerEl);
+      const openedDialog = await dialog.openInContainer(containerEl);
+
+      // iframe element should be a child of the container element.
+      expect(containerEl.contains(openedDialog.getIframe().getElement())).to.be
+        .true;
+
+      // The swg-container should also be created.
+      const container = openedDialog.getContainer();
+      expect(container.nodeType).to.equal(1);
+      expect(container.nodeName).to.equal('SWG-CONTAINER');
+    });
+
     it('should throw if container is missing', async () => {
       expect(() => dialog.getContainer()).to.throw('not opened yet');
     });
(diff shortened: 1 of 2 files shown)


ChrisAntaki (approved):

src/components/dialog.js
```diff
@@ -73,10 +73,13 @@ const resetViewStyles = {
  * - desktopConfig: Options for dialogs on desktop screens.
  * - maxAllowedHeightRatio: The max allowed height of the view as a ratio of the
  *       viewport height.
+ * - iframeCssClass: The CSS class for the iFrame. Defaults to 'swg-dialog' if
```
mmmicedcoffee, line 76: hmm, I wonder if iframeCssClassOverride is a better name for this property, because it overrides both 'swg-dialog' and the 'swg-dialog swg-wide-dialog' combo. in the future, there may be other configuration options that add CSS classes, too, and setting this property will override all of them.

src/components/dialog-test.js
```diff
@@ -272,6 +272,21 @@ describes.realWin('Dialog', {}, (env) => {
       expect(container.nodeName).to.equal('SWG-CONTAINER');
     });
 
+    it('should open the iframe inside the container element', async () => {
```
mmmicedcoffee, line 275: nit: I prefer omitting "should" in test descriptions, since the "should" part is usually implied. 

this description would read "opens the iframe inside the container element" instead.

src/components/dialog-test.js
```diff
@@ -317,6 +332,12 @@ describes.realWin('Dialog', {}, (env) => {
       expect(() => dialog.open()).to.throw('already opened');
     });
 
+    it('should throw if iframe already connected when opening in a container', async () => {
```
mmmicedcoffee, line 335: nit: same here, this description would start with "throws if...." instead

mmmicedcoffee: also, does this test need to be async? I don't see any awaits in the test body.

src/components/dialog.js
ChrisAntaki: Good point, what if this param just added additional classes instead of overriding the default ones?

src/components/dialog-test.js
li-cai: True, removed async and updated description

src/components/dialog-test.js
li-cai: Makes sense! Done.

src/components/dialog.js
li-cai: Good point! I renamed to iframeCssClassOverride as for my use case, the styles provided by swg-dialog are not needed, so I wanted to be able to override with a different class.

mmmicedcoffee (approved):

PR #1897, opened 2021-09-09
Analytics: Avoid logging subscription unlock events for crawlers

ChrisAntaki:
Subscription unlock events from crawlers aren't worth logging, they just pollute logs

### DIFF
--- a/src/runtime/entitlements-manager.js
+++ b/src/runtime/entitlements-manager.js
@@ -28,6 +28,7 @@ import {
   Entitlement,
   Entitlements,
   GOOGLE_METERING_SOURCE,
+  PRIVILEGED_SOURCE,
 } from '../api/entitlements';
 import {
   GetEntitlementsParamsExternalDef,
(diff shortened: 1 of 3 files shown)


mmmicedcoffee (approved):

src/runtime/entitlements-manager.js
```diff
@@ -595,13 +596,18 @@ export class EntitlementsManager {
...
-    this.deps_
-      .eventManager()
-      .logSwgEvent(
-        AnalyticsEvent.EVENT_UNLOCKED_BY_SUBSCRIPTION,
-        false,
-        params
-      );
+
+    // Avoid logging events from crawlers.
+    if (entitlement.source !== PRIVILEGED_SOURCE) {
```
jpettitt, line 601: Not sure about this.  Knowing that we used privilege to unlock is useful to publishers who are trying to reconcile their page reads fro  other analytics with subscription unlocks.  Maybe better to filter these out in our logs rather than stop logging them entirely.

PR #1902, opened 2021-09-17
Include OpenDialogRequest in api_messages

li-cai:
Used to notify that a dialog should be opened for the specified URL path


ChrisAntaki (approved):

mmmicedcoffee (approved):
This change by itself looks good, but as discussed offline, let's document the change to the shared API, and rationale, in the original design doc.

li-cai:
Sounds good, will do!

PR #1923, opened 2021-10-05
Update pay-flow to pass swgUserToken and orderId as URL params

li-cai:
Also fix the argument order when constructing a SubscribeResponse in parseSubscriptionResponse where the order of swgUserToken and payRecurrence were swapped


li-cai:
Need to make some more changes. Closing for now until they are ready.

PR #1925, opened 2021-10-06
Pass additional URL params to the payconfirmiframe

li-cai:
- Pass swgUserToken and orderId
- Add requestMetadata to SubscribeResponse in order to pass canonicalUrl and isAnonymous
- Also fix argument order when constructing a SubscribeResponse in parseSubscriptionResponse where the swgUserToken and payRecurrence arguments were swapped

### DIFF
--- a/src/runtime/pay-flow.js
+++ b/src/runtime/pay-flow.js
@@ -356,6 +356,10 @@ export class PayCompleteFlow {
         offerId: this.sku_,
         origin: parseUrl(this.win_.location.href).origin,
       };
+      if (response.requestMetadata) {
+        urlParams.canonicalUrl = response.requestMetadata.contentId;
+        urlParams.isAnonymous = response.requestMetadata.anonymous;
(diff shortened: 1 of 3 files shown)


src/runtime/pay-flow.js
```diff
@@ -349,12 +350,24 @@ export class PayCompleteFlow {
...
+        urlParams.isAnonymous = response.requestMetadata.anonymous;
+      }
+      if (response.swgUserToken) {
+        urlParams.sut = response.swgUserToken;
+      }
+      const orderId = parseOrderIdFromPurchaseDataSafe(response.purchaseData);
+      if (orderId) {
+        urlParams.orderId = orderId;
+      }
+      confirmFeUrl = feUrl('/payconfirmiframe', '', urlParams);
```
alin04, line 370: Any risk of these url params being copied ? Should these values be passed in web activities?

PR #1928, opened 2021-10-06
Add support for prefixedHostPath option to handle pending server change

adolgachev:
Add prefixedHostPath to ClientConfig and update feUrl to insert prefix appropriately. This is necessary to handle a pending switch in the back-end which will reorder prefix and 'swg' base path.

- Shuffle arguments for feUrl, moving prefix to end, adding boolean for host path prefix, moving params to second as most used
- Update all usages of feUrl
- Update feUrl to check new param and swap order of prefix appropriately
- Add prefixedHostPath to ClientConfig and pull from json in ClientConfigManager
- Update linkAccountsFlow to check config and pass along prefixedHostPath in async call


adolgachev:
On hold for https://github.com/subscriptions-project/swg-js/pull/1929

ChrisAntaki (approved):

PR #1944, opened 2021-10-14
E2E: Test minified JS

ChrisAntaki:
This PR updates our E2E tests to check minified JS, since that's what runs in productions

### DIFF
--- a/test/e2e/tests/buyflow.js
+++ b/test/e2e/tests/buyflow.js
@@ -15,11 +15,13 @@
  */
 
 module.exports = {
-  '@tags': ['buyflow'],
-  'Show offers on web': function (browser) {
-    const setup = browser.page.setup();
-    setup.navigate().select('local_min');
+  beforeEach(browser) {
+    browser.useMinifiedJs();
+  },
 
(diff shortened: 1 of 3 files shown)


test/e2e/tests/buyflow.js
```diff
@@ -18,7 +18,7 @@ module.exports = {
   '@tags': ['buyflow'],
   'Show offers on web': function (browser) {
     const setup = browser.page.setup();
-    setup.navigate().select('local');
+    setup.navigate().select('local_min');
```
mmmicedcoffee, line 21: do you think it's be worth extracting these two lines into a common setUp() function that's shared between all the test cases in this file?

or since it's shared across all the e2e tests, maybe a testing utility that selects the environment (local vs local_min, etc)? I'm not sure if it's worth since it's only two lines, but perhaps we'll have more common test setup in the future.

ChrisAntaki: Good call! I added a `useMinifiedJs` command, and added it to a `beforeEach` method in each test suite

PR #1972, opened 2021-11-01
Use important rule for Showcase CSS

ChrisAntaki:
Fixes http://b/202482741

### DIFF
--- a/src/utils/gaa.js
+++ b/src/utils/gaa.js
@@ -477,6 +477,7 @@ export class GaaMeteringRegwall {
     iframeUrl = addQueryParam(iframeUrl, 'lang', languageCode);
 
     // Create and style container element.
+    // TODO: Consider using a FriendlyIframe here, to avoid CSS conflicts.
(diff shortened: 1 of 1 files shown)


mmmicedcoffee (commented):
I read the description of the bug, and I'm a little confused about the root cause because I thought iframes don't inherit the styles from their parent/host page. Why are we seeing CSS interference for the contents of our iframe?

ChrisAntaki:
Nice catch! I accidentally applied the `!important` rule to the iframe styles, instead of the Regwall styles. It should be fixed now

mmmicedcoffee (approved):
The fix makes more sense to me now!

What are your thoughts on rendering the regwall dialog in an iframe vs directly in swgjs like it is done now? The former pattern is more common with other SwG components, and it would avoid the styling issues that we're running into now.

ChrisAntaki:
Good idea, I'll make a bug for the FriendlyIframe upgrade

PR #1976, opened 2021-11-02
Swap from gulp-nightwatch to directly call nightwatch

adolgachev:
gulp-nightwatch is getting long in the tooth, and this allows us to upgrade our nightwatch to 1.7 from 1.3


ChrisAntaki (approved):

ChrisAntaki:
Nice!

PR #1982, opened 2021-11-02
Remove need for setup in tests

adolgachev:
There are some issues with e2e tests having to hit the setup page and change to local minified js target.

Better to just allow this to be configured at server startup, for speed and simplicity.


ChrisAntaki (approved):
This is the way

PR #1983, opened 2021-11-03
Implementation of GIS sign in button

xiexr151e:


examples/google-signin/sign-in-with-google-iframe.html
```diff
@@ -16,6 +13,7 @@
   
     <script>
       GaaSignInWithGoogleButton.show({
+        clientId: '520465458218-e9vp957krfk2r0i4ejeh6aklqm7c25p4.apps.googleusercontent.com',
```
jpettitt, line 16: Add a comment telling them they need to replace with their own client id.

src/utils/gaa.js
```diff
@@ -38,6 +38,8 @@ import '../model/doc';
 import {AnalyticsEvent, EventOriginator} from '../proto/api_messages';
 import {showcaseEventToAnalyticsEvents} from '../runtime/event-type-mapping';
 
+const jsonwebtoken = require('jsonwebtoken');
```
jpettitt, line 41: jsonwebtoken is a dev dependency - in general we don't include non-google code in the production binary (see package.json - the only external deps are google authrored).  Can we avoid using it here?

jpettitt (changes requested):
See comments re JWT

src/utils/gaa.js
jpettitt: Nevermind - jasonwebtoken is an approved lib - you do need to move it to `dependencies`

jpettitt: For future ref check g3/third_party to see if we already use a module and also use [jsgl](https://github.com/google/js-green-licenses) to confirm it's an ok license (we're good in this case)

jpettitt: Hold on approving I'm checking on exactly how we need to include this - see email you are about to get.

xiexr151e: @ChrisAntaki mentioned that there is actually already a decoding function in our code, so I think there is no need to use this library.

src/utils/gaa-test.js
```diff
@@ -849,21 +847,20 @@ describes.realWin('GaaSignInWithGoogleButton', {}, () => {
 
   describe('show', () => {
     it('renders Google Sign-In button', async () => {
```
ChrisAntaki, line 849:
```suggestion
    it('renders Sign in with Google button', async () => {
```

src/utils/gaa.js
```diff
@@ -330,6 +331,18 @@ const GOOGLE_SIGN_IN_IFRAME_STYLES = `
...
+ * Conformation of GaaUserDef. Contains a subset of information returned from JWT issued by GIS.
+ * @typedef {{
+ *   name: string,
+ *   givenName: string,
+ *   familyName: string,
+ *   imageUrl: string,
+ *   email: string,
+ * }} GisUserDef
+ */
+export let GisUserDef;
```
ChrisAntaki, line 344: Could we delete this typedef, and just use the `GoogleUserDef` typedef?

src/utils/gaa.js
```diff
@@ -376,6 +389,19 @@ export let GoogleIdentityV1;
...
+ * https://developers.google.com/identity/gsi/web/reference/js-reference#credential
+ * @typedef {{
+ *  name: string,
+ *  given_name: string,
+ *  family_name: string,
+ *  picture: string,
+ *  email: string,
+ * }} CredentialResponseDef
+ */
+export let CredentialResponseDef;
```
ChrisAntaki, line 403: Could we delete this typedef, and just use the `GoogleUserDef` typedef?

src/utils/gaa.js
```diff
@@ -953,63 +1014,26 @@ export class GaaSignInWithGoogleButton {
...
-        /** @type {!GaaUserDef} */
-        const gaaUser = {
-          idToken: authorizationData.id_token,
-          name: basicProfile.getName(),
-          givenName: basicProfile.getGivenName(),
-          familyName: basicProfile.getFamilyName(),
-          imageUrl: basicProfile.getImageUrl(),
-          email: basicProfile.getEmail(),
-          authorizationData,
+    initButtonFnPromise
```
ChrisAntaki, line 1017: Just curious, does this promise chain need to be split up?

src/utils/gaa.js
ChrisAntaki: Oops, I meant to suggest `GoogleIdentityV1`

src/utils/gaa.js
ChrisAntaki: Oops, I meant to suggest `GoogleIdentityV1`

ChrisAntaki (approved):

jpettitt (approved):

slmartinet:
Hi amazing team :)
Quick question please. Why did we need to import `gis/intermediatesupport` please?  https://screenshot.googleplex.com/3PwsWmDWbzNAYiT

PR #1993, opened 2021-11-10
Focus the dialog contents after its opened

mmmicedcoffee:
WAI-ARIA best practices recommend that dialogs should be focused after they are opened, for convenience to users who navigate via keyboard controls only.


li-cai (approved):

mmmicedcoffee:
Hmm, I'm looking into why the test is hanging.

mmmicedcoffee:
Fixed the hanging test. It should've been a test failure since I was comparing the wrong two elements, but somehow the failure manifested as a stalled test rather than a failure.

PR #1997, opened 2021-11-12
Update e2e tests to work with Firefox.

michaelwilcox:
Update the web element passed to `frame()` to not use legacy JsonWire `ELEMENT`. `geckodriver` dropped support for legacy web element references in [v0.25.0](https://github.com/mozilla/geckodriver/releases/tag/v0.25.0#removed). This still works with `chromedriver`.

In the `Show offers after clicking smart button` test, the `click()` callback was returning `{ value: null }` for Firefox without a `status` field. The assertion on the callback to `click()` does not test that the offers dialog was opened. I updated the test to change the assertion to test that the SwG dialog is opened after click of the smart button.


test/e2e/tests/smartButton.js
```diff
@@ -41,10 +41,15 @@ module.exports = {
     const smartButton = browser.page.smartButton();
     smartButton
       .navigate()
+      .pause(2500)
```
adolgachev, line 44: Is this just for FF? Can we do only with FF or at least add a comment to note why?

adolgachev (changes requested):

test/e2e/tests/smartButton.js
michaelwilcox: It's for both Chrome and Firefox. It's to wait for loading external content of the Smart Box iFrame, which I've noticed can be slightly delayed. The test code traverses into that iFrame and clicks smart button to load the Offers frame. Thinking more if there was an event that it could listen for to know the Smart Box iFrame has been loaded that would be best?

michaelwilcox: Here is a recording to show the delay on load of iFrame

https://user-images.githubusercontent.com/161528/141587180-068d65e8-011b-4c80-a472-61395c23e9c1.mp4

michaelwilcox: I replaced `pause` with `waitForElementPresent` and `waitForElementVisible` in https://github.com/subscriptions-project/swg-js/pull/1997/commits/826278706de1962f846221472380af4c8f6c983e which should be more explicit what it is waiting for and should achieve same result. Similar to how we are doing in [contribution](https://github.com/subscriptions-project/swg-js/blob/main/test/e2e/tests/contribution.js#L24).

michaelwilcox:
I replaced `pause` with `waitForElementPresent` and `waitForElementVisible` in https://github.com/subscriptions-project/swg-js/pull/1997/commits/826278706de1962f846221472380af4c8f6c983e which should be more explicit what it is waiting for and should achieve same result. Similar to how we are doing in [contribution](https://github.com/subscriptions-project/swg-js/blob/main/test/e2e/tests/contribution.js#L24).

test/e2e/tests/smartButton.js
michaelwilcox: @adolgachev 

In 9cc3122, changed the click callback to handle different return values for chrome/FF.
This was the most simple fix, and with these changes tests should now be passing in chrome and firefox.
Reverted and shelved previous code that re-wrote test.

adolgachev (approved):

PR #2013, opened 2021-11-22
UI specifications and iframe documentation

xiexr151e:
This PR deals with 2 things:
1. Milind's UI specifications on the GIS button
2. Changing the documentation on the iframe to point to GIS website


src/utils/gaa.js
```diff
@@ -157,7 +157,7 @@ const REGWALL_HTML = `
   .gaa-metering-regwall--iframe {
     border: none !important;
     display: block !important;
-    height: 36px !important;
+    height: 44px !important;
```
slmartinet, line 160: Out of curiosity, why this change in height please?

slmartinet:
Looks good to me. Thank you

src/utils/gaa.js
xiexr151e: This is because the default height for the button is 44px.
I was previously having an issue in which the bottom edge of the button was being cut off. Changing this height resolved the issue.

adolgachev (approved):

PR #2020, opened 2021-11-27
Make skipAccountCreationScreen a config option

jpettitt:
Makes `skipAccountCreationScreen` an config option to avoid a race condition when AMP tries to suppress it.

If AMP tries to suppress the screen by immediately acknowledging it the message is getting lost because the iframe hasn't yet rendered.  It's not 100% clear why this started happening, best guess is that it is due to a couple of extra promise() resolutions in the codepath path from adding remote config that cause the AMP thread to schedule before the SWG one has built the iframe (in other words a latent bug)

Since we now have a flag to suppress the iframe this PR enabled it to be set by publisher code (in this case AMP)


jpettitt:
Not current test error will go away when swg.js accepts the new param.

ChrisAntaki:
Nice find!

ChrisAntaki (approved):

PR #2028, opened 2021-12-02
Correcting layout shift issue in GIS button

xiexr151e:
Layout shift issue was occurring because a height-width specification was not set.


ChrisAntaki:
Looks like some checks are failing 🤔

src/utils/gaa.js
```diff
@@ -970,6 +970,10 @@ export class GaaSignInWithGoogleButton {
           'theme': 'outline',
           'text': 'continue_with',
           'logo_alignment': 'center',
+          'width': self.document.getElementById(SIGN_IN_WITH_GOOGLE_BUTTON_ID)
```
ChrisAntaki, line 973: Could you create a local variable for this button element? (`self.document.getElementById(SIGN_IN_WITH_GOOGLE_BUTTON_ID)`). That would allow the script to just query for it once, and could improve readability too

ChrisAntaki: Oh you could actually just reference `buttonEl` directly

xiexr151e: Done.

src/utils/gaa.js
```diff
@@ -970,10 +970,8 @@ export class GaaSignInWithGoogleButton {
           'theme': 'outline',
           'text': 'continue_with',
           'logo_alignment': 'center',
-          'width': self.document.getElementById(SIGN_IN_WITH_GOOGLE_BUTTON_ID)
-            .offsetWidth,
-          'height': self.document.getElementById(SIGN_IN_WITH_GOOGLE_BUTTON_ID)
-            .offsetHeight,
+          'width': buttonEl.offsetWidth,
```
ChrisAntaki, line 973: Thanks! While you're at it, could you also do the same for line 967?

ChrisAntaki (approved):

PR #2035, opened 2021-12-09
Use article endpoint for swg-basic runtime

hartmannr76:
Adds functionality which allows us to migrate from using both the `/entitlements` and `/clientconfiguration` endpoints into the single `/article` endpoint for swg-basic runtime, which includes the combined `clientConfig` and `entitlements` responses under a single payload so we don't have to do multiple round-trips.

All of this is placed behind a flag (`use-article-endpoint`) so existing logic should stay the same for the time being


src/runtime/client-config-manager.js
```diff
@@ -30,9 +30,10 @@ export class ClientConfigManager {
   /**
    * @param {string} publicationId
    * @param {!./fetcher.Fetcher} fetcher
+   * @param {!./deps.DepsDef} deps
    * @param {!../api/basic-subscriptions.ClientOptions=} clientOptions
    */
-  constructor(publicationId, fetcher, clientOptions) {
+  constructor(publicationId, fetcher, deps, clientOptions) {
```
adolgachev, line 36: Can you please move deps to the first argument? We're not consistent, but I'd like to make the managers put deps in first in all cases. Since it's required, you'll need to update all the usages anyways.

adolgachev (changes requested):
Still looking over everything, but wanted to get this change request out first.

src/runtime/client-config-manager.js
hartmannr76: Makes sense. Done

jpettitt (approved):

adolgachev (approved):

PR #2040, opened 2021-12-13
Tempswgjs

ksebina:
Draft pull request


src/runtime/audience-activity-listener-test.js
```diff
@@ -0,0 +1,94 @@
...
+    let eventManagerCallback;
+    let audienceActivityEventListener;
+    let eventsLoggedToService;
+  
+    const productId = 'pub1:label1';
+  
+    beforeEach(() => {
+        setExperimentsStringForTesting('');
+        eventsLoggedToService = [];
+    sandbox
```
oyj9109, line 41: It doesn't need to be stubbed. 
(env.win.document.referrer is not referenced in the listener.)

src/runtime/audience-activity-listener-test.js
```diff
@@ -0,0 +1,94 @@
...
+      .callsFake((unusedUrl, message) => {
+        eventsLoggedToService.push(message);
+      });
+    sandbox
+      .stub(ClientEventManager.prototype, 'registerEventListener')
+      .callsFake((callback) => (eventManagerCallback = callback));
+    pageConfig = new PageConfig(productId);
+    runtime = new ConfiguredRuntime(env.win, pageConfig);
+    audienceActivityEventListener = new AudienceActivityEventListener(runtime, runtime.fetcher_);
+    });
```
oyj9109, line 56: Please run linter.

src/runtime/audience-activity-listener-test.js
```diff
@@ -0,0 +1,94 @@
...
+          // This triggers an event.
+          eventManagerCallback({
+            eventType: AnalyticsEvent.IMPRESSION_PAYWALL,
+            eventOriginator: EventOriginator.UNKNOWN_CLIENT,
+            isFromUserAction: null,
+            additionalParameters: null,
+          });
+  
+          // These wait for the listener to be ready to send data.
+          expect(audienceActivityEventListener.lastAction_).to.not.be.null;
```
oyj9109, line 70: The listener doesn't register anything on lastAction_, so it should be null. 
Also, it directly calls sendBeacon without having to wait for an async call, so await is not necessary.

This part should fail. Please make sure to run the tests and get them to pass.

src/runtime/audience-activity-listener.js
```diff
@@ -0,0 +1,82 @@
+
```
oyj9109, line 1: Please remove redundant white space at the top.

src/runtime/audience-activity-listener.js
```diff
@@ -0,0 +1,82 @@
...
+    
+    /**
+     *  Listens for new audience activity events from the events manager and sends them to the SwG Client Server.
+     * @param {!../api/client-event-manager-api.ClientEvent} event
+     */
+    handleClientEvent_(event) {
+      if (audienceActivityLoggingEvents.has(event.eventType)) {
+        const pubId = encodeURIComponent(this.deps_.pageConfig().getPublicationId());
+        const audienceActivityClientLogsRequest = this.createLogRequest(event);
+        const url = serviceUrl('/publication/' + pubId + '/audienceactivitylogs');
```
oyj9109, line 67: (Optional) Should we add SwgUserToken? You can add it in a follow-up change if you think it makes sense to separate it.

src/runtime/audience-activity-listener-test.js
```diff
@@ -0,0 +1,94 @@
...
+            eventOriginator: EventOriginator.UNKNOWN_CLIENT,
+            isFromUserAction: null,
+            additionalParameters: null,
+          });
+
+          // These wait for the listener to be ready to send data.
+          expect(audienceActivityEventListener.lastAction_).to.not.be.null;
+          await audienceActivityEventListener.lastAction_;
+          
+          expect(eventsLoggedToService.length).to.equal(1);
```
oyj9109, line 91: Can we check the event type to be logged here is equal to IMPRESSION_PAYWALL?
Also, can we add a check for path?

Refer to https://github.com/subscriptions-project/swg-js/blob/main/src/runtime/propensity-server-test.js#L106-L107

src/runtime/audience-activity-listener-test.js
```diff
@@ -0,0 +1,94 @@
...
+      });
+    sandbox
+      .stub(ClientEventManager.prototype, 'registerEventListener')
+      .callsFake((callback) => (eventManagerCallback = callback));
+    pageConfig = new PageConfig(productId);
+    runtime = new ConfiguredRuntime(env.win, pageConfig);
+    audienceActivityEventListener = new AudienceActivityEventListener(runtime, runtime.fetcher_);
+    });
+
+    describe('Audience Activity Events Experiment', () => {
```
oyj9109, line 58: What about AudienceActivityEventListener?
It will be appended by what's in "it". E.g., "AudienceActivityEventListener should not log audience ..."

src/runtime/audience-activity-listener.js
ksebina: Will do in a follow up cl

oyj9109 (commented):
Can you remove the yarn.lock file? I believe that's not a necessary update here.

PR #2044, opened 2021-12-15
Add an event listener to log audience activity

ksebina:


src/runtime/audience-activity-listener.js
```diff
@@ -0,0 +1,95 @@
...
+    );
+  }
+
+  /**
+   *  Listens for new audience activity events from the events manager and sends them to the SwG Client Server.
+   * @param {!../api/client-event-manager-api.ClientEvent} event
+   */
+  handleClientEvent_(event) {
+    if (event.eventType && audienceActivityLoggingEvents.has(event.eventType)) {
+
```
oyj9109, line 77: Line break seems redundant here. Would you remove it?

src/runtime/audience-activity-listener.js
```diff
@@ -0,0 +1,95 @@
+/**
+ * Copyright 2018 The Subscribe with Google Authors. All Rights Reserved.
```
oyj9109, line 2: Please update the copyright year.

oyj9109 (approved):
Awesome work! Left a few minor comments, but overall LGTM.

src/runtime/basic-runtime.js
```diff
@@ -313,6 +314,21 @@ export class ConfiguredBasicRuntime {
...
 
+    // Start listening to Audience Activity events.
+    if (
+      isExperimentOn(
+        this.doc_.getWin(),
+        ExperimentFlags.LOGGING_AUDIENCE_ACTIVITY
+      )
+    ) {
+      /** @private @const {!AudienceActivityEventListener} */
+      this.audienceActivityEventListener_ = new AudienceActivityEventListener(
```
ChrisAntaki, line 325: Could you test this new if statement?
https://app.codecov.io/gh/subscriptions-project/swg-js/compare/2044/diff

ksebina:
@hartmannr76 can you add any additional feedback if you have any? Thank you

src/runtime/audience-activity-listener.js
```diff
@@ -0,0 +1,95 @@
...
+      this.handleClientEvent_.bind(this)
+    );
+  }
+
+  /**
+   *  Listens for new audience activity events from the events manager and sends them to the SwG Client Server.
+   * @param {!../api/client-event-manager-api.ClientEvent} event
+   */
+  handleClientEvent_(event) {
+    if (event.eventType && audienceActivityLoggingEvents.has(event.eventType)) {
```
hartmannr76, line 76: Maybe this is just a general question - We should only want to log events if a user has a `swgUserToken`, right? Otherwise, I don't know how much value we get from unidentifiable messages. It looks like we have some [code-paths around that will save the token if we have it](https://github.com/subscriptions-project/swg-js/blob/2e15a1f5910b48d9e159e9fda1d0e1d2b849f472/src/runtime/basic-runtime.js#L468), but should we guard these requests behind a check to see if there is a token available? Also, this would mean we want that token as a query parameter, right?

src/runtime/audience-activity-listener.js
```diff
@@ -0,0 +1,95 @@
...
+    this.fetcher_ = fetcher;
+
+    /** @private @const {!Storage} */
+    this.storage_ = new Storage(this.win_);
+  }
+
+  /**
+   * Start listening to client events.
+   */
+  start() {
```
hartmannr76, line 65: Should these functions be annotated? `start` with `@public`, `handleClientEvent_` and `createLogRequest` with `@private`? I think that also means `createLogRequest` should be `createLogRequest_` to follow convention if it is meant to be private

src/runtime/audience-activity-listener.js
```diff
@@ -0,0 +1,95 @@
...
+    }
+  }
+
+  /**
+   * @param {!../api/client-event-manager-api.ClientEvent} event
+   * @return {!AudienceActivityClientLogsRequest}
+   */
+  createLogRequest(event) {
+    const request = new AudienceActivityClientLogsRequest();
+    request.setEvent(/** @type {!AnalyticsEvent} */ (event.eventType));
```
hartmannr76, line 92: Are the extra parens necessary?
```suggestion
    request.setEvent(/** @type {!AnalyticsEvent} */ event.eventType);
```

src/runtime/audience-activity-listener.js
ksebina: fixed

src/runtime/audience-activity-listener.js
ksebina: I was planning on adding SwgUserToken in a follow-up change however to your point I think it makes sense to add it to this change. it terms of the guarding the change behind the check, it looks like swg_user_token has a default value of USER_TOKEN within Constants, so just checking that local storage sets it?

jpettitt (approved):

PR #2047, opened 2021-12-16
Add callback for PayConfirmOpened to pass the ActivityIframeView

li-cai:
This allows attaching host request/response listeners to the passed pay confirm ActivityIframeView, and resizing this view if necessary.

### DIFF
--- a/src/runtime/pay-flow-test.js
+++ b/src/runtime/pay-flow-test.js
@@ -20,6 +20,7 @@ import {
   EntitlementsResponse,
   EventParams,
 } from '../proto/api_messages';
+import {ActivityIframeView} from '../ui/activity-iframe-view';
 import {ActivityPort} from '../components/activities';
 import {ClientConfig} from '../model/client-config';
 import {ConfiguredRuntime} from './runtime';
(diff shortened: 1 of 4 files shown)


src/runtime/pay-flow-test.js
```diff
@@ -649,6 +650,35 @@ describes.realWin('PayCompleteFlow', {}, (env) => {
     expect(PayCompleteFlow.waitingForPayClient_).to.be.true;
   });
 
+  it('should have valid flow constructed that triggers the payConfirmOpened callback', async () => {
```
mmmicedcoffee, line 653: Prefer test descriptions that directly state the expected outcome, rather than saying "should...". Example:
'constructs valid flow that triggers the payConfirmedOpened callback'

Or even:
'triggers the payConfirmedOpened callback when dialog opens after flow start'

Also, minor nit: wrap the test description if it exceeds the 80-char line limit.

mmmicedcoffee:
I'm not sure I fully understand the use case for this new callback. Do you need to resize the iframe immediately after it opens?

src/runtime/pay-flow-test.js
li-cai: Makes sense! Updated and wrapped description.

> I'm not sure I fully understand the use case for this new callback. Do you need to resize the iframe immediately after it opens?

In Twg code, I wanted to register a handler for this callback:

this.callbacks().setOnPayConfirmOpened((activityIframeView) => {
    activityIframeView.on(SkuSelectedResponse, this.startPayFlow_.bind(this));
    activityIframeView.on(EntitlementsResponse, ...);
    activityIframeView.on(ActionRequest, ...);
    ...
});

For the Twg pay confirm view, we want to attach response listeners from the host because the buy flow can be started again from the pay confirm view.

mmmicedcoffee (approved):

PR #2050, opened 2021-12-20
Reuse MeterToastApi for displaying the meteriframe

longmelon:
Create a new autoprompt type and a new flow to display the meteriframe.


src/api/basic-subscriptions.js
```diff
@@ -99,6 +99,7 @@ export const AutoPromptType = {
   NONE: 'none',
   CONTRIBUTION: 'contribution',
   CONTRIBUTION_LARGE: 'contribution_large',
+  METER_LARGE: 'meter_large',
```
hartmannr76, line 102: These are params that are provided on the page to indicate the flow (contribution vs subscription). I don't think this belongs as a "type" here since it is actually a modal that only comes up under a subscription flow

src/runtime/basic-runtime.js
```diff
@@ -502,6 +502,10 @@ export class ConfiguredBasicRuntime {
           isClosable: !this.pageConfig().isLocked(),
         });
       };
+    } else if (options.autoPromptType == AutoPromptType.METER_LARGE) {
+      options.displayLargePromptFn = () => {
```
hartmannr76, line 506: So this would mean that we can either only show a subscription or meter prompt on a page, which may not be correct. The prompt itself depends on the result of the entitlements response so I feel like we should determine whether or not to prompt this [based on entitlements results ](https://github.com/subscriptions-project/swg-js/blob/98d20d7dc1d31762b6f7b8bc986d9d1b5b09ee6a/src/runtime/auto-prompt-manager.js#L124-L144)

src/runtime/meter-flow.js
```diff
@@ -0,0 +1,79 @@
...
+    /** @private @const {!../components/dialog-manager.DialogManager} */
+    this.dialogManager_ = deps.dialogManager();
+
+    /** @private @const {!ActivityIframeView} */
+    this.activityIframeView_ = new ActivityIframeView(
+      this.win_,
+      this.activityPorts_,
+      feUrl('/meteriframe', {
+        publicationId: deps.pageConfig().getPublicationId(),
+        origin: location.origin,
```
hartmannr76, line 45: I think most flows are grabbing the origin through something like [this](https://github.com/subscriptions-project/swg-js/blob/5f9e29c11faf130772c4e08c7dd8f1e778c00611/src/runtime/pay-flow.js#L357)

src/runtime/meter-flow.js
```diff
@@ -0,0 +1,79 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {ActivityIframeView} from '../ui/activity-iframe-view';
+import {SubscriptionFlows} from '../api/subscriptions';
+import {ViewSubscriptionsResponse} from '../proto/api_messages';
+import {feUrl} from './services';
+
+export class MeterFlow {
```
hartmannr76, line 22: Is this not meant to share similar functionality with `MeterToastApi`? Is it possible to re-use that and just pass some different arguments (i.e. primarily the url)?

src/api/subscriptions.js
```diff
@@ -337,6 +337,12 @@ export class Subscriptions {
    * @return {?}
    */
   showBestAudienceAction() {}
+
+  /**
+   * Show the meter prompt if the user has free reads.
+   * @return {?}
+   */
+  showMeter() {}
```
hartmannr76, line 345: Should this be exposed on the runtime? My understanding of this surface area is that it is mostly meant for being able to be invoked by publisher code, which most of the swg-basic runtime is pretty fixed to our flows

src/runtime/runtime.js
```diff
@@ -1160,6 +1168,17 @@ export class ConfiguredRuntime {
...
   }
+
+  /** @override */
+  showMeter() {
+    return this.documentParsed_.then(() => {
+      this.setOnNativeSubscribeRequest(() => {
+        this.showOffers({isClosable: true});
+      });
+      const flow = new MeterFlow(this);
+      return flow.start();
```
ChrisAntaki, line 1179: Could you add a test for this?
https://app.codecov.io/gh/subscriptions-project/swg-js/compare/2050/diff

src/runtime/meter-flow.js
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  /**
+   * Starts the meter flow.
+   * @return {!Promise}
+   */
+  start() {
+    this.deps_.callbacks().triggerFlowStarted(SubscriptionFlows.SHOW_METER);
+    this.activityIframeView_.onCancel(() => {
+      this.deps_.callbacks().triggerFlowCanceled(SubscriptionFlows.SHOW_METER);
```
ChrisAntaki, line 59: Could you add a test for this?
https://app.codecov.io/gh/subscriptions-project/swg-js/compare/2050/diff

src/api/basic-subscriptions.js
longmelon: Synced offline and we believe the best approach is to reuse MeterToastApi for displaying the new meteriframe. For the entry point that decides which prompt to show, Ricky is working on a PR that the AutoPromptManager will be responsible for that.

src/runtime/basic-runtime.js
longmelon: Reverted.

src/runtime/meter-flow.js
longmelon: Reverted.

src/runtime/meter-flow.js
longmelon: MeterToastApi is reused.

src/api/subscriptions.js
longmelon: Reverted.

src/runtime/runtime.js
longmelon: Reverted.

src/runtime/meter-flow.js
longmelon: Reverted.

src/runtime/meter-toast-api.js
```diff
@@ -33,8 +33,10 @@ export const MINIMIZED_IFRAME_SIZE = '420px';
 export class MeterToastApi {
   /**
    * @param {!./deps.DepsDef} deps
+   * @param {string=} iframeUrl Relative URL of the iframe, e.g. "/meteriframe".
+   * @param {Object<string, string>=} iframeParams List of extra params appended to the URL.
    */
-  constructor(deps) {
+  constructor(deps, iframeUrl = '/metertoastiframe', iframeParams = {}) {
```
hartmannr76, line 39: Hmmm can we maybe make an interface for these additional params to fit into? I never loved the idea of adding more than a few params to functions when deconstructing is so easy in JS. @ChrisAntaki any thoughts?

longmelon: Good idea. Done.

src/runtime/meter-toast-api.js
```diff
@@ -29,12 +29,26 @@ import {warn} from '../utils/log';
...
+ */
+export let MeterToastApiParams;
 
 export class MeterToastApi {
   /**
    * @param {!./deps.DepsDef} deps
+   * @param {?MeterToastApiParams=} params
    */
-  constructor(deps) {
+  constructor(deps, params = undefined) {
```
hartmannr76, line 51: You can destructure function parameters with a default ([reference](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#setting_a_function_parameters_default_value))
```suggestion
  constructor(deps, { iframeUrl = DEFAULT_IFRAME_URL, iframeUrlParams = {} } = {}) {
```

longmelon: Thanks. The reason I didn't do this was because I wasn't sure how the jsdoc shoulds be. Please review if it is correct.

src/runtime/meter-toast-api.js
```diff
@@ -37,18 +37,23 @@ export const DEFAULT_IFRAME_URL = '/metertoastiframe';
...
  * }}
  */
 export let MeterToastApiParams;
 
 export class MeterToastApi {
   /**
    * @param {!./deps.DepsDef} deps
-   * @param {?MeterToastApiParams=} params
+   * @param {!MeterToastApiParams=} params
+   * @param {string} params.iframeUrl
```
hartmannr76, line 50: Hmmmm it seems pretty weird that you have to specify these as params here even though they are described in the typedef. Does it still work if you remove these? I think you're also saying the params are optional but not nullable?
It seems like you should just be able to have

```
@param {MeterToastApiParams=} params
```

or you could also just inline the type since I don't think it needs to be referenced

```
/**
   * @param {{
       *   iframeUrl: (string),
       *   iframeUrlParams: (Object<string, string>),
   }=} params
*/
constructor(...)
```

longmelon: > Hmmmm it seems pretty weird that you have to specify these as params here even though they are described in the typedef.

Yeah, I was following the guideline [here](https://jsdoc.app/tags-param.html#parameters-with-properties) since the [doc](https://github.com/google/closure-compiler/wiki/Types-in-the-Closure-Type-System) for Closure type system doesn't seem to mention how to document a destructuring param. But when I switched to simply use `@param {!MeterToastApiParams=} params` the type check doesn't complain so I think that's okay.

hartmannr76 (approved):
LGTM

jpettitt (approved):

PR #2052, opened 2021-12-21
Using for-of instead of forEach in src/ directory

ChrisAntaki:
This PR updates src/ code to use for-each loops (easier to control) instead of the forEach method

This PR also allows [spread syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax) and [destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment), since IE11 is no longer a concern

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -38,11 +38,9 @@
   "rules": {
     "swg-internal/dict-string-keys": 2,
     "swg-internal/enforce-private-props": 2,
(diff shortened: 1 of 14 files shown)


jpettitt (approved):

mmmicedcoffee:
Nice!

PR #2053, opened 2021-12-22
Showcase: Use async/await

ChrisAntaki:
This PR updates the Showcase code to use async/await instead of chaining promises


adolgachev (approved):

mmmicedcoffee:
It appears async/await isn't supported in amphtml yet (https://github.com/ampproject/amphtml/issues/29216), so this PR is causing problems when I try to push the latest snapshot of our code into AMP.

I will create a PR to revert this change.

PR #2055, opened 2021-12-23
Upgrade ESLint and its Prettier plugin

ChrisAntaki:
This PR upgrades ESLint and its Prettier plugin. This PR also creates a local, updated fork of the `gulp-eslint` NPM module


mmmicedcoffee:
What's the rationale behind having a local copy of `gulp-eslint`? Also, if we need to update it in the future, would we do it manually? Should we record the version # that we're forking somewhere with the plugin code?

ChrisAntaki:
The gulp-eslint module's development seems stalled. Their last update was actually 3.5 years ago. So they're still on ESLint 6, instead of 8. If ESLint 9 makes breaking API changes, some manual updates might be necessary

third_party/gulp-eslint/README.md
```diff
@@ -0,0 +1,5 @@
+This fork modernizes [gulp-eslint](https://github.com/adametry/gulp-eslint) to use the latest ESLint version.
```
ChrisAntaki, line 1: @mmmicedcoffee Hopefully this README helps explain the situation more

mmmicedcoffee: Thanks!

mmmicedcoffee (approved):

PR #2064, opened 2022-01-06
Handling actions in autoprompt

hartmannr76:
Sets up the auto-prompt-manager to recognize provided audience actions, allowing a prompt for a user to complete a flow for additional entitlements

### DIFF
--- a/src/runtime/auto-prompt-manager.js
+++ b/src/runtime/auto-prompt-manager.js
@@ -15,6 +15,7 @@
  */
 
 import {AnalyticsEvent} from '../proto/api_messages';
+import {AudienceActionFlow} from './audience-action-flow';
 import {AutoPromptType} from '../api/basic-subscriptions';
 import {MiniPromptApi} from './mini-prompt-api';
 import {assert} from '../utils/log';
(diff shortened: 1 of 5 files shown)


src/runtime/auto-prompt-manager.js
```diff
@@ -114,25 +133,32 @@ export class AutoPromptManager {
...
    * @return {!Promise}
    */
-  showAutoPrompt_(clientConfig, entitlements, params) {
+  showAutoPrompt_(clientConfig, entitlements, article, params) {
     return this.shouldShowAutoPrompt_(
       clientConfig,
       entitlements,
       params.autoPromptType
     ).then((shouldShowAutoPrompt) => {
+      const promptFn =
```
jpettitt, line 150: Nit - put this block inside the `if (!shouldShowAutoPrompt)` since it's only used if `shouldShowAutoPrompt` is false.

jpettitt (approved):

jpettitt:
Need to fix tests

src/runtime/auto-prompt-manager.js
hartmannr76: Updated to also passing this along to the miniprompt below

jpettitt (approved):

PR #2076, opened 2022-01-11
Adds parameter to possibly return raw JWT signature in GaaMeteringRegWall.show()

xiexr151e:


slmartinet:
I think there is a misunderstanding here. The idea is not to 'return JWT signature in addition to payload'. 

The first option I described was to add the entire header. The second option, as described by Ed in [this comment](https://buganizer.corp.google.com/issues/200458496#comment20), is to give the option of having the _entire_ raw JWT. In our call, I suggested going forward with this second option.

As mentioned in the call we had, we suggest moving forward with this second option.

xiexr151e:
> I think there is a misunderstanding here. The idea is not to 'return JWT signature in addition to payload'.
> 
> The first option I described was to add the entire header. The second option, as described by Ed in [this comment](https://buganizer.corp.google.com/issues/200458496#comment20), is to give the option of having the _entire_ raw JWT.
> 
> In our call, I suggested going forward with this second option.

Just so I'm understanding this correctly, by raw we mean the undecoded JWT response?

slmartinet:
> > I think there is a misunderstanding here. The idea is not to 'return JWT signature in addition to payload'.
> > The first option I described was to add the entire header. The second option, as described by Ed in [this comment](https://buganizer.corp.google.com/issues/200458496#comment20), is to give the option of having the _entire_ raw JWT.
> > In our call, I suggested going forward with this second option.
> 
> Just so I'm understanding this correctly, by raw we mean the undecoded JWT response?

Yes

xiexr151e:
Got it. I will update accordingly.

xiexr151e:
@slmartinet I've updated the code to return the undecoded JWT, if needed.

slmartinet:
Thanks. Where is the value of `rawJwt` set by the programmer?

xiexr151e:
> Thanks. Where is the value of `rawJwt` set by the programmer?

Right now, nothing sets `rawJwt`; this is something that I would like to discuss further.

I'm thinking of setting it by using query param, but I'm not sure if there's a better approach to this.

slmartinet (commented):
These last 2 commits LGTM, who could give the final green light?

xiexr151e:
> These last 2 commits LGTM, who could give the final green light?

I really want @ChrisAntaki to have a look over this before I submit it. I've also deliberately not written tests for it yet.

src/utils/gaa.js
```diff
@@ -984,16 +984,19 @@ export class GaaSignInWithGoogleButton {
...
+          );
+        }
 
         // Send GAA user to parent frame.
         sendMessageToParentFnPromise.then((sendMessageToParent) => {
           sendMessageToParent({
             stamp: POST_MESSAGE_STAMP,
             command: POST_MESSAGE_COMMAND_USER,
-            jwtPayload,
+            returnedJwt,
```
jpettitt, line 999: since `jwtPayload` is in deployed code we should continue to return it and mark it deprecated in a comment so we don't break anything.

ChrisAntaki:
Looking good so far!

src/utils/gaa.js
xiexr151e: I've updated the code to note that `jwtPayload` is being deprecated.

ChrisAntaki:
Looking good, I just want to verify one test locally

ChrisAntaki (approved):

jpettitt (approved):

PR #2077, opened 2022-01-12
Suppport 'Already registered? sign in' CTA in regwall / newsletter

longmelon:
This works the same as the "Already a subscriber" CTA in subscription / contribution which opens a window for sign-in and then checks entitlements.


src/runtime/basic-runtime.js
```diff
@@ -507,6 +507,13 @@ export class ConfiguredBasicRuntime {
           lastContributionsFlow.showNoEntitlementFoundToast();
           return;
         }
+
+        const lastAudienceActionFlow =
+          this.autoPromptManager_.getLastAudienceActionFlow();
+        if (lastAudienceActionFlow) {
+          lastAudienceActionFlow.showNoEntitlementFoundToast();
```
hartmannr76, line 514: It feels strange to push this into all iframes. There isn't a singular one we could re-use to do that?

longmelon: Since subscription / contribution / regwall and newsletter iframes share the same feature, I think it's reasonable to add this piece here.

hartmannr76: Sounds good

src/runtime/audience-action-flow.js
```diff
@@ -92,6 +109,11 @@ export class AudienceActionFlow {
       this.handleEntitlementsResponse_.bind(this)
```
hartmannr76, line 109: Should we update this to listen on `CompleteAudienceActionResponse` now that that is being passed back?

src/runtime/audience-action-flow.js
```diff
@@ -138,4 +160,23 @@ export class AudienceActionFlow {
...
+  handleLinkRequest_(response) {
+    if (response.getSubscriberOrMember()) {
+      this.deps_.callbacks().triggerLoginRequest({linkRequested: false});
+    }
+  }
+
+  /**
+   * Shows the toast of 'no entitlement found' on activity iFrame view.
+   */
+  showNoEntitlementFoundToast() {
```
hartmannr76, line 177: `@public`

src/runtime/audience-action-flow.js
longmelon: I have a follow-up PR for this :)

src/runtime/audience-action-flow.js
longmelon: Done.

jpettitt (approved):

PR #2090, opened 2022-01-15
feat: always adding entitlement params to the call

hartmannr76:
This change makes sure that the encodedParams on an entitlements call always have some metering data, necessary if the state of the reader is stored on the backend.

This also uses the `MeterClientTypes.METERED_BY_GOOGLE` as the default and if GAA parameters are provided, will also include `MeterClientTypes.LICENSED_BY_GOOGLE` to the clientTypes in the request.


davwood-alpha:
FYI, see also https://github.com/subscriptions-project/swg-js/pull/2091

adolgachev (approved):

jpettitt (approved):

PR #2091, opened 2022-01-15
EntitlementsManager and MeterToastApi enhanced to support METERED_BY_GOOGLE metering.

davwood-alpha:
Metered entitlement consumption occurs with or without GAA parameters.

Adjust EntitlementsManager to initialize the MeterToastApi using the MeterClientType.

Adjust MeterToastApi to use MeterClientType rather than an explicit iframeUrl and iframeUrlParams.

MeterToastApi always supplies a publication_id and origin on the querystring.

Passing of Entitlements to consumeMeter_ is redundant. Now, just pass Entititlement, and an optional
decoded JWT.

Fix comment on enablesThisWithGoogleMetering().


src/runtime/entitlements-manager.js
```diff
@@ -237,33 +237,41 @@ export class EntitlementsManager {
 
   /**
    * Sends a pingback that marks a metering entitlement as used.
-   * @param {!Entitlements} entitlements
+   * @param {!Entitlement|null} entitlement
+   * @param {JsonObject|null|undefined} jwtContents
    */
-  consumeMeter_(entitlements) {
-    const entitlement = entitlements.getEntitlementForThis();
+  consumeMeter_(entitlement, jwtContents) {
```
hartmannr76, line 243: I don't think you need to pass in this `jwtContents` since it's the `entitlement.subscriptionToken`. Could just do the JWT decoding as part of this function

hartmannr76: Nvm, I see you also use it below. Maybe we can move the decode function to be inside the `Entitlements` object

src/proto/api_messages.js
```diff
@@ -225,7 +226,7 @@ class AccountCreationRequest {
    */
   toArray(includeLabel = true) {
     const arr = [
-        this.complete_, // field 1 - complete
```
hartmannr76, line 228: I think these formatting changes can be backed out since this file is just autogenerated and copy-pasted over from the generator.

src/runtime/entitlements-manager.js
```diff
@@ -237,33 +237,41 @@ export class EntitlementsManager {
...
+    const gaaToken = queryStringHasFreshGaaParams(this.win_.location.search)
+      ? this.getGaaToken_()
+      : '';
 
     this.deps_
       .eventManager()
       .logSwgEvent(AnalyticsEvent.EVENT_UNLOCKED_BY_METER, false);
 
-    const token = this.getGaaToken_();
+    let entitlementSource = EntitlementSource.GOOGLE_SHOWCASE_METERING_SERVICE;
```
jpettitt, line 257: this feels kinda klunky.   If there is no gaa token and it's not METERED_BY_GOOGLE we should bail rather than pinging back a showcase meter without gaa params.  I suggest moving the gaa token check abpove in an `else` here and if there is no gaa token bail that way we only ever affirmatively set the `entitlmentSource`

src/runtime/entitlements-manager.js
jpettitt: agree is there a case where we need to pass in a jwt that is not from the subscriptionToken? - if there isn't then we should get it internally to this function.

src/proto/api_messages.js
hartmannr76: Thats my guess. This file shouldn't be modified by an editor since it's only copy-pasted out. You can reset it back to the main line with `git checkout origin/main src/proto/api_messages.js` or `git checkout me/main src/proto/api_messages.js` if you're using the `me/team` remote setup from the docs

src/runtime/entitlements-manager.js
hartmannr76: JWT decoding looks like [it's synchronous](https://github.com/subscriptions-project/swg-js/blob/f9ff7b33d81be0804d050556c7c8038a848d0588/src/utils/jwt.js#L39). What I would propose is passing the decoded token into the `Entitlements` constructor (or let the Entitlements do the decoding), this way you can just expose a `getToken` function to access it.

This way, you're only decoding once when you create the entitlement and you don't have to pass around this additional `jwtContents` because you can just access it off the `entitlement` variable you already have access to.

How does that sound?

src/proto/api_messages.js
davwood-alpha: Oh wow. I don't have time to fix this, but someone should really add a warning to this file with a link to the documentation.

src/runtime/entitlements-manager.js
davwood-alpha: It sounds quite awesome, actually. Done.

jpettitt (approved):

src/runtime/entitlements-manager.js
davwood-alpha: My "Done" submitted prematurely. The commit with this refactor is now in for your review.

src/api/entitlements.js
```diff
@@ -356,30 +362,45 @@ export class Entitlement {
...
     const products = json['products'] || [];
     const subscriptionToken = json['subscriptionToken'];
-    return new Entitlement(source, products, subscriptionToken);
+    let subscriptionTokenContents;
+    try {
+      subscriptionTokenContents = subscriptionToken
+        ? jwtHelper.decode(subscriptionToken)
+        : null;
+    } catch (e) {
+      subscriptionTokenContents = null;
```
hartmannr76, line 381: Maybe set to the passed in contents instead?

jpettitt (approved):

src/api/entitlements.js
davwood-alpha: So in these cases, the JWT decode has failed (381), or, there were was nothing to decode (379). 

I'm sorry to say this, but I don't understand what you're suggesting - what are the passed in contents in the latter case? Do you mean, in the former case, put the non-decodable string contents into the field intended to hold decoded token contents? That doesn't seem right.

hartmannr76: oh, nvm. the token itself is still accessible via `subscriptionToken`; so yeah, this doesnt make sense. ignore

hartmannr76 (approved):
LGTM

PR #2094, opened 2022-01-19
Append user token to the audience action iframe url

longmelon:


src/runtime/audience-action-flow.js
```diff
@@ -91,17 +91,25 @@ export class AudienceActionFlow {
...
-        'supportsEventManager': true,
-        'productType': this.productType_,
-      }),
-      /* shouldFadeBody */ true
-    );
+    /** @private @const {!Promise<?ActivityIframeView>} */
+    this.activityIframeViewPromise_ = deps
+      .storage()
+      .get(Constants.USER_TOKEN, true)
+      .then((token) => {
```
jpettitt, line 98: storage().get() can return NULL which will result in sud=NULL in your URL - is the your intent and can the backend cope with that?

[we can't guarantee that some extension hasn't trashed storage or if we are in an iframe on safari storage doesn't persist across sessions]

longmelon: In line 168, if storage().get() returns null, we will set empty string to the sut param.

jpettitt (approved):

src/runtime/audience-action-flow.js
```diff
@@ -91,17 +91,25 @@ export class AudienceActionFlow {
...
-      feArgs({
-        'supportsEventManager': true,
-        'productType': this.productType_,
-      }),
-      /* shouldFadeBody */ true
-    );
+    /** @private @const {!Promise<?ActivityIframeView>} */
+    this.activityIframeViewPromise_ = deps
+      .storage()
+      .get(Constants.USER_TOKEN, true)
```
hartmannr76, line 97: I think you can actually just pass it into the ctor. we have the entitlements response when we create this flow

src/runtime/audience-action-flow.js
```diff
@@ -110,45 +118,50 @@ export class AudienceActionFlow {
...
-    this.activityIframeView_.on(CompleteAudienceActionResponse, (response) =>
-      this.handleCompleteAudienceActionResponse_(response)
-    );
-
-    this.activityIframeView_.on(
-      AlreadySubscribedResponse,
-      this.handleLinkRequest_.bind(this)
-    );
-
-    const {fallback} = this.params_;
```
hartmannr76, line 126: 👍

adolgachev (approved):

src/runtime/audience-action-flow.js
longmelon: Looks like the entitlements object here doesn't contain the user token.

PR #2099, opened 2022-01-22
Remove params.encryption.swgUserToken.

davwood-alpha:
SwgUserTokens are stored when updated, and retrieved from storage when needed. Following this change, there is no alternative pathway for the SwgUserToken to be used a flow such as entitlements check.


ChrisAntaki (approved):

ChrisAntaki:
The less code the better 👍

davwood-alpha:
Friendly ping to the other reviewers. It would be helpful to us if we could have some additional feedback on this today.

alin04 (approved):

jpettitt (approved):

jpettitt:
fix codecov delta

PR #2106, opened 2022-01-25
feat(basic-runtime): Setup default entitlements handler to consume metered reads

hartmannr76:
Adds a default `setOnEntitlementsResponse` to the basic runtime that will consume a metered read if an entitlement provided enables the page via metering.

FYI @longmelon @davwood-alpha

### DIFF
--- a/src/runtime/basic-runtime-test.js
+++ b/src/runtime/basic-runtime-test.js
@@ -186,6 +186,49 @@ describes.realWin('installBasicRuntime', {}, (env) => {
       expect(basicSubscriptions).to.have.property(name);
     }
   });
+
+  describe('default onEntitlementsResponse', () => {
+    let entitlements;
+    let configuredCallback;
+
+    beforeEach(() => {
+      sandbox
+        .stub(BasicRuntime.prototype, 'setOnEntitlementsResponse')
+        .callsFake((callback) => {
+          configuredCallback = callback;
+        });
+      entitlements = {
+        consume: sandbox.stub(),
+        clone: () => entitlements,
+      };
+    });
+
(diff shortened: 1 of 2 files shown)


jpettitt:
Shouldn't this live in the publisher code (or swg-basic?)

Did you confirm this won't consume showcase meters?

src/runtime/basic-runtime-test.js
```diff
@@ -186,6 +186,49 @@ describes.realWin('installBasicRuntime', {}, (env) => {
...
+
+    beforeEach(() => {
+      sandbox
+        .stub(BasicRuntime.prototype, 'setOnEntitlementsResponse')
+        .callsFake((callback) => {
+          configuredCallback = callback;
+        });
+      entitlements = {
+        consume: sandbox.stub(),
+        clone: () => entitlements,
```
ChrisAntaki, line 202: Could we remove this `clone` method? Looks like the tests never call it, so it's reducing the coverage %
https://app.codecov.io/gh/subscriptions-project/swg-js/compare/2106/diff

ChrisAntaki:
Since this PR changes `basic-runtime.js`, I think the showcase meters should be safe

jpettitt:
@ChrisAntaki if we have a showcase pub using swg-basic what will happen?  Presumably we need to add a bunch of code to support that anyway right?  If so we're good here.

src/runtime/basic-runtime-test.js
hartmannr76: Whoops, yeah. Removed. Had left that before I decided to just stub out and capture the configured callback

ChrisAntaki (approved):

jpettitt (approved):

PR #2120, opened 2022-01-28
feat: we should explicitly opt-in to Google metering

hartmannr76:
Instead of always including metering parameters by default, we should require that a page explicitly opt-in to this functionality. It will be enabled by default for SwG Basic runtime.

### DIFF
--- a/src/runtime/basic-runtime-test.js
+++ b/src/runtime/basic-runtime-test.js
@@ -976,5 +976,16 @@ describes.realWin('BasicConfiguredRuntime', {}, (env) => {
       expect(isExperimentOn(win, ExperimentFlags.LOGGING_AUDIENCE_ACTIVITY)).to
         .be.false;
     });
+
+    it('should enable Google controlled metering on configuration', () => {
+      const entitlementsStub = sandbox.stub(
+        EntitlementsManager.prototype,
+        'enableGoogleControlledMetering'
+      );
+
+      configuredBasicRuntime = new ConfiguredBasicRuntime(win, pageConfig);
+
+      expect(entitlementsStub).to.be.calledOnce;
+    });
   });
 });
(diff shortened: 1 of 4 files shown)


src/runtime/basic-runtime-test.js
```diff
@@ -976,5 +976,16 @@ describes.realWin('BasicConfiguredRuntime', {}, (env) => {
       expect(isExperimentOn(win, ExperimentFlags.LOGGING_AUDIENCE_ACTIVITY)).to
         .be.false;
     });
+
+    it('should enable Google controlled metering on configuration', () => {
```
davwood-alpha, line 980: I probably would phrase "Google controlled" as "METERED_ BY_GOOGLE" to be more specific, here and in the symbols/constants. It's better to avoid introducing a new phrase/concept like "Google controlled" in favor of an existing one, wherever possible.

src/runtime/entitlements-manager-test.js
```diff
@@ -1040,81 +1032,138 @@ describes.realWin('EntitlementsManager', {}, (env) => {
...
+                  })
+                ),
+            })
+          );
+
+        manager.enableGoogleControlledMetering();
+        const ents = await manager.getEntitlements({
+          metering: {
+            state: {
+              id: 'u1',
```
davwood-alpha, line 1149: This is allowed in this mode, but sending publisher-provided IDs or attributes in this style with METERED_BY_GOOGLE will probably be unlikely or unusual, and you could argue it's not supported at this time. Perhaps remove?

davwood-alpha (approved):

src/runtime/basic-runtime-test.js
hartmannr76: Fair point. renamed to `enableMeteredByGoogle`

src/runtime/entitlements-manager-test.js
hartmannr76: We can't exercise this code-path unless a state is provided

hartmannr76: I think you're right that we shouldn't explicitly support this case today. Will just replace if a state was provided

adolgachev (approved):

PR #2128, opened 2022-01-31
Add the ability to fetch experiment config flags from the article end…

polyakovn:
…point. Thank-js (and in the future, swg-js) will use this method to fetch experiment flags in their runtimes, and populate the flags in experiments.js. For reference, cl/424628526 populates the experiment configuration returned by the article endpoint.


src/runtime/entitlements-manager.js
```diff
@@ -456,6 +461,23 @@ export class EntitlementsManager {
...
+   */
+  getExperimentConfigFlags() {
+    return this.getArticle().then((article) => {
+      const expConfig = article['experimentConfig'];
+      if (expConfig != null) {
+        const expFlags = expConfig['experimentFlags'];
+        if (expFlags != null) {
+          return expFlags;
+        }
+        return [];
```
ChrisAntaki, line 476: Could you add a test that covers this line?
https://app.codecov.io/gh/subscriptions-project/swg-js/compare/2128/diff

polyakovn: Yes! Done.

src/runtime/entitlements-manager.js
```diff
@@ -465,6 +470,23 @@ export class EntitlementsManager {
...
```
ChrisAntaki, line 485: Should the method return `[]` if `expConfig` is `null` ?

polyakovn: Ah yes, good catch! Updated & added another test

ChrisAntaki (approved):

src/runtime/entitlements-manager.js
ChrisAntaki: Thanks!

PR #2133, opened 2022-02-01
Move gaaNotifySignIn to GaaGoogle3pSignInButton

ed-bird:
Move gaaNotifySignIn to GaaGoogle3pSignInButton class so that it can be called.

At the moment it is exported but not declared in https://github.com/subscriptions-project/swg-js/blob/main/src/gaa-main.js so it is currently unusable.

Publishers will be able to access now with `GaaGoogle3pSignInButton.gaaNotifySignIn()`

### DIFF
--- a/src/utils/gaa.js
+++ b/src/utils/gaa.js
@@ -1181,18 +1181,17 @@ export class GaaGoogle3pSignInButton {
       }
     });
   }
-}
-
-/**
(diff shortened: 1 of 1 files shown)


ChrisAntaki:
CodeCov is reporting some coverage regressions

ChrisAntaki (approved):

PR #2134, opened 2022-02-01
Should append `sut` querystring in pingback entitlement request

longmelon:
`sut` querystring is missing in the post entitlement request which causes unwanted scenario that a new user token is created when registering user activity to the metering service.


davwood-alpha (approved):
LGTM - nice work!

adolgachev (approved):

PR #2150, opened 2022-02-11
Smart Button: Improve coverage and remove extra check

ChrisAntaki:


src/runtime/smart-button-api.js
```diff
@@ -93,12 +93,7 @@ export class SmartSubscriptionButtonApi {
    */
   handleSmartBoxClick_(smartBoxMessage) {
     if (smartBoxMessage && smartBoxMessage.getIsClicked()) {
-      if (!this.callback_) {
```
ChrisAntaki, line 96: I think this check is unnecessary. The only instantiator for the Smart Button always defines a callback

jpettitt (approved):

PR #2189, opened 2022-03-07
Fix the js error when parsing invalid json

longmelon:
The current code caused showing js error in the dev console because we did not properly catch the error when parsing invalid json.

### DIFF
--- a/src/runtime/fetcher.js
+++ b/src/runtime/fetcher.js
@@ -89,9 +89,18 @@ export class XhrFetcher {
       credentials: 'include',
       body: 'f.req=' + serializeProtoMessageForUrl(message),
     });
-    return this.fetch(url, init).then(
-      (response) => (response && response.json()) || {}
-    );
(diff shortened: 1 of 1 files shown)


adolgachev (approved):

jpettitt (commented):
Add a log message when if tails so pub knows

src/runtime/fetcher.js
```diff
@@ -89,9 +89,18 @@ export class XhrFetcher {
...
-    );
+    return this.fetch(url, init).then((response) => {
+      if (!response) {
+        return {};
+      }
+      return response.text().then((text) => {
+        try {
+          return parseJson(text);
+        } catch (e) {
+          return {};
```
jpettitt, line 100: log message here - should include the url that caused it.

longmelon: https://github.com/subscriptions-project/swg-js/pull/2192

PR #2192, opened 2022-03-08
Log error if it fails to parse json

longmelon:


adolgachev (approved):

jpettitt:
Can you switch it to throw async - see #2191

longmelon:
> Can you switch it to throw async - see #2191

Done.

jpettitt (approved):

PR #2194, opened 2022-03-10
Updates Showcase translations

xiexr151e:
This updates "Already have an account?" to "Already registered? Sign in"

For context, please see b/183593738. Note that translations are actually not in place yet.

### DIFF
--- a/package.json
+++ b/package.json
@@ -108,7 +108,7 @@
     "nodemon": "2.0.15",
     "plugin-error": "1.0.1",
     "postcss": "8.4.12",
-    "postcss-import": "14.0.2",
+    "postcss-import": "14.1.0",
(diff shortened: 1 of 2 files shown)


adolgachev (changes requested):
Pending reply to John's comment. I was assuming wording changed and then another process to get them translated?

jpettitt:
Don't merge this - we need to put in the actual translations for every language.  You can do a PR for the english version and then once you have the translations for the other languages add those in a separate PR if you want.

xiexr151e:
Okay, I'll keep this PR as-is right now.

adolgachev:
That process is centered more on adding strings and languages. This is an existing string being modified which would turn English in our next release.

PR #2212, opened 2022-03-22
fix swg user token query parameter

ksebina:


src/runtime/audience-activity-listener.js
```diff
@@ -75,24 +74,26 @@ export class AudienceActivityEventListener {
...
-      const url = serviceUrl(
-        '/publication/' +
-          pubId +
-          '/audienceactivitylogs' +
-          '&sut=' +
-          Constants.USER_TOKEN
-      );
-      this.fetcher_.sendBeacon(url, audienceActivityClientLogsRequest);
-    }
+    this.storage_.get(Constants.USER_TOKEN, true).then((swgUserToken) => {
```
oyj9109, line 77: For a slight performance improvement, I'd recommend checking event type out of the promise, so that we don't even have to enter the promise statement unnecessary for the majority of event types that are not in our allowlist.
```
if (event type not in audienceActivityLoggingEvents) {
  return;
}
this.storage_.get()...
```

oyj9109 (approved):

src/runtime/audience-activity-listener.js
```diff
@@ -76,23 +75,27 @@ export class AudienceActivityEventListener {
...
-      const audienceActivityClientLogsRequest = this.createLogRequest_(event);
-      const url = serviceUrl(
-        '/publication/' +
-          pubId +
-          '/audienceactivitylogs' +
-          '&sut=' +
-          Constants.USER_TOKEN
-      );
-      this.fetcher_.sendBeacon(url, audienceActivityClientLogsRequest);
+      return;
```
adolgachev, line 81: I think I'd prefer to wrap the below in the if check (closer to original) instead of returning for more clarity. In either case, a short comment making it explicit that checking first to avoid promise and a unit-test verifying promise not called would be great.

adolgachev (changes requested):

adolgachev (approved):

PR #2216, opened 2022-03-25
Allow redirect mode in GaaGoogle3pSignInButton

teiswit:
Add an option to open authorizationUrl in the same window to accommodate redirect flow sign-in. This works better on webviews.

### DIFF
--- a/src/utils/gaa-test.js
+++ b/src/utils/gaa-test.js
@@ -1210,6 +1210,7 @@ describes.realWin('GaaGoogle3pSignInButton', {}, () => {
       style.remove();
     }
 
+    self.document.getElementById(GOOGLE_3P_SIGN_IN_BUTTON_ID).remove();
     self.console.warn.restore();
   });
 
@@ -1255,7 +1256,8 @@ describes.realWin('GaaGoogle3pSignInButton', {}, () => {
 
     it('sends post message with button click event', async () => {
       // Show button.
-      GaaGoogle3pSignInButton.show({allowedOrigins}, GOOGLE_3P_AUTH_URL);
+      GaaGoogle3pSignInButton.show({
+        allowedOrigins, GOOGLE_3P_AUTH_URL, redirectMode: false});
       clock.tick(100);
       await tick(10);
 
--- a/src/utils/gaa-test.js
+++ b/src/utils/gaa-test.js
@@ -1209,7 +1209,6 @@ describes.realWin('GaaGoogle3pSignInButton', {}, () => {
     for (const style of [...self.document.head.querySelectorAll('style')]) {
       style.remove();
     }
-    self.document.getElementById(GOOGLE_3P_SIGN_IN_BUTTON_ID).remove();
 
     self.console.warn.restore();
   });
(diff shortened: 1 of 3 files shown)


src/utils/gaa.js
```diff
@@ -1099,9 +1102,9 @@ export class GaaGoogle3pSignInButton {
   /**
    * Renders the third party Google Sign-In button for external authentication.
    * @nocollapse
-   * @param {{ allowedOrigins: !Array<string>, authorizationUrl: string }} params
+   * @param {{ allowedOrigins: !Array<string>, authorizationUrl: string, redirectMode: boolean }} params
    */
-  static show({allowedOrigins, authorizationUrl}) {
+  static show({allowedOrigins, authorizationUrl, redirectMode = true}) {
```
adolgachev, line 1107: Confirming that changing the default behavior is indeed desired? Also, can you please elaborate and document when you would want to set the flag to false?

adolgachev (changes requested):

jpettitt:
Is there a doc or bug that explains the thinking behind this? It's not clear to me what the problem we are trying to solve is.

src/utils/gaa.js
jpettitt: I'm really uncomfortable changing the default behavior of published code without a very clear analysis of why and if it will break existing users implementations.

teiswit: I've changed the redirectMode to false to avoid breaking any prior implementation. Added explanation in jsdoc.

My thinking around this was that we have not published GaaGoogle3pSignInButton in our DevSite yet and that the publishers who are implementing  GaaGoogle3pSignInButton are the ones that we work with directly so we could inform them of the change. But definitely not changing the default behavior is the safest approach. Thanks!

teiswit:
> Is there a doc or bug that explains the thinking behind this? It's not clear to me what the problem we are trying to solve is.

Yes, please see: b/226696040

src/utils/gaa-test.js
```diff
@@ -1290,6 +1295,46 @@ describes.realWin('GaaGoogle3pSignInButton', {}, () => {
...
+      );
+      expect(self.open.getCalls()[0].args[1]).to.equal('_parent');
+    });
+
+    it('should open an authorizationUrl in a new window when redirectMode is false', async () => {
+      // Show button.
+      GaaGoogle3pSignInButton.show({
+        allowedOrigins,
+        authorizationUrl: GOOGLE_3P_AUTH_URL,
+        redirectMode: false,
```
adolgachev, line 1327: I think better to test default behavior when not passing redirectMode to ensure default behavior stays as before.

adolgachev (changes requested):

src/utils/gaa-test.js
teiswit: Done. Please take a look.

adolgachev (approved):

jpettitt (approved):

teiswit:
@adolgachev @jpettitt Thank you for approving the changes. As I don't have write access, could you please merge the pull request? Thank you.

PR #2218, opened 2022-03-28
feat: display Contribution before an action and if outside of prompt window

hartmannr76:
1) For Contribution flows, we should be displaying the Contribution before any potential audience actions
2) Only display an Audience Action for Contribution if 
    a) The Contribution flow has been previously dismissed
    b) The prompt frequency configured indicates that we shouldn't be showing a Contribution (i.e. Contribution prompt always takes precedence)
    c) Displays the first, non-previously dismissed action

Internal reference ([b/226602285](http://b/226602285))


src/runtime/auto-prompt-manager.js
```diff
@@ -345,10 +430,14 @@ export class AutoPromptManager {
...
    * Determines whether a larger, blocking prompt should be shown.
    * @param {!../api/entitlements.Entitlements} entitlements
+   * @param {!boolean} hasPotentialAudienceAction
    * @returns {boolean}
    */
-  shouldShowLockedContentPrompt_(entitlements) {
-    return this.pageConfig_.isLocked() && !entitlements.enablesThis();
+  shouldShowBlockingPrompt_(entitlements, hasPotentialAudienceAction) {
+    return (
+      (this.pageConfig_.isLocked() || hasPotentialAudienceAction) &&
```
longmelon, line 438: Why is this needed?

src/runtime/auto-prompt-manager.js
```diff
@@ -32,17 +33,13 @@ const impressionEvents = [
   AnalyticsEvent.IMPRESSION_SWG_SUBSCRIPTION_MINI_PROMPT,
   AnalyticsEvent.IMPRESSION_OFFERS,
   AnalyticsEvent.IMPRESSION_CONTRIBUTION_OFFERS,
-  AnalyticsEvent.IMPRESSION_NEWSLETTER_OPT_IN,
-  AnalyticsEvent.IMPRESSION_REGWALL_OPT_IN,
```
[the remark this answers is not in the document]
hartmannr76: The regwall or newsletter will keep showing if the user does not dismiss or complete, yes. The user has to explicitly dismiss for regwall/newsletter to be hidden for good. However, if the user dismisses/views the regwall/newsletter, we also don't want those metrics being attributed to the overall contribution prompt impressions and views hence why I'm removing them

src/runtime/auto-prompt-manager.js
hartmannr76: So the `isLocked` is used for premium content (i.e. not `isAccessibleForFree`). Contribution pages are all marked as `isAccessibleForFree: true` but we still want to show the action if we have one available so I pushed it into here.

This method also checks to see if the user is entitled to the page, in which we never want to show a blocking prompt

longmelon: I want to confirm if I understand correctly why we need to add the `hasPotentialAudienceAction` check here:

- For subscription model, this check should not change anything.
- For contribution model, this is added so that we don't cap regwall / newsletter? For example, if the user dismissed the contribution prompt, `shouldShowAutoPrompt` would be false. So we will need to rely on `shouldShowBlockingPrompt_()` to show the audience action and therefore we need to add the `hasPotentialAudienceAction` check here. A potential downside is that this action will not be capped by any frequency setting and user will keep seeing the action prompt until they explicitly dismiss it (which I guess it's fine since the prompt is blocking the user to read the content and if the user does want to read the content they will either dismiss or complete the prompt).

hartmannr76: Both are correct. If we can show a contribution prompt, we should; regardless of actions hence the additional check in this flow. You are also correct that the user will keep seeing a regwall/newsletter unless it is dismissed explicitly. I imagine this could be useful in the case of someone bouncing on the page and never seeing the prompt again

src/runtime/auto-prompt-manager-test.js
```diff
@@ -907,4 +974,228 @@ describes.realWin('AutoPromptManager', {}, (env) => {
...
+        autoPromptType: AutoPromptType.CONTRIBUTION,
+        alwaysShow: false,
+        displayLargePromptFn: alternatePromptSpy,
+      });
+      await tick(2);
+
+      expect(startSpy).to.not.have.been.called;
+      expect(actionFlowSpy).to.not.have.been.called;
+      expect(alternatePromptSpy).to.not.have.been.called;
+      expect(autoPromptManager.promptDisplayed_).to.equal(null);
```
longmelon, line 1152: > expect(autoPromptManager.promptDisplayed_).to.equal(null);
Should this be equal(AutoPromptType.CONTRIBUTION)?

hartmannr76: That would mean we would keep appending `contribution` to the dismissals, which I wouldn't want. Only need to do it the one time

longmelon: Hmm, not sure I follow.

I might miss something: It seems `this.promptDisplayed_` and `localStorage[dismissedprompts]` never get reset to null when it is in the next Contribution prompt time. How do you prevent `contribution` is not being appended to the dismissals?

Also, for this setup, if the user dismisses the contribution prompt, will we show regwall or newsletter next since the dismissals right now is `contribution,TYPE_REGISTRATION_WALL`? We might probably want to add a test case for it?

longmelon: > How do you prevent contribution from being appended to the dismissals?

Oh nvm, I found that `this.promptDisplayed_` is only set to `contribution` if the dismissals is null.

longmelon: >Also, for this setup, if the user dismisses the contribution prompt, will we show regwall or newsletter next since the dismissals right now is contribution,TYPE_REGISTRATION_WALL? We might probably want to add a test case for it?

Synced offline. In this case, newsletter should be shown next. This followed the product requirement: "if they dismissed a newsletter, they probably just never want to give any other info to the publisher".

longmelon (approved):

jpettitt (approved):

PR #2224, opened 2022-04-04
Disable content scrolling when offers dialog is visible

fikus:
Internal tracking: b/227460883

### DIFF
--- a/src/components/dialog.css
+++ b/src/components/dialog.css
@@ -90,3 +90,16 @@
     margin-left: 0 !important;
   }
 }
+
+/**
+ * Class applied to content page to disable scrolling.
+ */
+body.swg-disable-scroll {
+  margin: 0 !important;
+  height: 100% !important;
+  overflow: hidden !important;
+}
+
+body.swg-disable-scroll * {
+  overflow: hidden !important;
+}
--- a/src/runtime/contributions-flow.js
+++ b/src/runtime/contributions-flow.js
@@ -138,8 +138,30 @@ export class ContributionsFlow {
       );
       activityIframeView.on(SkuSelectedResponse, this.startPayFlow_.bind(this));
       this.activityIframeView_ = activityIframeView;
-      return this.dialogManager_.openView(this.activityIframeView_);
+      return this.clientConfigManager_
+        .getClientConfig()
+        .then((clientConfig) => {
+          if (!this.activityIframeView_) {
+            return;
+          }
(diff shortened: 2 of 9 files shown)


adolgachev (approved):

fikus:
@jpettitt can you take a look, please?

src/runtime/contributions-flow.js
```diff
@@ -73,6 +76,7 @@ export class ContributionsFlow {
                 'skus': (options && options.skus) || null,
                 'isClosable': isClosable,
                 'supportsEventManager': true,
+                'shouldDisableBodyScrolling': shouldDisableBodyScrolling,
```
hartmannr76, line 79: I think these are args that are passed into the iframe to receive. Any reason this prob can't just be passed into as a `DialogConfig` setting? https://github.com/subscriptions-project/swg-js/blob/f56b586bd4cb92c443b2db1667e1cc677123d8c6/src/components/dialog.js#L69-L85
I imagine it'd reduce the amount of filed being touched for that change

fikus: Updated. Thanks for the suggestion!

src/ui/activity-iframe-view.js
```diff
@@ -33,7 +33,7 @@ export class ActivityIframeView extends View {
    * @param {!Window} win
    * @param {!../components/activities.ActivityPorts} activityPorts
    * @param {string} src
-   * @param {!Object<string, ?>=} args
+   * @param {!Object<string, ?>=} args Additional data to be passed to the iframe.
```
hartmannr76, line 36: +1 thank you for the comment

hartmannr76 (approved):

src/components/dialog.css
```diff
@@ -90,3 +90,16 @@
     margin-left: 0 !important;
   }
 }
+
+/**
+ * Class applied to content page to disable scrolling.
+ */
+body.swg-disable-scroll {
+  margin: 0 !important;
```
jpettitt, line 98: Wondering if the margin 0 might screw up some sites (I know I suggested it but I'm not 100% sure it's needed to stop scrolling)

fikus: I'm also not sure if that is needed. I can remove it, to err on the side of being less intrusive.

jpettitt: Probably a good idea, we can always add it if needed later.

PR #2226, opened 2022-04-05
Use ImpressionConfig instead of maxImpressionsPerWeek for auto prompt

longmelon:
This change is to support the updated prompt frequency configs:

1. Maximum number of impressions can be capped not only by week. For example, for medium frequency, the maximum number of impressions is 2 per 3 days.
2. Introduce `backoffSeconds` (similar to the dismissal case) to prevent the prompt showing in consecutive page loads.

Internal references: b/226132329, b/226131635

### DIFF
--- a/src/model/auto-prompt-config.js
+++ b/src/model/auto-prompt-config.js
@@ -15,30 +15,49 @@
  */
 
 /**
+ * @typedef {{
+ *   displayDelaySeconds: (number|undefined),
+ *   dismissalBackoffSeconds: (number|undefined),
+ *   maxDismissalsPerWeek: (number|undefined),
+ *   maxDismissalsResultingHideSeconds: (number|undefined),
+ *   impressionBackoffSeconds: (number|undefined),
+ *   maxImpressions: (number|undefined),
+ *   maxImpressionsResultingHideSeconds: (number|undefined),
+ * }}
+ */
+export let AutoPromptConfigparams;
+
+/**
  * Container for the auto prompt configuation details.
  */
 export class AutoPromptConfig {
   /**
-   * @param {number|undefined} maxImpressionsPerWeek
+   * @param {!AutoPromptConfigparams=} params
    */
-  constructor(
-    maxImpressionsPerWeek,
+  constructor({
     displayDelaySeconds,
-    backoffSeconds,
+    dismissalBackoffSeconds,
     maxDismissalsPerWeek,
-    maxDismissalsResultingHideSeconds
-  ) {
-    /** @const {number|undefined} */
-    this.maxImpressionsPerWeek = maxImpressionsPerWeek;
-
+    maxDismissalsResultingHideSeconds,
+    impressionBackoffSeconds,
+    maxImpressions,
+    maxImpressionsResultingHideSeconds,
+  } = {}) {
     /** @const {!ClientDisplayTrigger} */
     this.clientDisplayTrigger = new ClientDisplayTrigger(displayDelaySeconds);
 
     /** @const {!ExplicitDismissalConfig} */
     this.explicitDismissalConfig = new ExplicitDismissalConfig(
-      backoffSeconds,
+      dismissalBackoffSeconds,
       maxDismissalsPerWeek,
       maxDismissalsResultingHideSeconds
+    );
+
+    /** @const {!ImpressionConfig} */
+    this.impressionConfig = new ImpressionConfig(
+      impressionBackoffSeconds,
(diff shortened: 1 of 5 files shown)


src/runtime/auto-prompt-manager.js
```diff
@@ -258,7 +258,7 @@ export class AutoPromptManager {
     }
 
     // Fetched config returned no maximum cap.
-    if (autoPromptConfig.maxImpressionsPerWeek === undefined) {
+    if (autoPromptConfig.impressionConfig.maxImpressions === undefined) {
```
hartmannr76, line 261: optional chaining?
```
if (autoPromptConfig.impressionConfig?.maxImpressions === undefined) {
```

longmelon: Probably not needed here since autoPromptConfig.impressionConfig is never undefined.

src/runtime/client-config-manager.js
```diff
@@ -192,13 +192,23 @@ export class ClientConfigManager {
     const autoPromptConfigJson = json['autoPromptConfig'];
     let autoPromptConfig = undefined;
     if (autoPromptConfigJson) {
-      autoPromptConfig = new AutoPromptConfig(
-        autoPromptConfigJson.maxImpressionsPerWeek,
```
hartmannr76, line 196: Hmmmm doesn't this mean the response has to be changed and synchronized with this JS change? We may want to build the new config type with the old version of the autoPrompt JSON coming from the backend. Another major reason for this (at least in the short term) is that a breaking change in the JSON response will effect older versions of swg-basic

hartmannr76: Caught up offline. The new field is already being passed down on the backend so new JS should be fine to accept impression config

longmelon: The backend is still sending the maxImpressionsPerWeek field so publishers using the old version of swg-js can still consume this field. Looks like at least in the short term, it's not safe to remove that from the backend though.

src/model/auto-prompt-config.js
```diff
@@ -14,32 +14,51 @@
...
+ *   displayDelaySeconds: (number|undefined),
+ *   dismissalBackoffSeconds: (number|undefined),
+ *   maxDismissalsPerWeek: (number|undefined),
+ *   maxDismissalsResultingHideSeconds: (number|undefined),
+ *   impressionBackoffSeconds: (number|undefined),
+ *   maxImpressions: (number|undefined),
+ *   maxImpressionsResultingHideSeconds: (number|undefined),
+ * }}
+ */
+export let AutoPromptConfigparams;
```
hartmannr76, line 28: `AutoPromptConfigParams`

longmelon: Good catch. Done.

src/runtime/auto-prompt-manager.js
```diff
@@ -298,13 +298,33 @@ export class AutoPromptManager {
           return false;
         }
 
-        // If the user has reached maxImpressionsPerWeek, don't show the prompt.
+        // If the user has reached the maxImpressions, and
+        // maxImpressionsResultingHideSeconds has not yet passed, don't show the
+        // prompt.
+        if (
+          autoPromptConfig.impressionConfig.maxImpressions !== undefined &&
```
hartmannr76, line 305: You should just be able to do `if (autoPromptConfig.impressionConfig.maxImpressions && ..)`

src/runtime/auto-prompt-manager.js
```diff
@@ -298,13 +298,33 @@ export class AutoPromptManager {
...
+          impressions.length >=
+            autoPromptConfig.impressionConfig.maxImpressions &&
+          Date.now() - impressions[impressions.length - 1] <
```
hartmannr76, line 308: I see `impressions[impressions.length - 1]` and `dismissals[dismissals.length - 1]` getting used a lot. Maybe pull them out earlier and have them as `lastDismissal` and `lastImpression`?

src/runtime/auto-prompt-manager.js
longmelon: Oh yeah, absolutely. Done.

src/runtime/auto-prompt-manager.js
longmelon: Definitely better for readability. Done.

adolgachev (approved):

hartmannr76 (approved):
nice

PR #2228, opened 2022-04-07
Add a new callback to trigger the non-native offers flow

longmelon:
For the meter prompt used by basic, when "Become a subscriber" is clicked, the offers flow should be triggered. Previously, `setOnNativeSubscribeRequest` was used for starting the offers flow. There are two issues by this approach:

1. The "View on" native offer card is always displayed.
2. The publisher would not be able to implement its own subscription callback.

This PR introduces a new callback (which is not exposed publicly) and the basic runtime will set it to trigger the "non-native" offer flows. In future, any prompts that want to start the offers flow but doesn't need the native option can use this new callback.

Internal reference: b/227343269

### DIFF
--- a/src/runtime/basic-runtime.js
+++ b/src/runtime/basic-runtime.js
@@ -266,13 +266,6 @@ export class BasicRuntime {
       runtime.processEntitlements()
     );
   }
-
-  /** Sets the callback when the offers flow is requested. */
-  setOnOffersFlowRequest(callback) {
-    return this.configured_(false).then((runtime) =>
-      runtime.setOnOffersFlowRequest(callback)
-    );
-  }
 }
 
 /**
@@ -325,7 +318,7 @@ export class ConfiguredBasicRuntime {
     this.entitlementsManager().enableMeteredByGoogle();
 
     // Handle clicks on the Metering Toast's "Subscribe" button.
-    this.setOnOffersFlowRequest(() => {
+    this.setOnOffersFlowRequest_(() => {
       this.configuredClassicRuntime_.showOffers();
     });
 
@@ -484,11 +477,6 @@ export class ConfiguredBasicRuntime {
     );
   }
 
-  /** Sets the callback when the offers flow is requested. */
-  setOnOffersFlowRequest(callback) {
-    this.callbacks().setOnOffersFlowRequest(callback);
-  }
-
   /**
    * Handler function to process EntitlementsResponse.
    * @param {!../components/activities.ActivityPortDef} port
@@ -610,6 +598,15 @@ export class ConfiguredBasicRuntime {
           }
         );
       });
+  }
+
+  /**
+   * Sets the callback when the offers flow is requested.
+   * @param {function()} callback
+   * @private
+   */
+  setOnOffersFlowRequest_(callback) {
+    this.callbacks().setOnOffersFlowRequest(callback);
   }
 }
 
--- a/src/runtime/basic-runtime.js
+++ b/src/runtime/basic-runtime.js
@@ -266,6 +266,13 @@ export class BasicRuntime {
       runtime.processEntitlements()
     );
   }
+
+  /** Sets the callback when the offers flow is requested. */
+  setOnOffersFlowRequest(callback) {
+    return this.configured_(false).then((runtime) =>
+      runtime.setOnOffersFlowRequest(callback)
+    );
+  }
 }
 
 /**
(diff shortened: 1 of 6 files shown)


src/runtime/basic-runtime.js
```diff
@@ -266,6 +266,13 @@ export class BasicRuntime {
       runtime.processEntitlements()
     );
   }
+
+  /** Sets the callback when the offers flow is requested. */
+  setOnOffersFlowRequest(callback) {
```
fikus, line 271: Is there a need to expose this? It seems that calling this would break the intended behavior implemented in the ConfiguredBasicRuntime constructor.

src/runtime/basic-runtime.js
```diff
@@ -477,6 +484,11 @@ export class ConfiguredBasicRuntime {
     );
   }
 
+  /** Sets the callback when the offers flow is requested. */
+  setOnOffersFlowRequest(callback) {
```
fikus, line 488: Similarly here, should this be a private method?

fikus, line 488: Similarly here, should this be a private method?

src/runtime/meter-toast-api.js
```diff
@@ -180,7 +180,10 @@ export class MeterToastApi {
           ViewSubscriptionsResponse,
           this.startNativeFlow_.bind(this)
         );
-        if (!this.deps_.callbacks().hasSubscribeRequestCallback()) {
+        if (
+          !this.deps_.callbacks().hasSubscribeRequestCallback() &&
```
fikus, line 184: Is this still dependent on setting a subscribe request callback?

fikus, line 184: Is this still dependent on setting a subscribe request callback?

src/runtime/meter-toast-api.js
```diff
@@ -180,7 +180,10 @@ export class MeterToastApi {
...
+          !this.deps_.callbacks().hasOffersFlowRequestCallback()
+        ) {
           const errorMessage =
             '[swg.js]: `setOnNativeSubscribeRequest` has not been set ' +
```
fikus, line 188: Looks like this error message needs to be updated.

fikus, line 188: Looks like this error message needs to be updated.

src/runtime/meter-toast-api.js
```diff
@@ -327,11 +330,13 @@ export class MeterToastApi {
...
+    this.removeCloseEventListener();
+    // We shouldn't decrement the meter on redirects, so don't call onConsumeCallback.
+    this.onConsumeCallbackHandled_ = true;
     if (response.getNative()) {
-      this.removeCloseEventListener();
-      // We shouldn't decrement the meter on redirects, so don't call onConsumeCallback.
-      this.onConsumeCallbackHandled_ = true;
       this.deps_.callbacks().triggerSubscribeRequest();
+    } else {
+      this.deps_.callbacks().triggerOffersFlowRequest();
```
fikus, line 339: I wouldn't expect startNativeFlow_ to handle non-native.

fikus, line 339: I wouldn't expect startNativeFlow_ to handle non-native.

src/runtime/meter-toast-api.js
longmelon: Close duplicate

src/runtime/meter-toast-api.js
longmelon: Close duplicate

src/runtime/meter-toast-api.js
longmelon: Close duplicate

src/runtime/basic-runtime.js
longmelon: Removed.

src/runtime/basic-runtime.js
longmelon: Set it as private.

src/runtime/meter-toast-api.js
longmelon: For meter toast used by basic, since offers flow request callback is set, the user will be able to subscribe from it directly. For meter toast integrated by the publishers, they still need to call `setOnNativeSubscribeRequest` if they want users to subscribe directly from the metering dialog.

src/runtime/meter-toast-api.js
longmelon: I think this error message is meant to inform the publishers who would like to integrate with this meter toast. For them, the option is still to call `setOnNativeSubscribeRequest`.

src/runtime/meter-toast-api.js
longmelon: Rename it to `startSubscriptionFlow_`

adolgachev (approved):

PR #2229, opened 2022-04-08
Remove referrer check

ed-bird:
See b/228586366 for more context


adolgachev (approved):

adolgachev:
If coverage report still fails, I can merge on your behalf. I see the other tests still call that same method so that tiny drop seems acceptable to override.

ed-bird:
> If coverage report still fails, I can merge on your behalf. I see the other tests still call that same method so that tiny drop seems acceptable to override.

Awesome, thanks @adolgachev !

PR #2238, opened 2022-04-13
Forward purchaseUnavailableRegion flag to subscriptions iframe

fikus:
Internal ref: b/228193791

### DIFF
--- a/src/model/auto-prompt-config.js
+++ b/src/model/auto-prompt-config.js
@@ -127,18 +127,27 @@ export class ImpressionConfig {
 }
 
 /**
- * Predicates of whether or not to show button and prompt.
+ * Predicates to control UI elements.
  */
+// TODO(b/228856600): Move to a more general location.
 export class UiPredicates {
   /**
    * @param {boolean|undefined} canDisplayAutoPrompt
    * @param {boolean|undefined} canDisplayButton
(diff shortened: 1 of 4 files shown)


jpettitt:
Do we need to add this to the button rendering case where it's button->purchase flow with no offers frame?

src/model/auto-prompt-config.js
```diff
@@ -127,18 +127,27 @@ export class ImpressionConfig {
...
 /**
- * Predicates of whether or not to show button and prompt.
+ * Predicates to control UI elements.
  */
+// TODO(b/228856600): Move to a more general location.
 export class UiPredicates {
   /**
    * @param {boolean|undefined} canDisplayAutoPrompt
    * @param {boolean|undefined} canDisplayButton
+   * @param {boolean|undefined} purchaseUnavailableRegion
```
adolgachev, line 137: I think 3 is probably the limit on this but should be okay and not worth changing to object right now. Can you please file another bug (or merge into above) to change to object args though for later?

adolgachev (approved):

src/model/auto-prompt-config.js
fikus: Filed b/229131111.

fikus:
@jpettitt I added you to a thread on the PRD.

PR #2276, opened 2022-05-05
Set forceDisableNative to true for paySwgVersion 3

li-cai:
TwG flows also do not support the native pay flow. This fixes the issue of TwG opening a failing native Pay view when using the new paySwgVersion.


alin04 (commented):
As simple as this is, add a test?

li-cai:
> 

Makes sense! Added test.

PR #2283, opened 2022-05-12
Add reserved user functionality to EA API

ed-bird:


src/utils/gaa.js
```diff
@@ -2045,33 +2063,54 @@ export class GaaMetering {
...
-              'timestamp': newUserState.registrationTimestamp,
+              'timestamp': registrationTimestamp,
             },
           },
         },
       },
     };
   }
 
+  static convertTimestampToSeconds(timestamp) {
```
adolgachev, line 2085: Please put this in date-utils.js

adolgachev (changes requested):

src/utils/gaa.js
ed-bird: Done!

ed-bird:
@adolgachev - please can you re-review the requested changes?

src/utils/gaa.js
```diff
@@ -1735,29 +1744,39 @@ export class GaaMetering {
...
         // User does not any access from publisher or Google so show the standard paywall
         callSwg((subscriptions) => {
-          subscriptions.setShowcaseEntitlement({
-            entitlement: ShowcaseEvent.EVENT_SHOWCASE_NO_ENTITLEMENTS_PAYWALL,
-            isUserRegistered: GaaMetering.isUserRegistered(
-              GaaMetering.userState
-            ),
-          });
+          switch (GaaMetering.userState.paywallReason) {
+            case 'RESERVED_USER':
```
hartmannr76, line 1756: reference the enum value above?

src/utils/gaa.js
```diff
@@ -1735,29 +1744,39 @@ export class GaaMetering {
...
         });
         // Show the paywall
         showPaywall();
       }
     });
   }
 
-  static isUserRegistered(userState) {
-    return userState.id !== undefined && userState.id != '';
+  static isUserRegistered() {
```
hartmannr76, line 1776: This isn't a backwards compatible change. Maybe it should be something like

```
static isUserRegistered(userState) {
    if (userState) {
        return userState.id !== undefined && userState.id != '';
    }

  return (
      GaaMetering.userState.id !== undefined && GaaMetering.userState.id != ''
    );
}
```

src/utils/gaa.js
```diff
@@ -2045,13 +2064,18 @@ export class GaaMetering {
   }
 
   static newUserStateToUserState(newUserState) {
+    // Convert registrationTimestamp
+    const registrationTimestamp = convertPotentialTimestampToSeconds(
```
hartmannr76, line 2068: I feel like we should enforce the timestamp provided to be a certain range, but lets maybe name this `registrationTimestampSeconds` just to be clear

src/utils/gaa.js
```diff
@@ -2108,47 +2134,45 @@ export class GaaMetering {
...
+          if (!('subscriptionTimestamp' in newUserState)) {
+            debugLog(
+              'subscriptionTimestamp is required if userState.grantReason is SUBSCRIBER'
+            );
+
+            noIssues = false;
+          } else if (
+            !(
+              typeof newUserState.subscriptionTimestamp === 'number' &&
+              newUserState.subscriptionTimestamp % 1 === 0
```
hartmannr76, line 2159: You could probably replace both of these with `Number.isInteger(newUserState.subscriptionTimestamp)`. Took me a second to grok what this was meant to read as

src/utils/gaa.js
```diff
@@ -2165,7 +2189,22 @@ export class GaaMetering {
...
+
+      if (PaywallReasonType[newUserState.paywallReason] === undefined) {
+        debugLog(
+          'userState.paywallReason has to be empty or set to RESERVED_USER.'
+        );
+        noIssues = false;
+      }
+    }
+    return noIssues;
   }
```
hartmannr76, line 2208: Not blocking for this PR but this file is getting quite large. We really should think about chunking it up

src/utils/gaa-test.js
```diff
@@ -3722,8 +3750,44 @@ describes.realWin('GaaMetering', {}, () => {
       showPaywall = sandbox.fake();
       GaaMetering.userState = {};
 
-      sandbox.stub(GaaMetering, 'isUserRegistered');
-      GaaMetering.isUserRegistered.returns(true);
+      sandbox.stub(GaaMetering, 'isUserRegistered').returns(true);
+
+      const googleEntitlementsPromise = new Promise((resolve) => {
```
hartmannr76, line 3755: This feels like an off pattern. Any reason something like 
```
Promise.resolve({ consume: () => {}, enablesThis: () => false, enablesThisWithGoogleMetering: () => false })
// or
Promise.resolve({ 
  consume: sandbox.fake,
  enablesThis: sandbox.fake.returns(false),
  enablesThisWithGoogleMetering: sandbox.fake.returns(false) 
});
```
wouldn't work?

ed-bird: Good idea, thanks!

src/utils/gaa.js
ed-bird: Yeah, it would be good to slim it down and remove redundant code

src/utils/gaa.js
ed-bird: I tried this but got a lint error:

```
2153:23  error  no ES2015 "Number" methods and properties allowed to be used  swg-internal/no-es2015-number-props
```

src/utils/gaa.js
ed-bird: Sounds good, thanks!

src/utils/gaa.js
ed-bird: Added a new function: `isCurrentUserRegistered`. Having said that, nobody should be using `isUserRegistered` so it should be safe to make this change

src/utils/gaa.js
ed-bird: Done, thanks!

ed-bird:
Thanks for the review @hartmannr76 !! Please could you take another look?

hartmannr76 (approved):

PR #2295, opened 2022-05-17
swg-basic - setOnEntitlementsResponse() should be available to publishers

hartmannr76:
The previous setup for swg-basic was to set a default handler for `setOnEntitlementsResponse` that would consume a Google meter if one was provided. This meant that if a publisher wanted to set their own handler, they would automatically override what we provided for them.

This change provides a default metering handler in the `EntitlementsManager` (disabled by default for Classic compatibility), which keeps the `setOnEntitlementsResponse` exposed to be set in any way.  In swg-basic, we enable the handler to true by default and provide a `disableDefaultMeteringHandler` to the `init` function. 

Internal reference: b/228050447

### DIFF
--- a/src/runtime/basic-runtime-test.js
+++ b/src/runtime/basic-runtime-test.js
@@ -186,48 +186,6 @@ describes.realWin('installBasicRuntime', {}, (env) => {
       expect(basicSubscriptions).to.have.property(name);
     }
   });
-
-  describe('default onEntitlementsResponse', () => {
-    let entitlements;
-    let configuredCallback;
-
-    beforeEach(() => {
-      sandbox
-        .stub(BasicRuntime.prototype, 'setOnEntitlementsResponse')
-        .callsFake((callback) => {
-          configuredCallback = callback;
-        });
-      entitlements = {
-        consume: sandbox.stub(),
-      };
-    });
-
-    it('should consume if entitlement enables with metering', async () => {
-      entitlements.enablesThisWithGoogleMetering = sandbox.stub().returns(true);
-
-      installBasicRuntime(win);
-      const entitlementsResponse = Promise.resolve(entitlements);
-      configuredCallback(entitlementsResponse);
-      await entitlementsResponse;
-
-      expect(entitlements.enablesThisWithGoogleMetering).to.be.calledOnce;
-      expect(entitlements.consume).to.be.calledOnce;
-    });
-
-    it('should not consume if entitlement is not for metering', async () => {
-      entitlements.enablesThisWithGoogleMetering = sandbox
-        .stub()
-        .returns(false);
-
-      installBasicRuntime(win);
-      const entitlementsResponse = Promise.resolve(entitlements);
-      configuredCallback(entitlementsResponse);
-      await entitlementsResponse;
-
-      expect(entitlements.enablesThisWithGoogleMetering).to.be.calledOnce;
-      expect(entitlements.consume).to.not.be.called;
-    });
-  });
 });
 
 describes.realWin('BasicRuntime', {}, (env) => {
@@ -263,7 +221,7 @@ describes.realWin('BasicRuntime', {}, (env) => {
         isPartOfProductId: 'herald-foo-times.com:basic',
       });
 
-      await basicRuntime.configured_(true);
+      const configuredRuntime = await basicRuntime.configured_(true);
 
       // init should have written the LD+JSON markup.
       const elements = doc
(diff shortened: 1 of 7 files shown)


src/runtime/basic-runtime-test.js
```diff
@@ -274,6 +232,12 @@ describes.realWin('BasicRuntime', {}, (env) => {
       // PageConfigResolver should have been created and attempted to resolve
       // the PageConfig.
       expect(resolveStub).to.be.calledOnce;
+
+      // Default metering handler in entitlements-manager should be enabled
+      // for BasicRuntime.
+      expect(
+        configuredRuntime.entitlementsManager().enableDefaultMeteringHandler_
```
fikus, line 239: Can you rewrite these assertions so they do not check a private variable? (ie, test behavior instead of implementation).

src/runtime/entitlements-manager-test.js
```diff
@@ -2172,6 +2172,82 @@ describes.realWin('EntitlementsManager', {}, (env) => {
...
+
+      it('should NOT consume a Google meter entitlement by default', async () => {
+        const entitlements = await manager.getEntitlements();
+
+        expect(entitlements.enablesThisWithGoogleMetering()).to.be.true;
+      });
+
+      it('should consume an entitlement if default handler is enabled and is Google meter', async () => {
+        // When enabled, we expect the pingback to occur as part of the getEntitlements process
+        manager.enableDefaultMeteringHandler_ = true;
```
fikus, line 2235: Can you call the constructor here instead of setting a private variable?

src/runtime/runtime-test.js
```diff
@@ -1145,6 +1145,20 @@ describes.realWin('ConfiguredRuntime', {}, (env) => {
     ).to.throw();
   });
 
+  it('should pass enableDefaultMeteringHandler into EntitlementsManager during constuction and default to false', () => {
+    runtime = new ConfiguredRuntime(win, config);
+
+    expect(runtime.entitlementsManager().enableDefaultMeteringHandler_).to.be
```
fikus, line 1151: Same comment here re checking private variables in the test.

fikus (changes requested):

src/runtime/entitlements-manager-test.js
```diff
@@ -2172,6 +2172,82 @@ describes.realWin('EntitlementsManager', {}, (env) => {
...
+                    signedEntitlements: 'SIGNED_DATA',
+                    swgUserToken: 'abc',
+                  })
+                ),
+            })
+          );
+      });
+
+      it('should NOT consume a Google meter entitlement by default', async () => {
+        const entitlements = await manager.getEntitlements();
```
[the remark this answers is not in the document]
hartmannr76: I definitely agree that the setup (even assertions included by default) are not clear due to the default before/after functions. We could just ignore the default setups and make our own - we should be pretty diligent at enforcing that pattern if we go down that route since there is already so much test code following it.

The assertions are (unfortunately) just included in the default afterAlls - one option could just be explicitly checking the mocks that have expected functionality we are explicitly setting up. Let me switch to explicitly checking instead of implicit.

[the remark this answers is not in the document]
fikus: I know there are assertions in afterAll blocks checking that all the mocks are called with expected arguments, and agree this is an unfortunate pattern. Assertions should also be moved out of before blocks wherever possible. In this case, the test isn't adding any specific mocks with expected arguments and the only specific assertion is the same as in the test for the other case on L2248, so I'm not seeing what this is actually testing.

src/runtime/basic-runtime-test.js
hartmannr76: Done

src/runtime/entitlements-manager-test.js
hartmannr76: Done

src/runtime/entitlements-manager-test.js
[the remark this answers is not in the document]
hartmannr76: Moved the setup/expectation code into each function respectively. Commented that the specific check that existed here was to assert that the handler code _should_ have triggered if it were enabled

src/runtime/runtime-test.js
hartmannr76: Done

adolgachev (approved):

jpettitt (approved):

PR #2311, opened 2022-05-26
Make registerUserPromise optional for 3P sign-in

teiswit:
Make `registerUserPromise` optional when `authorizationUrl` is provided.
Context: user registration for 3P sign-in will take place on the server-side, therefore it is not needed to use `registerUserPromise`.

Internal reference: b/233581122


fikus:
Can you check the test failures?

teiswit:
The E2E failure was unrelated to the change (backward compatibility issue in Play SKUs), but nevertheless I've fixed the issue.

adolgachev (approved):

PR #2321, opened 2022-05-31
swg.js IFrames should pass sut=<SwgUserToken> and publicationId by query string

mhkawano:
IFrames should pass the SwgUserToken and publicationId by query string. This adds the change in ActivityPort .openIframe(), and removes it in other places where the token is set for iFrames.


src/runtime/entitlements-manager.js
```diff
@@ -850,9 +850,9 @@ export class EntitlementsManager {
         }
 
         // Add swgUserToken param.
-        if (swgUserToken) {
-          url = addQueryParam(url, 'sut', swgUserToken);
-        }
+        //if (swgUserToken) {
```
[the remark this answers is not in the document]
mhkawano: Done. The toast at the end of getEntitlements is getting an iframe, which threw me off. Changed tests to account for multiple get calls to storage for the token.

src/components/activities.js
```diff
@@ -314,7 +317,24 @@ export class ActivityPorts {
     if (addDefaultArguments) {
       args = this.addDefaultArguments(args);
     }
-    return this.openActivityIframePort_(iframe, url, args);
+    return this.deps_
+      .storage()
+      .get(Constants.USER_TOKEN, true)
+      .then((swgUserToken) => {
+        const split = url.split('?');
```
hartmannr76, line 324: I'm pretty sure you don't need to do this and can just do

```javascript
const queryParams = new URL(url).searchParams;
if (swgUserToken && !queryParams.has('sut')) {
    // stuffs
}
```

mhkawano: Done. URL() is picky about the input, so I had to change `/hello` to `https://www.google.com/` in the tests. Don't think that will be an issue, but wanted to bring that up.

src/components/activities-test.js
```diff
@@ -240,6 +242,40 @@ describes.realWin('Activity Components', {}, (env) => {
...
+            iframe,
+            `${url}?sut=${TOKEN}&publicationId=${publicationId}`,
+            {}
+          )
+          .once();
+        await activityPorts.openIframe(iframe, url, {});
+        callMock.verify();
+      });
+
+      it('does not duplicate already existing sut and publicationId', async () => {
```
hartmannr76, line 261: I'd split this into 2 separate tests, 1 for each `if` statement (existing sut, does not replace; existing pulicationId, does not replace). If the logic below is updated to 
```
if (swgUserToken && !queryParams.has('sut') && pubId && !queryParams.has('publicationId') ) { ... }
```
both these tests pass. Also, we should have a test making sure that if the swgUserToken is empty from storage, it should not add to the url

hartmannr76 (approved):
Nice!

src/components/activities-test.js
```diff
@@ -240,6 +242,80 @@ describes.realWin('Activity Components', {}, (env) => {
...
+        callMock.verify();
+      });
+
+      it('does not replace or duplicate already existing sut', async () => {
+        const callMock = sandbox
+          .mock(activityPorts, 'openActivityIframePort_')
+          .expects('openActivityIframePort_')
+          .withExactArgs(
+            iframe,
+            `${url}?sut=Original${TOKEN}&publicationId=${publicationId}`,
```
fikus, line 289: There's no need for this to depend on the value of TOKEN.

src/components/activities-test.js
```diff
@@ -240,6 +242,80 @@ describes.realWin('Activity Components', {}, (env) => {
...
+        callMock.verify();
+      });
+
+      it('does not replace or duplicate already existing publicationId', async () => {
+        const callMock = sandbox
+          .mock(activityPorts, 'openActivityIframePort_')
+          .expects('openActivityIframePort_')
+          .withExactArgs(
+            iframe,
+            `${url}?publicationId=Original${publicationId}&sut=${TOKEN}`,
```
fikus, line 307: Similarly, no reason for this to depend on publicationId.

src/components/activities.js
```diff
@@ -314,7 +317,22 @@ export class ActivityPorts {
     if (addDefaultArguments) {
       args = this.addDefaultArguments(args);
     }
-    return this.openActivityIframePort_(iframe, url, args);
+    return this.deps_
+      .storage()
+      .get(Constants.USER_TOKEN, true)
```
fikus, line 322: nit: Add inline comment explaining mean of true literal.

src/runtime/meter-toast-api.js
```diff
@@ -125,141 +124,127 @@ export class MeterToastApi {
...
-        };
-        if (swgUserToken) {
-          iframeUrlParams['sut'] = swgUserToken;
-        }
+    const iframeUrl =
+      IframeUrlByMeterClientType[
+        this.meterClientType_ ?? MeterClientTypes.LICENSED_BY_GOOGLE
+      ];
+    const iframeUrlParams = {
+      'publicationId': this.deps_.pageConfig().getPublicationId(),
```
fikus, line 135: Can publicationId be removed here?

fikus (changes requested):

src/components/activities-test.js
mhkawano: Done.

src/components/activities-test.js
mhkawano: Done.

src/runtime/meter-toast-api.js
mhkawano: Yes. Done.

src/components/activities.js
mhkawano: Done.

fikus (approved):

PR #2324, opened 2022-06-03
Analytics for twg wordpress plugin

dethstrobe:
We're looking to track the wordpress plugin version for Thank With Google


fikus (changes requested):
Can you revert the formatting changes to the api_messages files to avoid adding noise to the diff?

PR #2369, opened 2022-07-28
feat: include experiment params to service/iframe calls

hartmannr76:
It's hard to trigger experiment flags (both in API calls and on iframes) even though the servers accept them. This will include them in our calls so we can view server experiments from a given page.

Internal reference: b/240682061


fikus (approved):
Can you reference a bug for future reference, and make sure this capability is documented?

hartmannr76:
Done. Updated internal docs and added some information in the `experiments.js` file as well

PR #2377, opened 2022-08-04
Changing analytics event

ksebina:
Created a separate event for the 3P button click and now updating code to log the correct event, representing one of the flows for b/232215758.
This change deals with immediate redirect following the button click, with a delay of very short/unperceivable time using setTimeout.


fikus (changes requested):
Can you add a test exercising the change?

jpettitt (commented):
Was anybody external already using this event under the old name?

jpettitt (approved):

src/utils/gaa-test.js
```diff
@@ -829,6 +829,29 @@ describes.realWin('GaaMeteringRegwall', {}, () => {
...
+    it('check 3p button click', () => {
+      GaaMeteringRegwall.showWithNative3PRegistrationButton({
+        authorizationUrl: GOOGLE_3P_AUTH_URL,
+        caslUrl: CASL_URL,
+      });
+
+      // Click button.
+      self.document.getElementById(GOOGLE_3P_SIGN_IN_BUTTON_ID).click();
+
+      sandbox.stub(self, 'setTimeout').callsFake((callback, timeout) => {
```
hartmannr76, line 842: I think this needs to be setup before you click the button

hartmannr76 (approved):
LGTM. Maybe we should also make a tracking ticket to fix the full downstream calls to be chained. Lots of unmanaged promises

src/utils/gaa.js
```diff
@@ -971,11 +974,13 @@ export class GaaMeteringRegwall {
     buttonEl.addEventListener('click', () => {
       // Track button clicks.
       logEvent({
-        analyticsEvent: AnalyticsEvent.ACTION_SHOWCASE_REGWALL_GSI_CLICK,
+        analyticsEvent: AnalyticsEvent.ACTION_SHOWCASE_REGWALL_3P_BUTTON_CLICK,
         isFromUserAction: true,
       });
       // Redirect user using the parent window.
-      self.open(authorizationUrl, '_parent');
+      self.setTimeout(() => {
```
fikus, line 981: Can you file a bug and add a TODO to clean this up?

fikus (approved):

src/utils/gaa.js
ksebina: Yes! b/242998655

PR #2378, opened 2022-08-16
add clientOption for allowScroll

justinchou-google:
[b/238669349](https://b.corp.google.com/issues/238669349)

Adding client option to allow scrolling behind a subscription or contribution dialog, called `allowScroll`.

### DIFF
--- a/src/runtime/client-config-manager.js
+++ b/src/runtime/client-config-manager.js
@@ -113,6 +113,15 @@ export class ClientConfigManager {
   }
 
   /**
+   * Returns whether scrolling on main page should be allowed when
+   * subscription or contribution dialog is displayed.
+   * @return {boolean}
+   */
+  shouldAllowScroll() {
+    return !!this.clientOptions_.allowScroll;
+  }
+
+  /**
    * Returns whether iframes should also use the language specified in the
    * client options, rather than the default of letting the iframes decide the
    * display language. Note that this will return false if the lang option is
--- a/src/runtime/contributions-flow.js
+++ b/src/runtime/contributions-flow.js
@@ -159,7 +159,7 @@ export class ContributionsFlow {
    * @return {!../components/dialog.DialogConfig}
    */
   getDialogConfig_(clientConfig) {
-    return clientConfig.useUpdatedOfferFlows
+    return clientConfig.useUpdatedOfferFlows && !clientConfig.allowScroll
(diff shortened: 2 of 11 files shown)


src/runtime/client-config-manager.js
```diff
@@ -112,6 +112,15 @@ export class ClientConfigManager {
     return this.clientOptions_.theme || ClientTheme.LIGHT;
   }
 
+  /**
+   * Returns whether scrolling on main page should be allowed when
+   * subscription or contribution dialog is displayed.
+   * @return {boolean}
+   */
+  shouldAllowScroll() {
```
justinchou-google, line 120: may be worth converting this to `shouldDisableScroll` or (some other negation) to match the config `shouldDisableBodyScrolling`

justinchou-google:
Example of `allowScroll` for subscriptions prompt (note the scroll bar on the right):
Screen Shot 2022-08-17 at 5 07 12 PM

src/runtime/offers-flow.js
```diff
@@ -261,11 +261,11 @@ export class OffersFlow {
    * @param {!../model/client-config.ClientConfig} clientConfig
```
oyj9109, line 261: nit: update description for the new param.

src/runtime/contributions-flow.js
```diff
@@ -147,7 +147,7 @@ export class ContributionsFlow {
           return this.dialogManager_.openView(
             this.activityIframeView_,
             /* hidden */ false,
-            this.getDialogConfig_(clientConfig)
+            this.getDialogConfig_(clientConfig, this.clientConfigManager_)
```
oyj9109, line 150: How about calling `shouldAllowScroll` here and passing a boolean in instead?

src/runtime/offers-flow.js
justinchou-google: Missed this, good catch!

src/runtime/client-config-manager.js
justinchou-google: Will leave as `shouldAllowScroll`.

oyj9109 (approved):
LGTM. Thanks for leaving detailed notes and screencasts. 
I know there's a discussion on whether to apply this to other prompts, but I think we can follow up in a separate PR if needed.

jpettitt (approved):

PR #2379, opened 2022-08-17
Update swg-mini-prompt styling to match new ref-mocks.

rbeckthomas:
The current mini-prompt style does not match with the new current ref-mock. This change updates those styles to match the current design. It does not change the current copy.


jpettitt:
@rbeckthomas is this still applicable?

PR #2380, opened 2022-08-23
Logging 3P button click with new analytics event

ksebina:
Click event logging is currently absent now updating code to log it representing one of the flows (#1) for b/232215758.
This change deals handles redirect with a delay of very short/unperceivable time using setTimeout.

### DIFF
--- a/src/utils/gaa-test.js
+++ b/src/utils/gaa-test.js
@@ -1826,6 +1826,7 @@ describes.realWin('GaaGoogle3pSignInButton', {}, () => {
     sandbox.stub(self.console, 'warn');
 
     sandbox.stub(self, 'open');
+    // Makes sure no div elements on the test dom exist.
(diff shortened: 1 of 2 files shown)


src/utils/gaa-test.js
```diff
@@ -1798,6 +1826,10 @@ describes.realWin('GaaGoogle3pSignInButton', {}, () => {
     sandbox.stub(self.console, 'warn');
 
     sandbox.stub(self, 'open');
+    const elements = self.document.getElementsByTagName('div');
```
hartmannr76, line 1829: maybe leave a comment on this setup task? something like
```
// Makes sure no div elements on the test dom exist since this test creates some
```

ksebina: sg

hartmannr76 (approved):

PR #2381, opened 2022-08-24
Add ppid as a query param for swg-basic and swg-classic

ksebina:
SwgJS updated for Swg Zero and Classic publisher-supplied IDs. This change is for b/207177840


src/api/subscriptions.js
```diff
@@ -337,6 +337,13 @@ export class Subscriptions {
    * @return {?}
    */
   showBestAudienceAction() {}
+
+  /**
+   * Set the publisherProvidedId.
```
fikus, line 342: nit: "Sets". Method descriptions should be third person sentences written as if they start with "This method...".

src/runtime/entitlements-manager.js
```diff
@@ -853,6 +853,17 @@ export class EntitlementsManager {
         if (swgUserToken) {
           url = addQueryParam(url, 'sut', swgUserToken);
         }
+        // Add publisherProvidedId param for swg-basic
```
fikus, line 856: nit: End these comments with periods.

src/runtime/entitlements-manager.js
```diff
@@ -853,6 +853,17 @@ export class EntitlementsManager {
...
+        if (this.config_.publisherProvidedId) {
+          url = addQueryParam(url, 'ppid', this.config_.publisherProvidedId);
+        }
+        // Add publisherProvidedId param for swg-classic
+        else if (
+          params?.publisherProvidedId &&
+          typeof params.publisherProvidedId === 'string'
```
fikus, line 863: Why does the type need to be checked here?

fikus (approved):

src/api/subscriptions.js
```diff
@@ -337,6 +337,13 @@ export class Subscriptions {
...
+   * @param {string} publisherProvidedId
+   * @return {?}
+   */
+  setPpid(publisherProvidedId) {}
```
hartmannr76, line 346: Lets use the full name in our public interfaces `setPublisherProvidedId`

src/runtime/entitlements-manager.js
hartmannr76: +1 if we're going to do some type of assertion, let's do it above in the `getEntitlements` function and if we are going to ignore it we should give a `warn` on the page to let the developer know why

src/api/subscriptions.js
ksebina: Great, thanks for the example.

src/api/subscriptions.js
ksebina: okay will update

src/runtime/entitlements-manager.js
ksebina: The reason I was doing this check is because I was getting a JSC_TYPE_MISMATCH error when trying to using params.publisherProvidedId as a parameter in addQueryParam, saying it found : (string|undefined) and required: string. If this check is done in getEntitlements  to see if it is undefined I don't think a warn would get rid of the error? Would we want to throw an error?

hartmannr76: Spoke offline. We do a similar check below with `meteringStateId` which is the same thing as `publisherProvidedId` only masked in the metering configs. So this check is more for consistency with what we already do since its a param that the publisher can give to us directly

hartmannr76 (approved):

src/runtime/runtime.js
```diff
@@ -803,6 +813,14 @@ export class ConfiguredRuntime {
             error = 'Unknown skipAccountCreationScreen value: ' + value;
           }
           break;
+        case 'publisherProvidedId':
+          if (
+            value != undefined &&
+            !(typeof value === 'string' && value != '')
+          ) {
+            error = 'Unknown publisherProvidedId value: ' + value;
```
jpettitt, line 821: If it must be a string, can we make the error message say that? (more helpful for developers)

jpettitt (approved):

PR #2382, opened 2022-08-25
Add experiment which disables miniprompt button for wide desktop screen.

rbeckthomas:
This PR adds code which bypasses forces the display of a large prompt instead of a miniprompt, when the desktop width is greater than 480px.
This PR also creates logging, which notes for which publisher the changes have been made so that we can track the impact of this change.


fikus (changes requested):
As discussed, I don't think that `showPrompt_` is the most appropriate place to add the override logic. The logic of this method should be limited to displaying the prompt using the passed-in arguments.

Instead, you can create a utility method in this class to potentially override the given prompt type, and update callers to use that instead, eg.

```js
this.showPrompt_(getPromptTypeToDisplay(params.promptType), ...);
```

(note, I did not review the test changes in detail).

src/runtime/auto-prompt-manager.js
```diff
@@ -430,26 +439,74 @@ export class AutoPromptManager {
...
+        (autoPromptType === AutoPromptType.SUBSCRIPTION_LARGE ||
+          autoPromptType === AutoPromptType.CONTRIBUTION_LARGE) &&
+        displayLargePromptFn
+      ) {
+        displayLargePromptFn();
+      }
+  }
+
+  /**
+   * Returns which type of prompt to display based on the type specified
```
fikus, line 467: missing comma?

src/runtime/auto-prompt-manager.js
```diff
@@ -430,26 +439,74 @@ export class AutoPromptManager {
...
+   * the viewport width, and whether the disableDesktopMiniprompt experiment
+   * is enabled.
+   * If the disableDesktopMiniprompt experiment is enabled and
```
fikus, line 470: Stray sentence fragment.

src/runtime/auto-prompt-manager.js
```diff
@@ -430,26 +439,74 @@ export class AutoPromptManager {
...
+        this.logDisableMinipromptEvent_(promptType);
+        return AutoPromptType.CONTRIBUTION_LARGE;
+      }
+    } 
+
+    return promptType;
+  }
+
+  /** 
+   * Logs the disable miniprompt event
```
fikus, line 497: nit: End comment with period.

src/runtime/auto-prompt-manager-test.js
```diff
@@ -1076,6 +1083,41 @@ describes.realWin('AutoPromptManager', {}, (env) => {
...
+
+    await autoPromptManager.showAutoPrompt({
+      autoPromptType: AutoPromptType.CONTRIBUTION,
+      alwaysShow: true,
+      displayLargePromptFn: alternatePromptSpy,
+    });
+    expect(alternatePromptSpy).to.be.calledOnce;
+  });
+
+  it('should only display the mini prompt if the viewport is narrower than 480px and DISABLE_DESKTOP_MINIPROMPT is enabled', async () => {
```
fikus, line 1098: The condition "viewport narrower than 480px" should be set explicit in the test.

src/runtime/auto-prompt-manager-test.js
```diff
@@ -1076,6 +1083,41 @@ describes.realWin('AutoPromptManager', {}, (env) => {
     expect(alternatePromptSpy).to.not.be.called;
   });
 
+  it('should force display the large prompt if the viewport is wider than 480px and DISABLE_DESKTOP_MINIPROMPT is enabled', async () => {
```
fikus, line 1086: Event logging should also be covered by the tests.

src/runtime/auto-prompt-manager-test.js
```diff
@@ -1076,6 +1083,74 @@ describes.realWin('AutoPromptManager', {}, (env) => {
...
+  it('should log events when a large prompt overrides the miniprompt', async () => {
+    win./*OK*/ innerWidth = 500;
+    setExperiment(win, ExperimentFlags.DISABLE_DESKTOP_MINIPROMPT, true);
+
+    await autoPromptManager.showAutoPrompt({
+      autoPromptType: AutoPromptType.CONTRIBUTION,
+      alwaysShow: true,
+      displayLargePromptFn: alternatePromptSpy,
+    });
+    await eventManagerCallback({
```
fikus, line 1095: It isn't clear how eventManagerCallback is intended to verify the behavior. This test seems to pass even with the call to logEvent commented out in the implementation code.

src/runtime/auto-prompt-manager-test.js
```diff
@@ -1076,6 +1083,74 @@ describes.realWin('AutoPromptManager', {}, (env) => {
...
+      eventType: AnalyticsEvent.EVENT_DISABLE_MINIPROMPT_DESKTOP,
+      eventOriginator: EventOriginator.SWG_CLIENT,
+      isFromUserAction: false,
+      additionalParameters: null,
+    });
+    expect(alternatePromptSpy).to.be.calledOnce;
+  });
+
+  it('should replace the  contribution miniprompt with a large prompt if DISABLE_DESKTOP_MINIPROMPT is enabled and viewport is wider than 480px', async () => {
```
fikus, line 1104: nit: extra space before contribution

src/runtime/auto-prompt-manager-test.js
```diff
@@ -1076,6 +1083,74 @@ describes.realWin('AutoPromptManager', {}, (env) => {
...
+    clientConfigManagerMock
+      .expects('getClientConfig')
+      .returns(Promise.resolve(clientConfig))
+      .once();
+    // alwaysShow is false
+    miniPromptApiMock.expects('create').never();
+
+    await autoPromptManager.showAutoPrompt({
+      autoPromptType: AutoPromptType.CONTRIBUTION,
+      alwaysShow: false,
```
fikus, line 1148: You can consistently set alwaysShow: true for the tests instead of setting up the entitlements.

src/runtime/auto-prompt-manager.js
```diff
@@ -452,6 +467,54 @@ export class AutoPromptManager {
     }
   }
 
+  /**
+   * Returns which type of prompt to display based on the type specified,
+   * the viewport width, and whether the disableDesktopMiniprompt experiment
+   * is enabled.
+   *
+   * If the disableDesktopMiniprompt experiment is enabled and the desktop is
```
fikus, line 475: I think this is backwards? The large prompt will be substituted for the miniprompt.

src/runtime/auto-prompt-manager.js
```diff
@@ -452,6 +467,54 @@ export class AutoPromptManager {
...
+   * wider than 480px then the miniprompt type will be substituted for the large
+   * prompt. The miniprompt will be returned as is in all other cases.
```
fikus, line 477: The fallback behavior is not specific to miniprompts, it is true regardless of the type of input prompt type.

fikus (changes requested):

src/runtime/auto-prompt-manager-test.js
```diff
@@ -1076,6 +1085,77 @@ describes.realWin('AutoPromptManager', {}, (env) => {
...
+        promptType: AutoPromptType.CONTRIBUTION,
+      },
+    };
+
+    await autoPromptManager.showAutoPrompt({
+      autoPromptType: AutoPromptType.CONTRIBUTION,
+      alwaysShow: true,
+      displayLargePromptFn: alternatePromptSpy,
+    });
+    expect(logEventSpy).to.be.calledWith(expectedEvent);
```
fikus, line 1106: I think there is a calledOnceWith method you can use instead, and remove the next line.

fikus (approved):
Please make sure to squash the commits when you merge to avoid polluting the history.

src/runtime/auto-prompt-manager-test.js
rbeckthomas: Good call! I've combined those

src/runtime/auto-prompt-manager-test.js
rbeckthomas: Updated to use a spy instead.

PR #2383, opened 2022-08-31
Add read_time query param to entitlement fetch

mhkawano:
Use `read_time` from local storage as a query param for fetching entitlements. The value is updated whenever the user completes an action that can affect their entitlements, and is used to decide if we should use a strong read on the backend.


src/runtime/audience-action-flow.js
```diff
@@ -163,6 +164,8 @@ export class AudienceActionFlow {
     } else {
       this.showFailedOptedInToast_();
     }
+    const now = Date.now().toString();
```
hartmannr76, line 167: So the only granularity we really care about is seconds. I feel like maybe we just store and use that value and just make sure our documentation indicates that

src/runtime/audience-action-flow.js
```diff
@@ -163,6 +164,8 @@ export class AudienceActionFlow {
...
+    this.deps_.storage().set(Constants.READ_TIME, now, true);
```
hartmannr76, line 168: I think we should actually be fine to put this in session storage. The max staleness that we are going to start with will just be 10s so its ok for this property to be ephemeral

mhkawano: Done.

src/runtime/audience-action-flow-test.js
```diff
@@ -27,8 +27,10 @@ import {Constants} from '../utils/constants';
 import {PageConfig} from '../model/page-config';
 import {ProductType} from '../api/subscriptions';
 import {Toast} from '../ui/toast';
+import {nowAsSecondsString} from '../utils/date-utils';
 
 const WINDOW_LOCATION_DOMAIN = 'https://www.test.com';
+const CURRENT_TIME = 1615416442; // GMT: Wednesday, March 10, 2021 10:47:22 PM
```
hartmannr76, line 33: Weird.. documentation suggests this should be `1615416442000` for this specific date

src/runtime/audience-action-flow-test.js
```diff
@@ -132,6 +135,10 @@ describes.realWin('AudienceActionFlow', {}, (env) => {
       .expects('set')
       .withExactArgs(Constants.USER_TOKEN, 'fake user token', true)
       .exactly(1);
+    storageMock
+      .expects('set')
+      .withExactArgs(Constants.READ_TIME, nowAsSecondsString(), false)
```
hartmannr76, line 140: Instead of calling the function directly, lets compare against a constant string value

src/runtime/audience-action-flow-test.js
mhkawano: Yeah weird. I copied it from another place in the code, but the date doesn't really matter so I'll just remove it.

src/runtime/audience-action-flow-test.js
mhkawano: Done.

src/runtime/audience-action-flow.js
```diff
@@ -145,6 +146,7 @@ export class AudienceActionFlow {
    * On a successful response from the dialog, we should:
    * 1) Store the updated user token
    * 2) Clear existing entitlements from the page
+   * 3) Update READ_TIME in local storage to indicate that entitlements may have changed recently
    * 3) Re-fetch entitlements which may potentially provide access to the page
```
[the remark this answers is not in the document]
mhkawano: Done.

jpettitt:
What are you comparing this to on the back end?  There is no guarantee the users clock is even close to accurate (phones tend to be ok, desktops not so much)

jpettitt:
Just had chat with @hartmannr76 - a better design here would be to set a "strongRead" flag if the locally stored time is less than some threshold old (30 seconds?). That way we're only ever comparing the same time source so the absolute value doesn't matter.

hartmannr76:
Makes sense. Caught up with @mhkawano and we'll go that direction

mhkawano:
Updated to sending the time since the action to determine staleness.

src/runtime/audience-action-flow.js
```diff
@@ -163,6 +164,8 @@ export class AudienceActionFlow {
     } else {
       this.showFailedOptedInToast_();
     }
+    const now = Date.now().toString();
+    this.deps_.storage().set(Constants.READ_TIME, now, false);
```
hartmannr76, line 168: `useLocalStorage` defaults to false so you shouldn't need to provide the explicit value. If you want to, maybe include a comment to indicate what the param name is
```js
this.deps_.storage().set(Constants.READ_TIME, now, /*useLocalStorage*/ false);
```

src/runtime/entitlements-manager.js
```diff
@@ -827,16 +827,21 @@ export class EntitlementsManager {
     // Get swgUserToken from local storage
     const swgUserTokenPromise = this.storage_.get(Constants.USER_TOKEN, true);
 
+    // Get read_time from local storage
```
hartmannr76, line 830: `Get read_time from session storage`

src/runtime/entitlements-manager.js
```diff
@@ -854,6 +859,18 @@ export class EntitlementsManager {
           url = addQueryParam(url, 'sut', swgUserToken);
         }
 
+        // Add interaction_age param.
+        if (readTime) {
+          const last = parseInt(readTime, 10);
+          if (last) {
+            const now = Date.now();
+            const interaction_age = Math.floor((now - last) / 1000);
```
hartmannr76, line 867: I would just do the `Date.now()` directly inside. No need for the extra variable since it's only used on this line and doesn't provide any additional context

src/runtime/audience-action-flow.js
mhkawano: Done. Opted to keep the param with a comment because I remember being confused by this in the past.

src/runtime/entitlements-manager.js
mhkawano: Done.

src/runtime/entitlements-manager.js
mhkawano: Done.

hartmannr76 (approved):

alin04 (approved):

PR #2387, opened 2022-09-01
Opt in failure translations

mhkawano:
Translations for the opt in failure messages. The message ids are `6823823019943927988` and `4525404757272971742`. 

Manually updated, don't know if there is a better way.  Waiting on Portuguese(pt) to be translated. Will double check once I have all of them.


mhkawano:
pt was already translated, don't know why I didn't see it before.

fikus (approved):

PR #2394, opened 2022-09-14
fix: having the prefix included for empty client default to true

hartmannr76:
The LinkComplete class was setup to load client configuration but never fetched it.

Due to server changes causing problems (reference b/246641990), this defaults the empty configuration to use the prefix and sets the default of the `feUrl` function to use the prefix


alin04 (approved):

jpettitt:
Thanks for the quick fix

PR #2410, opened 2022-10-05
Remove check for secure channel in entitlements check

fikus:
Remove check for secure channel to fix entitlements check on Firefox. When the activity runs as a redirect, it is not considered secure.

b/190614158

### DIFF
--- a/src/runtime/basic-runtime.js
+++ b/src/runtime/basic-runtime.js
@@ -506,7 +506,7 @@ export class ConfiguredBasicRuntime {
       port,
       feOrigin(),
       /* requireOriginVerified */ true,
-      /* requireSecureChannel */ true
+      /* requireSecureChannel */ false
     );
     return promise.then((response) => {
       const jwt = response['jwt'];
--- a/src/runtime/basic-runtime-test.js
+++ b/src/runtime/basic-runtime-test.js
@@ -454,7 +454,7 @@ describes.realWin('BasicRuntime', {}, (env) => {
         port,
         sandbox.match.any,
         true,
-        true
+        false
       );
       expect(data['jwt']).to.equal('abc');
       expect(data['usertoken']).to.equal('xyz');


src/runtime/basic-runtime.js
```diff
@@ -506,7 +506,7 @@ export class ConfiguredBasicRuntime {
       port,
       feOrigin(),
       /* requireOriginVerified */ true,
-      /* requireSecureChannel */ true
+      /* requireSecureChannel */ false
```
hartmannr76, line 509: Is there any reason the incoming "port" can't be an `ActivityPort`? It looks like `ActivityPort` extends it so it _should_ be fine. The `getMode` [function](https://github.com/subscriptions-project/swg-js/blob/39c86fa5efd13fe44bf489eaef330e4fe458eb8d/src/components/activities.js#L48) from that can tell us how the result was provided (redirect, popup, iframe) and I feel like if the goal is to only have this case in redirect mode, we can limit our requirement to those types of responses. Thoughts?

fikus: Do you mean adding logic like `requireSecureChannel = port.getMode() !== 'redirect'`?

If so, that would be redundant, because `isSecureChannel()` is hard-coded false for redirect and hard-coded true for popup and iframe.

hartmannr76: Isn't this value used for [this check](https://github.com/fikus/swg-js/blob/7eec4e039fc77b6cc96b2704f96680f8bbbb7b5b/src/utils/activity-utils.js#L34)?

In my head, we'd only want to require a secure channel if we know it came from a certain response. Where `isSecureChannel` would be false (in the case of redirect) right now we are expecting it to be `true` (thats what this change is meant to address, right?) The idea would be only expect it to be `false` in certain instances (redirect) instead of all instances

hartmannr76 (approved):

PR #2415, opened 2022-10-11
Providing subscriptionTimestamp to Garamond EA

ryan-minyu:


src/api/entitlements.js
```diff
@@ -294,8 +294,15 @@ export class Entitlement {
    * @param {!Array<string>} products
    * @param {string} subscriptionToken
    * @param {JsonObject|null|undefined} subscriptionTokenContents
+   * @param {Timestamp} subscriptionTimestamp
    */
-  constructor(source, products, subscriptionToken, subscriptionTokenContents) {
+  constructor(
```
jpettitt, line 299: Adding this to the entitlement object feels wrong - can we do a VC to discuss

jpettitt (changes requested):
need to understand what you are trying to to here

src/api/entitlements.js
ryan-minyu: Sure, this PR is still in progress. But we can discuss about that.

src/api/entitlements.js
```diff
@@ -294,8 +295,15 @@ export class Entitlement {
    * @param {!Array<string>} products
    * @param {string} subscriptionToken
    * @param {JsonObject|null|undefined} subscriptionTokenContents
+   * @param {!Timestamp} subscriptionTimestamp
```
jpettitt, line 298: Must be nullable for things like metering entitlements where it makes no sense

src/api/entitlements.js
```diff
@@ -373,18 +384,29 @@ export class Entitlement {
...
     } catch (e) {
       subscriptionTokenContents = null;
     }
+    try {
+      subscriptionTimestamp = new Timestamp(
+        [timestampJson.seconds_, timestampJson.nanos_],
+        false
+      );
+    } catch (e) {
+      subscriptionTimestamp = new Timestamp();
```
jpettitt, line 402: Timestamp constructor has non nullable params

jpettitt (changes requested):

ryan-minyu:
The auto-generated file api_messages-test.js will have error after my change. Since there will be multiple Timestamp data type fields in a single message. The protobuf lite will declare `const timestamp = new Timestamp()` twice which causes the error. Any suggestions on this? Thanks!

src/api/entitlements.js
ryan-minyu: Fixed, thanks!

src/utils/date-utils-test.js
```diff
@@ -63,3 +67,26 @@ describe('convertPotentialTimestampToSeconds', () => {
     );
   });
 });
+
+describe('convertPotentialTimestampToMilliseconds', () => {
+  it('converts seconds to milliseconds ', () => {
+    const nowInSeconds = Math.floor(Date.now() * 1000);
+    const nowInMilliseconds = Date.now();
```
ChrisAntaki, line 74: Could we just call `Date.now()` once in this test? Right now I think there's a small chance the test would fail once in a while

src/utils/date-utils-test.js
```diff
@@ -63,3 +67,26 @@ describe('convertPotentialTimestampToSeconds', () => {
     );
   });
 });
+
+describe('convertPotentialTimestampToMilliseconds', () => {
```
fikus, line 71: Please also update these tests so they are not reimplementing the logic being verified.

ryan-minyu: Sorry, which logic are they reimplementing? I thought they are not related to the bug I fixed?

fikus: Line 73 here is just copied from line 63 of the implementation, for example.

ChrisAntaki: Would you these tests to use literal values?

fikus: Yes, that would be the straightforward way to remove the logic.

src/api/entitlements.js
```diff
@@ -373,18 +384,29 @@ export class Entitlement {
     const products = json['products'] || [];
     const subscriptionToken = json['subscriptionToken'];
     let subscriptionTokenContents;
+    const timestampJson = json['subscriptionTimestamp'];
+    let subscriptionTimestamp;
```
ChrisAntaki, line 388: Could you move these variable declarations down so they're just above the new try/catch block? I think that might make the code a little simpler, since the first try/catch block doesn't need these new variables

src/runtime/entitlements-manager.js
```diff
@@ -366,17 +369,22 @@ export class EntitlementsManager {
...
+    optionalIsUserRegistered = null,
+    optionalSubscriptionTimestamp = null
   ) {
     const message = new EntitlementsRequest();
     message.setUsedEntitlement(usedEntitlement);
     message.setClientEventTime(toTimestamp(Date.now()));
     message.setEntitlementResult(entitlementResult);
     message.setEntitlementSource(entitlementSource);
     message.setToken(optionalToken);
+    message.setSubscriptionTimestamp(optionalSubscriptionTimestamp);
```
ChrisAntaki, line 381: Seems like this line (381) and lines 385-387 are doing the same thing a bit differently?

ChrisAntaki: Is the `if` statement necessary? Just curious, because I saw one on line 745 as well

src/utils/date-utils.js
```diff
@@ -45,3 +45,22 @@ export function convertPotentialTimestampToSeconds(timestamp) {
   }
   return timestampInSeconds;
 }
+
+/**
```
ChrisAntaki, line 49: Could you add a description that mentions why this method accepts three kinds of timestamps? (seconds, milliseconds, microseconds)

src/utils/gaa.js
```diff
@@ -2260,4 +2263,12 @@ export class GaaMetering {
       }
     });
   }
+
+  static getSubscriptionTimestamp() {
+    if (GaaMetering.userState && GaaMetering.userState.subscriptionTimestamp) {
```
ChrisAntaki, line 2268: Could this method body be simplified to `return GaaMetering?.userState?.subscriptionTimestamp` ?

src/runtime/entitlements-manager.js
ryan-minyu: > Is the `if` statement necessary? Just curious, because I saw one on line 745 as well

I was trying not to break current path. I wasn't sure about an `undefined` or something else may break it.

src/utils/gaa.js
ryan-minyu: Sounds good, thanks!

src/runtime/entitlements-manager.js
ChrisAntaki: In that case, would it make sense to remove line 381?

ryan-minyu: Sure, actually I'm not sure why I added line 381. It should be removed. Thanks!

src/runtime/runtime.js
```diff
@@ -1168,6 +1172,15 @@ export class ConfiguredRuntime {
       showcaseEventToAnalyticsEvents(entitlement.entitlement) || [];
     const params = new EventParams();
     params.setIsUserRegistered(entitlement.isUserRegistered);
+    if (entitlement.subscriptionTimestamp) {
+      params.setSubscriptionTimestamp(
```
ChrisAntaki, line 1176: Would it be possible to add test coverage for this line?

ryan-minyu: New test can cover this line, thanks!

ChrisAntaki: Sweet!

ChrisAntaki (approved):

jpettitt (approved):

PR #2420, opened 2022-10-18
add logging to GA on Survey Submission

justinchou-google:
b/242735734
implement logging to Google Analytics for survey submission


justinchou-google:
hi @oyj9109, do you know how I can resolve this type error? ```src/runtime/audience-action-flow.js:284:34: ERROR - [JSC_INEXISTENT_PROPERTY] Property gtag never defined on Window
  [28](https://github.com/subscriptions-project/swg-js/actions/runs/3277456901/jobs/5394759701#step:5:29)4|     const gtag = this.deps_.win().gtag;```

I don't see a type defined for it in:
https://github.com/subscriptions-project/swg-js/blob/main/src/runtime/google-analytics-event-listener.js#L68

oyj9109:
> hi @oyj9109, do you know how I can resolve this type error? `src/runtime/audience-action-flow.js:284:34: ERROR - [JSC_INEXISTENT_PROPERTY] Property gtag never defined on Window [28](https://github.com/subscriptions-project/swg-js/actions/runs/3277456901/jobs/5394759701#step:5:29)4| const gtag = this.deps_.win().gtag;`
> 
> I don't see a type defined for it in: https://github.com/subscriptions-project/swg-js/blob/main/src/runtime/google-analytics-event-listener.js#L68

I guess existence of the property should be checked before assigning it to a variable. What about just defining a local variable for window object? I think `win.gtag` should be okay.

justinchou-google:
> > hi @oyj9109, do you know how I can resolve this type error? `src/runtime/audience-action-flow.js:284:34: ERROR - [JSC_INEXISTENT_PROPERTY] Property gtag never defined on Window [28](https://github.com/subscriptions-project/swg-js/actions/runs/3277456901/jobs/5394759701#step:5:29)4| const gtag = this.deps_.win().gtag;`
> > I don't see a type defined for it in: https://github.com/subscriptions-project/swg-js/blob/main/src/runtime/google-analytics-event-listener.js#L68
> 
> I guess existence of the property should be checked before assigning it to a variable. What about just defining a local variable for window object? I think `win.gtag` should be okay.

This worked, thanks!

src/runtime/audience-action-flow.js
```diff
@@ -266,12 +266,38 @@ export class AudienceActionFlow {
...
+   * @private
+   */
+  logSurveyDataToGoogleAnalytics(request) {
+    const gtag = this.deps_.win().gtag || null;
+    if (gtag && typeof gtag !== 'function') {
+      return false;
+    }
+    request.getSurveyQuestionsList().map((question) => {
+      const answer = question.getSurveyAnswersList()[0];
+      gtag('event', 'survey submission', {
```
justinchou-google, line 290: Do we want to put this event message into https://github.com/subscriptions-project/swg-js/blob/756d962e748475a9f6f5f95c7fe376709ed613a1/src/runtime/event-type-mapping.js#L126-L218?

oyj9109: No, I think your approach is right that the handler directly invokes `gtag` function if available, rather than delegating it to event manager, so that we have control over success/failure of the data transfer.

src/runtime/audience-action-flow-test.js
```diff
@@ -64,14 +104,20 @@ describes.realWin('AudienceActionFlow', {}, (env) => {
...
-    });
+    sandbox.stub(runtime, 'win').returns(win);
     onCancelSpy = sandbox.spy();
     sandbox.useFakeTimers(CURRENT_TIME);
   });
 
+  function setGtagToNonfunction() {
+    runtime.win.restore();
+    sandbox.stub(runtime, 'win').returns(
+      Object.assign({}, win, {
```
oyj9109, line 115: What about removing the property which is the expected condition when gtag is not available?
`delete win.gtag`

src/runtime/audience-action-flow-test.js
```diff
@@ -34,6 +36,35 @@ const WINDOW_LOCATION_DOMAIN = 'https://www.test.com';
...
+const TEST_CATEGORY_2 = 'Test Category 2';
+const TEST_QUESTION_TEXT_2 = 'Test Question 2';
+const TEST_ANSWER_TEXT_1 = 'Test Answer 1';
+const TEST_ANSWER_TEXT_2 = 'Test Answer 2';
+
+const TEST_SURVEYANSWER_1 = new SurveyAnswer();
+TEST_SURVEYANSWER_1.setAnswerCategory(TEST_CATEGORY_1);
+TEST_SURVEYANSWER_1.setAnswerText(TEST_ANSWER_TEXT_1);
+const TEST_SURVEYQUESTION_1 = new SurveyQuestion();
+TEST_SURVEYQUESTION_1.setQuestionCategory(TEST_CATEGORY_1);
```
oyj9109, line 50: AnswerCategory and QuestionCategory will be different values.

src/runtime/audience-action-flow-test.js
```diff
@@ -445,10 +508,38 @@ describes.realWin('AudienceActionFlow', {}, (env) => {
...
-    messageCallback(surveyDataTransferRequest);
+    const messageCallback = messageMap[TEST_SURVEYDATATRANSFERREQUEST.label()];
+    messageCallback(TEST_SURVEYDATATRANSFERREQUEST);
+
+    winMock.verify();
+    activityIframeViewMock.verify();
+  });
+
+  it(`handles a SurveyDataTransferRequest with failed logging`, async () => {
+    setGtagToNonfunction();
```
oyj9109, line 519: (Optional) consider having the implementation within the test if the function is going to be used only for this.

src/runtime/audience-action-flow.js
```diff
@@ -266,12 +266,38 @@ export class AudienceActionFlow {
...
+    if (typeof gtag !== 'function') {
+      return false;
+    }
+    request.getSurveyQuestionsList().map((question) => {
+      const answer = question.getSurveyAnswersList()[0];
+      gtag('event', 'survey submission', {
+        'event_category': 'survey',
+        'survey_question_category': question.getQuestionCategory(),
+        'survey_question': question.getQuestionText(),
+        'survey_answer_category': answer.getAnswerCategory() || null,
```
oyj9109, line 294: Is ` || null` necessary here? I think answer category is a required field that should be always present in the request?
Same for answer text.

oyj9109 (approved):
LGTM. A couple of comments for nit.

src/runtime/audience-action-flow-test.js
justinchou-google: I created a function for this firstly for clarity, so that the env setup is separate from the test and with the other setup as well, and secondly so this can be reused / refactored for other types of failure tests.

src/runtime/audience-action-flow.js
```diff
@@ -266,12 +266,38 @@ export class AudienceActionFlow {
...
   }
 
+  /**
+   * Logs SurveyDataTransferRequest to Google Analytics. Returns boolean
+   * for whether or not logging was successful.
+   * @param {SurveyDataTransferRequest} request
+   * @return {boolean}
+   * @private
+   */
+  logSurveyDataToGoogleAnalytics(request) {
```
hartmannr76, line 283: @oyj9109 do we not want to use the event messaging system to send the data instead of sending to GA directly from here? Trying to think about future use-cases where they use their own endpoint instead of GA to capture survey information

oyj9109: Having a direct access to the GA function here is preferred as it's guaranteed the function is available and actually invoked. If we delegate it to event manager, it's hard to handle it via async flow.
Even though some publishers implement their own endpoint and provides callbacks in the future, I think we'll still support GA-based data collection so this will be still beneficial.

hartmannr76: Yeah, we have an internal bug around being able to ensure something captured the event from the event manager, but I bring it up because the current logic **requires** GA to be set up. We should make sure that:
1) Publishers know they need to have it
2) We may not even want to show the survey on the page if they dont. I'd rather push a warning to the console indicating it is required vs failing this late in the flow

oyj9109: Definitely! Great call outs.
1. In experiment, we'll clearly communicate this through. They should know it because it's the only way they can collect data. In the future, we'll need to make it clear in the publisher configuration flow. (either callback or GA is required)
2. Seems like Justin already started working on it. See https://github.com/subscriptions-project/swg-js/pull/2421

jpettitt: +1 on refactor

src/runtime/google-analytics-event-listener.js
```diff
@@ -65,16 +75,61 @@ export class GoogleAnalyticsEventListener {
...
+            'non_interaction': gaEvent.nonInteraction,
+          },
+          !!analyticsParams?.eventCategory
+            ? {
+                'event_category': analyticsParams.eventCategory,
+              }
+            : null,
+          !!analyticsParams?.surveyQuestion
+            ? {
+                'survey_question': analyticsParams.surveyQuestion || '',
```
justinchou-google, line 117: Needed to add another null check here to satisfy the compiler, but this should never be `null` nor `undefined` after the check on 115, same for `surveyAnswerCategory`.

src/runtime/audience-action-flow.js
justinchou-google: Please see new changes!

src/api/client-event-manager-api.js
```diff
@@ -53,7 +53,7 @@ export class ClientEventManagerApi {
   /**
    * Call this function to log an event. The registered listeners will be
    * invoked unless the event is filtered.
-   * @param {!function(!ClientEvent)} listener
+   * @param {!function(!ClientEvent, Object)} listener
```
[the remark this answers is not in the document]
oyj9109: Yeah seems like adding a new argument for additional params makes the most sense to me.

A minor comment for nit:
Can we define a type for this param to handle type checking better? Something like
```
 *  @typedef {{
 *    googleAnalyticsParameters: ?Object,
 * }}
 */
export let ClientEventParams;
```
and I think we can even define a type for the `googleAnalyticsParameters`

src/runtime/client-event-manager.js
```diff
@@ -86,7 +86,7 @@ export class ClientEventManager {
    * @param {!Promise} configuredPromise
    */
   constructor(configuredPromise) {
-    /** @private {!Array<function(!../api/client-event-manager-api.ClientEvent)>} */
+    /** @private {!Array<function(!../api/client-event-manager-api.ClientEvent, Object)>} */
```
oyj9109, line 89: And also listeners gets optional params here so that we can directly pass `eventParams` in the listeners in ` logEvent` below?

src/runtime/google-analytics-event-listener.js
```diff
@@ -65,16 +75,61 @@ export class GoogleAnalyticsEventListener {
...
 
     // TODO(b/234825847): Remove it once universal analytics is deprecated in 2023.
-    if (typeof this.win_.ga === 'function') {
-      this.win_.ga('send', 'event', gaEvent);
+    const ga = this.win_.ga || null;
+    if (isFunction(ga)) {
+      ga(
+        'send',
+        'event',
+        Object.assign(
```
oyj9109, line 83: Can we overwrite gaEvent with the analyticsParams and use it across `ga` and `gtag`?
```
// Overwite ...
gaEvent = {
  ... gaEvent,
  ... analyticsParams
}
```

src/runtime/google-analytics-event-listener.js
```diff
@@ -65,16 +75,61 @@ export class GoogleAnalyticsEventListener {
...
+            : null,
+          !!analyticsParams?.eventLabel
+            ? {eventLabel: analyticsParams.eventLabel}
+            : null
+        )
+      );
     }
 
-    if (typeof this.win_.gtag === 'function') {
-      this.win_.gtag('event', gaEvent.eventAction, {
```
oyj9109, line 73: Then you can split `gaEvent.eventAction` and the rest.

```
const {eventAction, ...gtagEventParam} = gaEvent;
this.win_.gtag('event', eventAction, gtagEventParam);
```

src/api/client-event-manager-api.js
```diff
@@ -72,7 +72,8 @@ export class ClientEventManagerApi {
    * stop the event if a filterer cancels it.  After that, it will call each
    * listener asynchronously.
    * @param {!ClientEvent} event
+   * @param {Object} eventParams
```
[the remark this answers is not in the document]
justinchou-google: I believe if we make it optional, it's allowed to be null/undefined, but the compiler will still complain if the number of arguments don't match. I can still change it to be optional, but I think it is more clear to the callers if it is passing an empty object rather than null..

src/runtime/google-analytics-event-listener.js
```diff
@@ -37,11 +38,20 @@ export class GoogleAnalyticsEventListener {
     );
   }
 
+  /*    * @param {?{
```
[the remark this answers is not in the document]
justinchou-google: forgot to delete this, thanks!

src/runtime/google-analytics-event-listener.js
justinchou-google: The event is slightly different for `ga` and `gtag`, right? `ga` requires `eventCategory` and `eventLabel`, whereas `gtag` requires `event_category`, `event_label`, and the other survey attributes. Is it safe to pass additional attributes to each respective function, i.e. `event_category` to `ga`?

src/runtime/google-analytics-event-listener.js
justinchou-google: Sorry, can you be more explicit here? How does `gtagEventParam` have the required attributes defined? Another step we have to do is convert the attributes from `snake_case` to `camelCase`.

src/runtime/google-analytics-event-listener.js
justinchou-google: new changes make this comment obsolete.

src/api/client-event-manager-api.js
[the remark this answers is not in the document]
oyj9109: Optional and nullable are different. I think the right syntax here is `@param {!Object=} eventParams` to make it optional (maybe undefined but if present, must not be null).
https://github.com/google/closure-compiler/wiki/Types-in-the-Closure-Type-System#nullable-vs-optional-parameters-and-properties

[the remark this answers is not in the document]
justinchou-google: I see, allowing undefined makes the function overloading work!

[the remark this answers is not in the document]
justinchou-google: This still doesn't get around the issue for me for when we call `logEvent` (above comment) or when we register listeners / bind the callback function:
```
src/runtime/client-event-manager.js:109:25: ERROR - [JSC_TYPE_MISMATCH] actual parameter 1 of Array.prototype.push does not match formal parameter
found 
```
```
src/runtime/google-analytics-event-listener.js:37:6: ERROR - [JSC_TYPE_MISMATCH] actual parameter 1 of ClientEventManager$$module$src$runtime$client_event_manager.prototype.registerEventListener does not match formal parameter
found 
```

src/runtime/google-analytics-event-listener.js
oyj9109: Oh, I didn't notice they are based on different cases styles. In that case, I think it would make more sense for the callers to set the attributes in `analyticsParams` in snake_case when it passes it in (as `ga` will be deprecated next year).

Then we can do some manual overwites, e.g., 
```
gaEvent = {
   ...gaEvent,
   eventCategory: analyticsParams.event_category || gaEvent.eventCategory,
   eventLabel: analyticsParams.event_label || gaEvent. eventLabel,
}
```

For `gtag`, we can do the manual conversion (as it was doing) and then overwrite.

src/api/client-event-manager-api.js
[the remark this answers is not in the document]
justinchou-google: seems like I was able to resolve this with some additional `|undefined` annotations, marking resolved.

src/runtime/google-analytics-event-listener.js
```diff
@@ -40,8 +41,9 @@ export class GoogleAnalyticsEventListener {
   /**
    *  Listens for new events from the events manager and logs appropriate events to Google Analytics.
    * @param {!../api/client-event-manager-api.ClientEvent} event
+   * @param {(!../api/client-event-manager-api.ClientEventParams|undefined)=} analyticsParams
    */
-  handleClientEvent_(event) {
+  handleClientEvent_(event, analyticsParams = undefined) {
```
oyj9109, line 46: Should this be `eventParams` and
`const analyticsParams = eventParams.googleAnalyticsParameters;` ?

src/runtime/audience-action-flow.js
```diff
@@ -267,12 +270,45 @@ export class AudienceActionFlow {
...
+    }
+    request.getSurveyQuestionsList().map((question) => {
+      const answer = question.getSurveyAnswersList()[0];
+      const event = {
+        eventType: AnalyticsEvent.ACTION_SURVEY_DATA_TRANSFER,
+        eventOriginator: EventOriginator.SWG_CLIENT,
+        isFromUserAction: true,
+        additionalParameters: null,
+      };
+      const analyticsParams = {
```
oyj9109, line 301: Should this be put in the `googleAnalyticsParameters` property?

src/runtime/basic-runtime.js
```diff
@@ -233,6 +233,7 @@ export class BasicRuntime {
     });
     this.setOnLoginRequest();
     this.processEntitlements();
+    this.processEntitlements();
```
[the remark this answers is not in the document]
justinchou-google: thanks!

src/runtime/google-analytics-event-listener.js
justinchou-google: good catch!

src/runtime/audience-action-flow.js
justinchou-google: yes, good catch!

src/runtime/google-analytics-event-listener.js
justinchou-google: how do these changes look?

oyj9109: Looks great!

src/runtime/google-analytics-event-listener-test.js
```diff
@@ -169,6 +169,51 @@ describes.realWin('GoogleAnalyticsEventListener', {}, (env) => {
...
+        ga: () => {},
+        gtag: () => {},
+      }),
+      true
+    );
+    const gaEvent = analyticsEventToGoogleAnalyticsEvent(
+      AnalyticsEvent.IMPRESSION_OFFERS
+    );
+    const gaEventWithParams = Object.assign({}, gaEvent, {
+      eventCategory: 'TEST CATEGORY',
```
oyj9109, line 184: Can we set an empty value (or a different value) for `eventCategory` to make sure it's overwritten by the given params?

oyj9109 (approved):
LGTM. Thanks for the refactoring!

src/runtime/google-analytics-event-listener-test.js
oyj9109: Never mind. It's actually overwriting what's returned by `analyticsEventToGoogleAnalyticsEvent`. I misunderstood it originally.

jpettitt (approved):

PR #2421, opened 2022-10-19
Add AudienceAction filter for survey action eligibility

justinchou-google:
b/254455836
Filter out action_type `TYPE_REWARDED_SURVEY` if gtag is not registered.


justinchou-google:
hi @oyj9109 , before I work on the tests, can you confirm this is essentially what we want for this ticket?

src/utils/action-utils.js
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Returns boolean for whether publisher is eligible for gTag
+ * @param {!../runtime/deps.DepsDef} deps
+ * @return {boolean}
+ */
+export function isGtagEligible(deps) {
```
justinchou-google, line 22: I will use this to refactor the changes in https://github.com/subscriptions-project/swg-js/pull/2420 once that is merged.

oyj9109 (approved):

src/utils/action-utils.js
hartmannr76: I'm not thrilled about this location. What I'm wondering is if we can just add it as a static function on the `GoogleAnalyticsEventListener` class. I feel like the check should just use something like `(self.gtag || null) == 'function` instead of needing the deps for the window object. @ChrisAntaki do you have a preference

ChrisAntaki: I like the idea of defining a static method on the `GoogleAnalyticsEventListener` class, especially if it accepts a `deps` param. We might be able to simplify the conditional logic to `typeof deps.win().gtag === 'function'`

justinchou-google: Agree, I can create a static method, something else I wanted to do was to refactor this utils in favor of https://github.com/subscriptions-project/swg-js/blob/main/src/utils/types.js#L49

justinchou-google: Made the change, but one thing I want to double check, can we guarantee that `self.gtag` === `deps.win().gtag` (and similarly for `ga`)?

ChrisAntaki: They might differ, but I think that's fine. Sometimes AMP will pass in a `win` object that differences from the current `window` object, right?

justinchou-google: Futhermore, the changes in https://github.com/subscriptions-project/swg-js/pull/2420 are executing `gtag` from `win`, whereas this check is checking `self`; does this open up the possibility for bad GA logging?

ChrisAntaki: Pretty sure we should consistently check the "win" object instead of referencing the current global scope with "self"

ChrisAntaki: Could the new static methods both require a `deps` parameter?

ChrisAntaki: I think checking `self` (global scope) might lead to issues with AMP

oyj9109: +1 to passing `deps` (or `win`) object from callers to avoid the possibility for mis logging.

src/runtime/google-analytics-event-listener.js
```diff
@@ -77,4 +78,20 @@ export class GoogleAnalyticsEventListener {
       });
     }
   }
+
+  /**
+   * Function to determine whether publisher is eligible for GA.
+   * @returns {boolean}
+   */
+  static isGaEligible() {
```
hartmannr76, line 86: We want to pass `deps` into this, dont we?

justinchou-google: Yep, just updated. Sorry for the delay.

src/utils/action-utils.js
justinchou-google: Updated function to accept `deps` parameter.

hartmannr76 (approved):

PR #2428, opened 2022-10-26
Suggest "w" flag instead of "watch", since Nodejs 18+ intercepts the "watch" flag now

ChrisAntaki:
Nodejs 18 added its own "--watch" param. This is a pretty cool upgrade. I noticed Nodejs is intercepting that "--watch" param now, which was breaking ours. I renamed ours to "--persist", which seems to work well enough
https://nodejs.org/en/blog/release/v18.11.0/#watch-mode-experimental


ChrisAntaki:
If this looks good, I'll follow it up with some updates to internal docs

fikus:
Since `-w` is already supported, is there still a need for this change? I think that npx always intercepting --watch is a bug so we might as well leave this in, in case it is fixed upstream. If the references to --watch might be confusing, maybe we should just add a comment?

ChrisAntaki:
I added back the checks for `argv.watch`. Now this PR just updates the task.flags objects, so `npx gulp --tasks` suggests the `w` flag instead of `watch`
image

fikus (approved):

PR #2435, opened 2022-10-26
date-utils bug fix

ryan-minyu:
Fix a bug when converts microseconds to seconds.

### DIFF
--- a/src/utils/date-utils-test.js
+++ b/src/utils/date-utils-test.js
@@ -57,7 +57,7 @@ describe('convertPotentialTimestampToSeconds', () => {
   });
   it('converts microseconds to seconds', () => {
     const nowInMicroseconds = Date.now() * 1000;
-    const nowInSeconds = Math.floor(nowInMicroseconds / 1000000);
+    const nowInSeconds = Math.floor(nowInMicroseconds / 1e6);
     expect(convertPotentialTimestampToSeconds(nowInMicroseconds)).to.equal(
       nowInSeconds
     );
--- a/src/utils/date-utils-test.js
+++ b/src/utils/date-utils-test.js
@@ -57,7 +57,7 @@ describe('convertPotentialTimestampToSeconds', () => {
   });
   it('converts microseconds to seconds', () => {
     const nowInMicroseconds = Date.now() * 1000;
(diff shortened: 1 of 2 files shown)


src/utils/date-utils.js
```diff
@@ -35,7 +35,7 @@ export function convertPotentialTimestampToSeconds(timestamp) {
   let timestampInSeconds;
   if (timestamp >= 1e14 || timestamp <= -1e14) {
     // Microseconds
-    timestampInSeconds = Math.floor(timestamp / 100000);
+    timestampInSeconds = Math.floor(timestamp / 1000000);
```
fikus, line 38: Consider writing the literal as 1e6 or 1_000_000 for clarity (either style would have made the bug evident).

fikus (approved):

src/utils/date-utils-test.js
```diff
@@ -57,7 +57,7 @@ describe('convertPotentialTimestampToSeconds', () => {
   });
   it('converts microseconds to seconds', () => {
     const nowInMicroseconds = Date.now() * 1000;
-    const nowInSeconds = Math.floor(nowInMicroseconds / 100000);
+    const nowInSeconds = Math.floor(nowInMicroseconds / 1000000);
```
fikus, line 60: Sorry, also please rewrite this test so it is not duplicating the implementation.

ryan-minyu: Did you mean the `const nowInMicroseconds = Date.now() * 1000;` is duplicated with `const nowInMilliseconds = Date.now();` for instance?
We can have literal values for second/milliseconds/microsecond to finish the tests?

fikus: It's the same issue as the one mentioned here: https://github.com/subscriptions-project/swg-js/pull/2415#pullrequestreview-1157110290. The test should not duplicate the implementation logic, since then the test will pass even if the implementation is faulty (as happened when this bug was introduced).

ryan-minyu: Thanks, let me know if the new commit addressed your concern.

fikus: Yes, thanks!

PR #2437, opened 2022-10-27
Do not show toast on completion for survey.

oyj9109:
Survey will display confirmation prompt within its own iframe. Toast on completion should be displayed only for the opt-in actions (regwall/newsletter) which have dedicated message.

Internal tracker: b/255762663


src/runtime/audience-action-flow.js
```diff
@@ -186,16 +186,22 @@ export class AudienceActionFlow {
...
+        break;
+      case 'TYPE_NEWSLETTER_SIGNUP':
+        customText = msg(
+          SWG_I18N_STRINGS.NEWSLETTER_SIGNED_UP_LANG_MAP,
+          lang
+        ).replace(placeholderPatternForEmail, userEmail);
+        break;
+      default:
+        // Do not show toast for other types.
+        return;
```
ChrisAntaki, line 204: Would it be possible to add test coverage for this line?

oyj9109: Added!

ChrisAntaki: Thanks!

ChrisAntaki (approved):

PR #2439, opened 2022-10-27
Improves coverage by always setting default origin

ChrisAntaki:
This PR removes an uncovered line by always setting a default origin


ChrisAntaki:
At first I tried adding a test that would trigger the `else` condition and cover line 129, but I actually couldn't find a URL that would do this, at least in Chrome

jpettitt:
I think origin also includes port

jpettitt (approved):

PR #2441, opened 2022-10-31
refactor more eligibility checks to use isGaEligible and isGtagEligib…

justinchou-google:
refactor more eligibility checks to use isGaEligible and isGtagEligible static functions

### DIFF
--- a/src/runtime/google-analytics-event-listener.js
+++ b/src/runtime/google-analytics-event-listener.js
@@ -22,8 +22,8 @@ export class GoogleAnalyticsEventListener {
    * @param {!./deps.DepsDef} deps
    */
   constructor(deps) {
-    /** @private @const {!Window} */
-    this.win_ = deps.win();
+    /** @private @const {!./deps.DepsDef} deps */
+    this.deps_ = deps;
 
     /** @private @const {!./client-event-manager.ClientEventManager} */
     this.eventManager_ = deps.eventManager();
@@ -46,8 +46,8 @@ export class GoogleAnalyticsEventListener {
   handleClientEvent_(event, eventParams = undefined) {
     // Bail immediately if neither ga function (analytics.js) nor gtag function (gtag.js) exists in Window.
     if (
-      typeof this.win_.ga !== 'function' &&
-      typeof this.win_.gtag !== 'function'
+      !this.constructor.isGaEligible(this.deps_) &&
+      !this.constructor.isGtagEligible(this.deps_)
     ) {
       return;
     }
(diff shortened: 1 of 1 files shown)


src/runtime/google-analytics-event-listener.js
```diff
@@ -73,13 +73,13 @@ export class GoogleAnalyticsEventListener {
       eventLabel: analyticsParams.event_label || gaEvent.eventLabel,
     };
     // TODO(b/234825847): Remove it once universal analytics is deprecated in 2023.
-    const ga = this.win_.ga || null;
-    if (isFunction(ga)) {
+    if (this.constructor.isGaEligible(this.deps_)) {
+      const ga = this.deps_.win().ga || null;
       ga('send', 'event', gaEvent);
```
ChrisAntaki, line 78: If `ga` is null, this would throw an error, right?

oyj9109 (approved):

src/runtime/google-analytics-event-listener.js
```diff
@@ -73,13 +73,13 @@ export class GoogleAnalyticsEventListener {
...
     }
 
-    const gtag = this.win_.gtag || null;
-    if (isFunction(gtag)) {
+    if (this.constructor.isGtagEligible(this.deps_)) {
+      const gtag = this.deps_.win().gtag || null;
```
ChrisAntaki, line 82: if `gtag` is `null`, then line 89 would throw an error, right?

src/runtime/google-analytics-event-listener.js
```diff
@@ -46,8 +46,8 @@ export class GoogleAnalyticsEventListener {
   handleClientEvent_(event, eventParams = undefined) {
     // Bail immediately if neither ga function (analytics.js) nor gtag function (gtag.js) exists in Window.
     if (
-      typeof this.win_.ga !== 'function' &&
-      typeof this.win_.gtag !== 'function'
+      !this.constructor.isGaEligible(this.deps_) &&
```
ChrisAntaki, line 49: I think `GoogleAnalyticsEventListener` would be more readable than `this.constructor`, thoughts?

src/runtime/google-analytics-event-listener.js
justinchou-google: It shouldn't actually ever be null due to the `isGaEligible` check; is there a way to satisfy the type check? Since `ga`/`gtag` are not defined on the `Window` type, without the `null` fallback it complains:
```
ERROR - [JSC_INEXISTENT_PROPERTY] Property gtag never defined on Window

Comment (ChrisAntaki) [reply src/runtime/google-analytics-event-listener.js:78]: You could try casting the type with inline JSDoc comments?

Comment (justinchou-google) [reply src/runtime/google-analytics-event-listener.js:81]: Neat, how is this?

File: src/runtime/google-analytics-event-listener.js (hunk 2 above)
Comment (justinchou-google) [reply src/runtime/google-analytics-event-listener.js]: Can consolidate into above thread, marking resolved.

File: src/runtime/google-analytics-event-listener.js (hunk 1 above)
Comment (ChrisAntaki) [reply src/runtime/google-analytics-event-listener.js:81]: I was thinking something more like this, when you declare the `ga` variable

```
const ga = /** @type {!function(string, string, Object)} */ (this.deps_.win().ga);
```

Could that work?

src/runtime/google-analytics-event-listener.js
justinchou-google: Agree, will change it to use the explicit class name.

ChrisAntaki: Thanks!

src/runtime/google-analytics-event-listener.js
justinchou-google: I believe it's the `Window` type that is the problem: `win().ga`. This suggested change still throws the same error `Property ga never defined on Window`

ChrisAntaki: Good call! I just made some suggestions here via a PR https://github.com/justinchou-google/swg-js/pull/1/

justinchou-google: Thanks, LGTM

ChrisAntaki (approved):

jpettitt (approved):

PR #2446, opened 2022-11-02
Add aria labels to iframes.

rbeckthomas:
This change updates the iframes to have an aria-label. 

b/255709148.

### DIFF
--- a/src/components/dialog.js
+++ b/src/components/dialog.js
@@ -130,6 +130,7 @@ export class Dialog {
 
     /** @private @const {!FriendlyIframe} */
     this.iframe_ = new FriendlyIframe(doc.getWin().document, {
+      'aria-label': 'SwG Dialog iframe',
       'class': iframeCssClass,
     });
 
@@ -267,7 +268,7 @@ export class Dialog {
     iframeBody.appendChild(this.loadingView_.getElement());
 
     // Container for all dynamic content, including 3P iframe.
-    this.container_ = createElement(iframeDoc, 'swg-container', {});
+    this.container_ = createElement(iframeDoc, 'swg-container', {'aria-label': 'SwG iframe Container',});
     iframeBody.appendChild(this.container_);
     this.setPosition_();
 
--- a/src/runtime/analytics-service.js
+++ b/src/runtime/analytics-service.js
@@ -86,7 +86,7 @@ export class AnalyticsService {
 
     /** @private @const {!HTMLIFrameElement} */
     this.iframe_ = /** @type {!HTMLIFrameElement} */ (
-      createElement(this.doc_.getWin().document, 'iframe', {})
+      createElement(this.doc_.getWin().document, 'iframe', {'aria-label': 'SwG Analytics iframe'})
     );
     setImportantStyles(this.iframe_, iframeStyles);
     this.doc_.getBody().appendChild(this.getElement());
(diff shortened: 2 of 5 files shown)


ChrisAntaki:
I think "container" would make more sense than "iframe" for non developers hearing these labels from a screen reader

ChrisAntaki (approved):

fikus (changes requested):
I don't think it's typical to add aria labels saying that elements are containers. We should verify the desired behavior before submitting.

src/components/dialog.js
```diff
@@ -130,6 +130,7 @@ export class Dialog {
 
     /** @private @const {!FriendlyIframe} */
     this.iframe_ = new FriendlyIframe(doc.getWin().document, {
+      'aria-label': 'SwG Dialog Container',
```
fikus, line 133: SwG is not a meaningful term to end users.

fikus, line 133: This is user-facing text so would need to be localized.

fikus (changes requested):

src/components/dialog.js
rbeckthomas: > I don't think it's typical to add aria labels saying that elements are containers. We should verify the desired behavior before submitting.

Makes sense, I double checked the [lighthouse output](https://dequeuniversity.com/rules/axe/4.4/frame-title?utm_source=lighthouse&utm_medium=devtools) and they suggest adding a descriptive title to the frames. 

Perhaps something like 'Subscription offers dialog'? or 'Blocking dialog containing subscription offers'?

Regarding the child iframes I'm not sure how to distinguish them from the parent without using some similar language. Maybe 'Subscription offers dialog activity view', or 'Subscription offers dialog toast view'?

ChrisAntaki: I noticed that link talks about `title` attributes, instead of `aria-label` attributes

fikus: According to the bug, this requirement can be satisfied by using `title`, `aria-label`, `aria-labelledby`, or `role="presentation"`.

I'm not sure of the implications of using role="presentation". If we do need to provide a descriptive label here, I don't know what would be meaningful. We may want to make this a parameter passed when creating a dialog. Another possibility would be to set this from the iframe content JS.

rbeckthomas: I did some investigation and it looks like using `role="presentation"` indicates to the screen reader not to announce those frames, unless they are labeled, which does solve the problem for the nested frame. 

Setting the name from the iframe content would be the most descriptive so I'll pursue that as well.

rbeckthomas: I investigated using iframe content to set the name but the iframe controller doesn't have any access to parent elements so it wouldnt be feasible without sending a message back to swgjs which seems like an unnecessary roundtrip. 

Would it make sense to instead name the dialog generically as `Dialog` and then allow the screen reader to bypass to the content, and then read out the title of the card actually indicating the content.  (will attach a screen recording for clarity)

rbeckthomas: Here is a screencast showing the experience when all the iframes are labeled `role=presentation`
[Screencast no labels.webm](https://user-images.githubusercontent.com/64608436/200056284-48d96452-1542-4b78-bf4f-63b0853574bd.webm)  The relevant portion starts at 25 seconds. As you can see without any labels the screen reader goes directly to the 'scroll offers left button'

however with the label of 'dialog' on the parent iframe the screen reader is able to identify the dialog first before entering the group and announcing the 'scroll offers left button'
[Screencast with label.webm](https://user-images.githubusercontent.com/64608436/200056775-ad760645-2f71-4f08-981a-28a5790d6f12.webm)

I think we should label the iframe as 'dialog' and then update the iframe content so that the screen reader reads the title before reading the navigation buttons.  Thoughts?

PR #2447, opened 2022-11-03
Derive a canonical URL from a page URL when a canonical tag is not present

teiswit:
Pages missing `<link rel="canonical">` could cause an infinite EA grant as we treat them as having the same canonical URL (i.e. an empty string). As canonical pages do not necessarily require the canonical tag, we should treat the page URL as canonical when the canonical tag is not present to avoid the infinite grant issue. 

Internal reference: b/256975824


src/utils/url.js
```diff
@@ -199,12 +199,17 @@ export function getHostUrl(url) {
 }
 
 /**
+ * Returns the canonical URL from the canonical tag. If the canonical tag is
+ * not present, treat the doc URL itself as canonical.
  * @param {!../model/doc.Doc} doc
  * @return {string}
  */
 export function getCanonicalUrl(doc) {
   const node = doc.getRootNode().querySelector("link[rel='canonical']");
```
ChrisAntaki, line 208: Looks like we call `doc.getRootNode()` 3x in this method. Would it make sense to call it once? Ex: `const rootNode = doc.getRootNode()`

src/utils/url.js
```diff
@@ -199,12 +199,17 @@ export function getHostUrl(url) {
...
-  return (node && node.href) || '';
+  return (
+    (node && node.href) ||
```
ChrisAntaki, line 210:
```suggestion
    node?.href ||
```

src/runtime/entitlements-manager-test.js
```diff
@@ -103,6 +108,15 @@ describes.realWin('EntitlementsManager', {}, (env) => {
     deps = new DepsDef();
     sandbox.stub(deps, 'win').returns(win);
     const globalDoc = new GlobalDoc(win);
+
+    globalDoc.getRootNode = function () {
```
ChrisAntaki, line 112: Arrow functions could help this code be more concise. Ex:
```

ChrisAntaki:
This LGTM overall

@jpettitt , could you verify this approach of falling back to the doc's origin+path (without query string) is alright?

src/runtime/entitlements-manager-test.js
teiswit: Thanks for the suggestion. Simplified accordingly.

src/utils/url.js
teiswit: Totally makes sense. Revised accordingly.

ChrisAntaki (approved):

ChrisAntaki:
Looks good! Thanks for being open to feedback

PR #2460, opened 2022-11-07
Fix clicks event logging for SIWG

teiswit:
- Use the click_listener callback provided by SIWG library to capture SIWG click events. The existing onclick listener on a parent div does not work on the child iframe where SIWG is rendered.
- Differentiate GSI vs. SIWG click events.

Internal reference: b/200458976


src/utils/gaa.js
```diff
@@ -870,14 +873,24 @@ export class GaaMeteringRegwall {
...
           isFromUserAction: true,
         });
       }
+      if (
+        e.data.stamp === POST_MESSAGE_STAMP &&
+        e.data.command === POST_MESSAGE_COMMAND_SIWG_BUTTON_CLICK
+      ) {
+        // Log button click event.
+        logEvent({
+          analyticsEvent: AnalyticsEvent.ACTION_SHOWCASE_REGWALL_SWIG_CLICK,
```
ChrisAntaki, line 890: Looks like this PR logs this event in two places. This line isn't covered by tests though  (https://app.codecov.io/gh/subscriptions-project/swg-js/pull/2460). Just checking, does the event need to be logged here?

teiswit (commented):
Thanks for the feedback. PTAL

src/utils/gaa.js
ChrisAntaki: Thanks for explaining, that makes sense

PR #2462, opened 2022-11-07
Add survey submit to localstorage to filter action for users who completed survey

justinchou-google:
b/254325571
Two features:

1. On ACTION_SURVEY_SUBMIT_CLICK, save to localStorage current time for key `surveycompleted`. This mapping is maintained through `COMPLETED_ACTION_TO_STORAGE_KEY_MAP`.
2. During eligibility check of potential actions, for action type `TYPE_REWARDED_SURVEY`, check user has submitted survey and skip action if key `surveycompleted` is saved to local storage.


src/runtime/auto-prompt-manager.js
```diff
@@ -44,6 +45,23 @@ const dismissEvents = [
   AnalyticsEvent.ACTION_CONTRIBUTION_OFFERS_CLOSED,
   AnalyticsEvent.ACTION_SUBSCRIPTION_OFFERS_CLOSED,
 ];
+/** @const {!Array<!AnalyticsEvent>} */
+const completedActions = [AnalyticsEvent.ACTION_SURVEY_SUBMIT_CLICK];
```
oyj9109, line 49: Would `ACTION_SURVEY_DATA_TRANSFER` be more accurate to determine when it's completed?

src/runtime/auto-prompt-manager.js
```diff
@@ -180,45 +198,50 @@ export class AutoPromptManager {
...
-      if (!shouldShowAutoPrompt) {
-        if (
-          this.shouldShowBlockingPrompt_(
-            entitlements,
-            /* hasPotentialAudienceAction */ !!potentialActionPromptType
-          ) &&
-          promptFn
-        ) {
-          promptFn();
+    )
```
oyj9109, line 206: Seems like unnecessary line change

src/runtime/auto-prompt-manager.js
```diff
@@ -365,56 +388,59 @@ export class AutoPromptManager {
...
     article,
     autoPromptType,
     dismissedPrompts,
     shouldShowAutoPrompt,
   }) {
-    let potentialActions = article?.audienceActions?.actions || [];
-    potentialActions = potentialActions.filter((action) =>
+    const audienceActions = article?.audienceActions?.actions || [];
+
+    const filterBooleans = audienceActions.map((action) =>
```
oyj9109, line 401: What about reading local storage in this function and pass an argument in `checkActionEligibility_`?

src/runtime/auto-prompt-manager.js
```diff
@@ -365,56 +388,59 @@ export class AutoPromptManager {
...
       this.checkActionEligibility_(action.type)
     );
+    return Promise.all(filterBooleans).then((booleans) => {
+      let potentialActions = audienceActions.filter((_, i) => booleans[i]);
```
oyj9109, line 405: nit: I think it can be simplified as 
```
 return audienceActions.filter(...).then((potentialActions) => {
    ...
}
```

src/runtime/auto-prompt-manager.js
justinchou-google: I refactored the promises to be resolved prior to the eligibility check, and save the local storage fetches to `fetchedLocalStorage`. PTAL.

src/runtime/auto-prompt-manager.js
justinchou-google: I believe this requires the promises to resolve prior to the filtering action, as filtering on just the promises will always be truthy. Nonetheless, I have refactored this part, so marking resolved.

src/runtime/auto-prompt-manager.js
```diff
@@ -44,6 +46,28 @@ const dismissEvents = [
...
+let LocalStorage;
+/* eslint-enable no-unused-vars */
+
+/**
+ * Returns storage key for a given event.
+ * @param {!AnalyticsEvent} event
+ * @return {string}
+ * @protected
+ */
+function completedActionToStorageKey_(event) {
```
oyj9109, line 63: nit: can we define a map instead of function? then we won't need the allowlist `completedActions` but can just do `if (COMPLETED_ACTION_TO_STORAGE_KEY.has(event)) { ... }`

src/runtime/auto-prompt-manager.js
```diff
@@ -365,56 +392,77 @@ export class AutoPromptManager {
...
-        this.promptDisplayed_ = AutoPromptType.CONTRIBUTION;
-        return undefined;
-      }
-      const previousPrompts = dismissedPrompts.split(',');
-      potentialActions = potentialActions.filter(
-        (action) => !previousPrompts.includes(action.type)
+    return promiseChain.then(() => {
+      // No audience actions means use the default prompt.
+      let potentialActions = audienceActions.filter((action) =>
+        this.checkActionEligibility_(action.type, fetchedLocalStorage)
```
oyj9109, line 427: Can we directly pass in `surveyCompleted`? (Would `surveyCompletions` be representing the value better which is timestamps of completions?)

src/runtime/auto-prompt-manager.js
```diff
@@ -365,56 +392,77 @@ export class AutoPromptManager {
...
-    if (potentialActions.length === 0) {
-      return undefined;
+    const audienceActions = article?.audienceActions?.actions || [];
+
+    const fetchedLocalStorage = /** @type {LocalStorage} */ ({});
+    let promiseChain = Promise.resolve();
+    const actionsIncludeSurvey = audienceActions
+      .map((action) => action.type)
+      .includes(TYPE_REWARDED_SURVEY);
+    if (actionsIncludeSurvey) {
```
oyj9109, line 410: I would prefer making the logic simpler without having this condition unless local storage reads hinder performance.

```
this.getEvent_(
   completedActionToStorageKey_(
      AnalyticsEvent.ACTION_SURVEY_DATA_TRANSFER
   )
)
.then(surveyCompletions => { 
   let potentialActions = audienceActions.filter((action) =>
   ... 
})
```

src/runtime/auto-prompt-manager.js
```diff
@@ -180,45 +204,48 @@ export class AutoPromptManager {
...
-            entitlements,
-            /* hasPotentialAudienceAction */ !!potentialActionPromptType
-          ) &&
-          promptFn
-        ) {
-          promptFn();
+    )
+      .then((shouldShowAutoPrompt_) => {
+        shouldShowAutoPrompt = shouldShowAutoPrompt_;
+        return this.getAudienceActionPromptType_({
```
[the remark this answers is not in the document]
justinchou-google: This has the issue of passing the promise object, rather than resolving it. I instead refactored to use `Promise.all` to pass an array of `potentialActionPromptType` and `shouldShowAutoPrompt` so that the local variable is no longer used; how does this look to you?

src/runtime/auto-prompt-manager.js
justinchou-google: Sure, I originally implemented this way because of the condition around the survey action, but if that is removed then I can pass this directly.

src/runtime/auto-prompt-manager.js
```diff
@@ -44,6 +46,23 @@ const dismissEvents = [
...
+ * @param {!AnalyticsEvent} event
+ * @return {string}
+ * @protected
+ */
+function completedActionToStorageKey_(event) {
+  switch (event) {
+    case AnalyticsEvent.ACTION_SURVEY_DATA_TRANSFER:
+      return STORAGE_KEY_SURVEY_COMPLETED;
+    default:
+      return 'unknownaction';
```
ChrisAntaki, line 63: Would it be possible to cover this line with a test?

justinchou-google: Refactored function into a `Map`, so this is no longer applicable. Marking resolved.

src/runtime/auto-prompt-manager.js
justinchou-google: Refactored into a `Map`, though this creates a new type check warning:
```
src/runtime/auto-prompt-manager.js:51:48: WARNING - [JSC_CONFORMANCE_VIOLATION] Violation: Map is not allowed
  51| const COMPLETED_ACTION_TO_STORAGE_KEY_MAP = new Map([
```

Note, we are already using `Set` which raises the same warning.

src/runtime/auto-prompt-manager.js
justinchou-google: Resolving given more recent discussion below.

src/runtime/auto-prompt-manager.js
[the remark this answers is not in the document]
oyj9109: Looks better! Thanks :)

oyj9109 (approved):

src/runtime/auto-prompt-manager-test.js
```diff
@@ -1540,36 +1589,42 @@ describes.realWin('AutoPromptManager', {}, (env) => {
...
-      storageMock
-        .expects('get')
-        .withExactArgs(
-          STORAGE_KEY_DISMISSED_PROMPTS,
-          /* useLocalStorage */ true
-        )
-        .resolves(dismissedPrompts)
-        .exactly(dismissedPromptGetCallCount);
+  });
+  function setupPreviousImpressionAndDismissals(
```
ChrisAntaki, line 1593: Could this `setupPreviousImpressionAndDismissals` method accept a single object param to improve readability instead of so many params? I'm finding the invocations of the method a little hard to read

ChrisAntaki: For example, instead of this...
```
setupPreviousImpressionAndDismissals(
        storageMock,
        null,
        null,
        null,
        1,
        null,
        true
      );
```

Perhaps this...
```
setupPreviousImpressionAndDismissals({
        storageMock,
        dismissedPromptGetCallCount: 1,
        getUserToken: true,
      });
```

ChrisAntaki: Actually, maybe someone could tackle this in a future PR

src/runtime/auto-prompt-manager.js
```diff
@@ -184,41 +191,45 @@ export class AutoPromptManager {
...
-        article,
-        autoPromptType: params.autoPromptType,
-        dismissedPrompts,
-        shouldShowAutoPrompt,
-      });
-      const promptFn = potentialActionPromptType
-        ? this.audienceActionPrompt_({
-            action: potentialActionPromptType,
+    )
+      .then((shouldShowAutoPrompt) => {
```
ChrisAntaki, line 195: This could be simplified with an arrow function expression (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions)
```suggestion
      .then((shouldShowAutoPrompt) => Promise.all([
```

src/runtime/auto-prompt-manager.js
```diff
@@ -365,56 +376,62 @@ export class AutoPromptManager {
...
+      // No audience actions means use the default prompt.
+      let potentialActions = audienceActions.filter((action) =>
+        this.checkActionEligibility_(action.type, surveysCompleted)
       );
 
-      // If all actions have been dismissed or the frequency indicates that we
-      // should show the Contribution prompt again regardless of previous dismissals,
-      // we don't want to record the Contribution dismissal
-      if (potentialActions.length === 0 || shouldShowAutoPrompt) {
+      if (potentialActions.length === 0) {
```
ChrisAntaki, line 399: Could you add a comment for this check?

src/runtime/auto-prompt-manager.js
```diff
@@ -533,9 +550,9 @@ export class AutoPromptManager {
   }
 
   /**
-   * Listens for relevant prompt impression and dismissal events, and logs them
-   * to local storage for use in determining whether to display the prompt in
-   * the future.
+   * Listens for relevant prompt impression, dismissal events, and completed
```
ChrisAntaki, line 553:
```suggestion
   * Listens for relevant prompt impression events, dismissal events, and completed
```

src/runtime/auto-prompt-manager.js
ChrisAntaki: Maybe that comment from line 394 could move down so it's just above this check, like before?

src/runtime/auto-prompt-manager.js
```diff
@@ -365,56 +376,62 @@ export class AutoPromptManager {
...
-      if (!dismissedPrompts) {
-        this.promptDisplayed_ = AutoPromptType.CONTRIBUTION;
-        return undefined;
-      }
-      const previousPrompts = dismissedPrompts.split(',');
-      potentialActions = potentialActions.filter(
-        (action) => !previousPrompts.includes(action.type)
+    const audienceActions = article?.audienceActions?.actions || [];
+
+    return this.getEvent_(
```
ChrisAntaki, line 389:
```suggestion
    // Count completed surveys.
    return this.getEvent_(
```

src/runtime/auto-prompt-manager.js
```diff
@@ -365,56 +376,62 @@ export class AutoPromptManager {
...
+      COMPLETED_ACTION_TO_STORAGE_KEY_MAP.get(
+        AnalyticsEvent.ACTION_SURVEY_DATA_TRANSFER
+      )
+    ).then((surveysCompleted) => {
```
ChrisAntaki, line 393:
```suggestion
    ).then((surveysCompletionTimestamps) => {
      const surveysCompleted = surveysCompletionTimestamps.length;
```

ChrisAntaki:
Overall it looks good! Just shared some code suggestions

src/runtime/auto-prompt-manager.js
justinchou-google: Ah yes, this was unintended. Thanks for pointing this out.

src/runtime/auto-prompt-manager.js
justinchou-google: Also renamed `completedSurvey` to `hasCompletedSurveys` for clarity.

ChrisAntaki: Awesome

ChrisAntaki (approved):

PR #2464, opened 2022-11-08
Fix gtag and ga test setup

justinchou-google:
add-ons to b/254455836

### DIFF
--- a/src/runtime/auto-prompt-manager-test.js
+++ b/src/runtime/auto-prompt-manager-test.js
@@ -123,15 +123,15 @@ describes.realWin('AutoPromptManager', {}, (env) => {
   });
 
   function setWinWithAnalytics(gtag, ga) {
-    const winWithNoGtag = Object.assign({}, win);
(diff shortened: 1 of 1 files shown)


src/runtime/auto-prompt-manager-test.js
```diff
@@ -122,9 +122,14 @@ describes.realWin('AutoPromptManager', {}, (env) => {
     miniPromptApiMock.verify();
   });
 
-  function setWinWithoutGtag() {
+  function setWinWithAnalytics(gtag, ga) {
     const winWithNoGtag = Object.assign({}, win);
```
ChrisAntaki, line 126: Would renaming this const make sense, since it can contain a gtag now?
```suggestion
    const winWithAnalytics = Object.assign({}, win);
```

justinchou-google: Yes, I agree. Good catch.

ChrisAntaki (approved):

PR #2470, opened 2022-11-14
Corrected typo of button analytic event

xiexr151e:
b/259009380


teiswit (commented):
Thanks for the super fast fix. Could you also update the event name used in gaa.js as well? Thanks!
https://github.com/subscriptions-project/swg-js/blob/main/src/utils/gaa.js#L890

teiswit (approved):

ChrisAntaki (approved):

PR #2469, opened 2022-11-11
Suppress meter toast for survey users

justinchou-google:
b/256888870
Define a new MeterType: `SUPPRESSED` for when to exit iframe flow and do not show meter toast. Currently, this is mapped only to `survey_user`. Other known user types are still mapped to `KNOWN` toast.

PR #2475, opened 2022-11-17
Consume callback on suppressed meter flow

justinchou-google:
b/256888870

In PR: https://github.com/subscriptions-project/swg-js/pull/2469, meter toast is suppressed for survey users, as we plan on implementing a toast dedicated for survey completion. This change is to consume the free read entitlement earned by the survey completion.


hartmannr76:
Spoke offline: lets hold off on merging this for now. We suppress on the server so there doesn't seem to be a need to add this logic

PR #2476, opened 2022-11-17
Add API to initiate subscription linking

fikus:
Add a method to initiate the subscription linking flow. The method opens /linksaveiframe with the appropriate parameters for subscription linking. When the iframe messages the result, it returns the outcome to the caller.

b/236000111

### DIFF
--- a/src/api/subscriptions.js
+++ b/src/api/subscriptions.js
@@ -265,6 +265,13 @@ export class Subscriptions {
   saveSubscription(requestCallback) {}
 
   /**
+   * Starts the subscription linking flow.
+   * @param {!LinkSubscriptionRequest} linkSubscriptionRequest
+   * @return {!Promise<!LinkSubscriptionResult>} promise indicating result of the operation
+   */
+  linkSubscription(linkSubscriptionRequest) {}
+
+  /**
    * Creates an element with the SwG button style and the provided callback.
    * The default theme is "light".
    *
@@ -640,3 +647,18 @@ export let SmartButtonOptions;
  * }}
  */
 export let SubscriptionRequest;
+
+/**
+ * @typedef {{
+ *  publisherProvidedId: string,
+ * }}
+ */
+export let LinkSubscriptionRequest;
+
+/**
+ * @typedef {{
+ *  publisherProvidedId: (?string|undefined),
+ *  success: boolean,
+ * }}
+ */
+export let LinkSubscriptionResult;
--- a/src/runtime/subscription-linking-flow.js
+++ b/src/runtime/subscription-linking-flow.js
@@ -1,0 +1,83 @@
+/**
+ * Copyright 2022 The Subscribe with Google Authors. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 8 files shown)


hartmannr76 (approved):

src/runtime/subscription-linking-flow.js
```diff
@@ -0,0 +1,83 @@
...
+  /**
+   * Starts the subscription linking flow.
+   * @param {!../api/subscriptions.LinkSubscriptionRequest} request
+   * @return {!Promise<!../api/subscriptions.LinkSubscriptionResult>}
+   */
+  start(request) {
+    const {publisherProvidedId} = request;
+    const publicationId = this.pageConfig_.getPublicationId();
+    const args = feArgs({
+      'publicationId': publicationId,
```
ChrisAntaki, line 51: This could be simplified a bit
```suggestion
      publicationId,
```

src/runtime/subscription-linking-flow.js
```diff
@@ -0,0 +1,83 @@
...
+    });
+    const activityIframeView = new ActivityIframeView(
+      this.win_,
+      this.activityPorts_,
+      feUrl('/linksaveiframe', {
+        'subscriptionLinking': true,
```
ChrisAntaki, line 57: Swgjs' compiler doesn't change object property names, so you could remove these single quotes
```suggestion
        subscriptionLinking: true,
```

ChrisAntaki (approved):

src/runtime/subscription-linking-flow.js
fikus: I wasn't sure if our compiled JS would rename unquoted property names. Either way, in my mind the two styles imply different meanings: quoted strings are values and unquoted strings are symbol names. I chose to use quoted property names here because the string values are significant. What do you think?

ChrisAntaki: Swgjs' compiler renames local symbols and private methods (with the `_` suffix). Personally, I would drop the quotes, but it's your call

src/runtime/subscription-linking-flow.js
fikus: Done.

src/runtime/subscription-linking-flow.js
fikus: Done.

fikus:
@jpettitt please take a look for the API changes.

src/api/subscriptions.js
```diff
@@ -640,3 +647,18 @@ export let SmartButtonOptions;
  * }}
  */
 export let SubscriptionRequest;
+
+/**
+ * @typedef {{
+ *  publisherProvidedId: string,
```
jpettitt, line 653: do we also need an (optional?) initial set of product ID's and expire info here - otherwise we'll have to wait for the publisher to push an update before the link become effective.

jpettitt: On reflection bad idea as it would introduce an abuse vector- we do need to document that the publisher must make the backend call to push subscription datas as soon as the link completes otherwise the use could be left wondering why the account they just linked can't access things.

jpettitt (approved):

PR #2477, opened 2022-11-17
Simplify AMP release script

ChrisAntaki:
This PR simplifies the AMP release script. Now the script pulls from the new `amp` branch. We can kick off AMP releases whenever that branch has important updates


shortcuts.sh
```diff
@@ -132,32 +129,26 @@ function swgjs_create_amp_release_pr() {
...
   __swgjs_install_amp
 
   # Build Swgjs for AMP.
   cd $SWGJS_PATH
   git fetch team
-  git checkout team/main
+  git checkout team/amp
   npx gulp build
-  npx gulp export-to-amp --swgVersion=$SWG_VERSION
+  npx gulp export-to-amp --swgVersion=AMP
```
hartmannr76, line 140: Is it ok for this to always be the same version going forward?

ChrisAntaki: Yes, I think that's fine. AMP has its own build/versioning system, and this script just exports a dependency for that

hartmannr76: Got it. Would it make sense for us to still maintain a version just so we can tie a specific export to a swg version? I imagine we should still create git tags for the amp versions, they should just be treated differently than the main line

ChrisAntaki: Good idea, I added the Git commit hash as a version

hartmannr76 (approved):

shortcuts.sh
```diff
@@ -132,19 +129,19 @@ function swgjs_create_amp_release_pr() {
...
   __swgjs_install_amp
 
-  # Build Swgjs for AMP.
-  cd $SWGJS_PATH
+  # Switch to clean AMP branch.
   git fetch team
-  git checkout team/main
+  git checkout team/amp
+  if [[ `git status --porcelain` ]]; then
+    # Avoid building with local changes.
```
fikus, line 139: Show an error message?

shortcuts.sh
```diff
@@ -132,19 +129,19 @@ function swgjs_create_amp_release_pr() {
...
+    return
```
fikus, line 140: return value should be non-zero to indicate failure.

shortcuts.sh
```diff
@@ -132,19 +129,19 @@ function swgjs_create_amp_release_pr() {
...
+  fi
+
+  # Build Swgjs for AMP.
+  SWG_VERSION=$(git rev-parse --short HEAD)
```
fikus, line 144: I think that pushing from HEAD should generally be avoided because it could point to an unexpected revision when the script runs (for example, if one engineer pushes a commit while another engineer is making the release). That might not be an issue here if we aren't expecting much activity on the amp branch though.

shortcuts.sh
ChrisAntaki: By this point, Git already showed a pretty descriptive error message

shortcuts.sh
ChrisAntaki: Yeah, this seems safe for the `amp` branch specifically, since the same people working on the branch will be kicking of the release on-demand

shortcuts.sh
ChrisAntaki: Good call! I just created https://github.com/subscriptions-project/swg-js/pull/2488 to fix this

shortcuts.sh
fikus: When does git show an error message? git st --porcelain does now show an error if there are local changes.

ChrisAntaki: Sorry about that. During testing, I was seeing an error message from Git talking about local changes. I'm not sure what was triggering that. Anyways, I'll create a new PR adding an error message. Nice catch on this!

ChrisAntaki: Here's the error message I'm seeing from Git _sometimes_
image

ChrisAntaki: Anyways, I just added an error message here: https://github.com/subscriptions-project/swg-js/pull/2488

PR #2489, opened 2022-11-21
Move constants to utils/constants

justinchou-google:


justinchou-google:
There are some references to third_party constants that are not changed:  
https://github.com/subscriptions-project/swg-js/blob/main/third_party/gpay/src/redirect_verifier_helper.js#L89,L108

though I do not see it defined in https://github.com/subscriptions-project/swg-js/blob/main/third_party/gpay/src/constants.js

src/utils/constants.js
```diff
@@ -23,6 +23,67 @@ const Constants = {};
  */
 Constants.USER_TOKEN = 'USER_TOKEN';
 
+/**
+ * Local storage key for read time.
+ *
+ * @const {string}
+ */
 Constants.READ_TIME = 'READ_TIME';
```
justinchou-google, line 31: Excluding these keys from standardization, but LMK if they should be included.

src/runtime/audience-action-flow.js
```diff
@@ -305,7 +303,9 @@ export class AudienceActionFlow {
           AnalyticsEvent.EVENT_SURVEY_DATA_TRANSFER_FAILED,
           /* isFromUserAction */ false
         );
-      this.storage_.storeEvent(STORAGE_KEY_EVENT_SURVEY_DATA_TRANSFER_FAILED);
+      this.storage_.storeEvent(
+        Constants.STORAGE_KEY_EVENT_SURVEY_DATA_TRANSFER_FAILED
```
ChrisAntaki, line 307: Is it possible to create a single object for all the storage keys? I think that might improve readability
```suggestion
        StorageKeys.EVENT_SURVEY_DATA_TRANSFER_FAILED
```

ChrisAntaki (approved):

jpettitt (approved):

PR #2491, opened 2022-11-21
Async/await for src/utils/gaa.js

ChrisAntaki:


src/utils/gaa.js
```diff
@@ -1388,13 +1379,15 @@ export class GaaGoogle3pSignInButton {
       tabIndex: 0,
     });
     buttonEl./*OK*/ innerHTML = GOOGLE_3P_SIGN_IN_BUTTON_HTML;
-    buttonEl.onclick = () => {
+    buttonEl.onclick = async () => {
+      // TODO: Consider awaiting this promise.
```
fikus, line 1383: If there is some future work to do for this, please add a bug ID to the TODO for reference.

fikus (approved):

src/utils/gaa.js
ChrisAntaki: Sounds good, I just added a reference to b/259994494, and will follow up with a PR resolving the TODO

ChrisAntaki: Here's a PR that resolves the TODO
https://github.com/subscriptions-project/swg-js/pull/2492

PR #2495, opened 2022-11-22
Async/await for util scripts

ChrisAntaki:
Updates some util scripts with async/await where it helps

### DIFF
--- a/src/utils/string.js
+++ b/src/utils/string.js
@@ -206,6 +206,5 @@ export async function hash(stringToHash) {
     return Promise.reject(message);
   }
 
-  const digest = await subtle.digest('SHA-512', utf8EncodeSync(stringToHash));
-  return toHex(digest);
+  return toHex(await subtle.digest('SHA-512', utf8EncodeSync(stringToHash)));
 }
--- a/src/utils/string.js
+++ b/src/utils/string.js
@@ -196,7 +196,7 @@ function toHex(buffer) {
  * @param {string} stringToHash
  * @return {!Promise<string>}
  */
-export function hash(stringToHash) {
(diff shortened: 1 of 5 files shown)


src/utils/string.js
```diff
@@ -206,7 +206,6 @@ export function hash(stringToHash) {
     return Promise.reject(message);
   }
 
-  return subtle
-    .digest('SHA-512', utf8EncodeSync(stringToHash))
-    .then((digest) => toHex(digest));
+  const digest = await subtle.digest('SHA-512', utf8EncodeSync(stringToHash));
+  return toHex(digest);
```
hartmannr76, line 210: Can you not just inline it?

```js
return toHex(await subtle.digest('SHA-512', utf8EncodeSync(stringToHash)));`
```

src/utils/xhr.js
```diff
@@ -320,21 +325,9 @@ export class FetchResponse {
...
-      this.drainText_().then(parseJson)
-    );
-  }
-
-  /**
-   * Drains the response and returns a promise that resolves with the response
-   * ArrayBuffer.
-   * @return {!Promise<!ArrayBuffer>}
-   */
-  arrayBuffer() {
```
hartmannr76, line 334: Unused?

hartmannr76 (approved):

src/utils/xhr.js
ChrisAntaki: Yes... also it seems to return a Uint8Array instead of an ArrayBuffer

ChrisAntaki: I'll rollback the changes to XHR for this PR. I'm seeing some opportunities to remove polyfills

PR #2496, opened 2022-11-22
Encode swgUserToken prior to adding it as a query param

ksebina:
This is a fix for b/257991492 which investigates the failure of DecryptSwgUserToken when ramping up the logging-audience-activity experiment. The swgUserToken param was not encoded prior to adding it as  a query param therefore it was replacing "+" characters with white spaces which caused issues when the DecryptSwgUserToken method was called.


fikus (dismissed):
Can you add a unit test that covers this?

fikus:
Sorry, I didn't notice this was still a draft PR.

src/runtime/audience-activity-listener.js
```diff
@@ -84,14 +84,15 @@ export class AudienceActivityEventListener {
           const pubId = encodeURIComponent(
             this.deps_.pageConfig().getPublicationId()
           );
+          const encodedSwgUserToken = encodeURIComponent(swgUserToken);
           const audienceActivityClientLogsRequest =
             this.createLogRequest_(event);
           const url = serviceUrl(
```
oyj9109, line 90: Can we use `addQueryParam` function which handles encoding instead of manually appending it?

```
const url = `/publication/${pubId}/audienceactivity`;
this.fetcher_.sendBeacon(serviceUrl(url), audienceActivityClientLogsRequest);
```

src/runtime/audience-activity-listener-test.js
```diff
@@ -94,8 +94,9 @@ describes.realWin('AudienceActivityEventListener', {}, (env) => {
     );
 
     const path = new URL(capturedUrl);
-    expect(path.pathname).to.equal(
-      '/swg/_/api/v1/publication/pub1/audienceactivity'
+    expect(path.toString()).to.equal(
+      '$frontend$/swg/_/api/v1/publication/pub1/audienceactivity?sut=' +
```
oyj9109, line 98: What about checking the encoded value here to make sure it fixes the specific issue? (i.e., no white space in the url)
`'...?sut=ab%2Bc'`

src/runtime/audience-activity-listener.js
ksebina: Changed!

oyj9109 (approved):

hartmannr76 (approved):

PR #2497, opened 2022-11-22
Removes XHR polyfill

ChrisAntaki:
Swgjs' supported browsers have Beacon and Fetch APIs
- https://caniuse.com/beacon
- https://caniuse.com/fetch


src/runtime/fetcher.js
```diff
@@ -33,8 +32,8 @@ export class Fetcher {
 
   /**
    * @param {string} unusedUrl
-   * @param {!../utils/xhr.FetchInitDef} unusedInit
-   * @return {!Promise<!../utils/xhr.FetchResponse>}
+   * @param {!Object} unusedInit
```
fikus, line 35: Can this be typed as `RequestInit`?

src/utils/xhr.js
```diff
@@ -1,369 +0,0 @@
...
-    return this.fetch_(input, init)
-      .catch((reason) => {
-        /*
-         * If the domain is not valid for SwG we return 404 without
-         * CORS headers and the browser throws a CORS error.
-         * We include some helpful text in the message to point the
-         * publisher towards the real problem.
-         */
-        const targetOrigin = parseUrl(input).origin;
-        throw new Error(
```
fikus, line 104: Preserve this behavior?

src/utils/xhr.js
```diff
@@ -1,369 +0,0 @@
...
-    const creds = init.credentials;
-    assert(
-      creds === undefined || creds == 'include' || creds == 'omit',
-      'Only credentials=include|omit support: %s',
-      creds
-    );
-    // Fallback to xhr polyfill since `fetch` api does not support
-    // responseType = 'document'. We do this so we don't have to do any parsing
-    // and document construction on the UI thread which would be expensive.
-    if (init.responseType == 'document') {
```
fikus, line 82: Do we need to preserve this behavior? responseType in init is not part of the fetch standard AFAICT.

ChrisAntaki: Swgjs only uses `fetch` in a few places, through the Fetcher. In those places, I didn't see any evidence we needed this behavior

src/runtime/fetcher.js
ChrisAntaki: That works, added !

src/utils/xhr.js
ChrisAntaki: Sure, I just added it here:
suggests devs look at Publisher Center setup when a fetch fails

fikus (approved):

PR #2520, opened 2022-11-28
Modernize sample publisher code

ChrisAntaki:
Mainly using async/await


examples/sample-pub/metering.js
```diff
@@ -73,7 +76,7 @@ const MeteringDemo = {
    * so it lives past 7 days in Safari.
    * https://webkit.org/blog/10218/full-third-party-cookie-blocking-and-more/
    */
-  registerUser: (gaaUser) => {
+  registerUser: async (gaaUser) => {
```
fikus, line 79: Doesn't look like this needs to be async.

fikus (approved):

examples/sample-pub/metering.js
ChrisAntaki: Removed 👍

PR #2527, opened 2022-11-29
Fix pop up on expand animation.

oyj9109:
Context: on expansion, immediate style change is queued and then transition animation begins. They could collide within the same animation frame which resulted in pop ups (refer to the below screencasts). 
https://github.com/subscriptions-project/swg-js/blob/02339a41383c78d01c3f9b2bbc1249f553d15ca9/src/components/dialog.js#L504-L514

Calling `requestAnimationFrame` to make sure they are executed in the expected time sequence.

Before: https://screencast.googleplex.com/cast/NTg0NDI5OTEyNTA5NjQ0OHw2YWVjMGEwOC0xZA
After: https://screencast.googleplex.com/cast/NTc3MDE0NzE4NTgxOTY0OHxiM2E4YjdhZi1jZA


fikus (approved):

oyj9109:
Added availability check in case some browser doesn't support the function. (Refer to https://caniuse.com/?search=requestAnimationFrame)

src/components/dialog.js
```diff
@@ -504,14 +504,23 @@ export class Dialog {
...
+            transition(
+              this.getElement(),
+              {
+                'transform': this.getDefaultTranslateY_(),
+              },
+              300,
+              'ease-out'
+            );
+          };
+          if (typeof requestAnimationFrame === 'function') {
```
ChrisAntaki, line 517: So I think this check could be removed. Swgjs' supported browsers have the `requestAnimationFrame` method

oyj9109: Oh cool, good to know. Reverted the last commit.

ChrisAntaki: Thanks!

PR #2540, opened 2022-12-01
Fix a minified swgjs issue

ryan-minyu:
Fix a issue that cause converting Timestamp failed in minified swgjs.


ChrisAntaki:
Nice catch! The prod build shortens property names that end with _

ChrisAntaki (approved):

jpettitt (approved):

PR #2541, opened 2022-12-01
Add a fallback for getCanonicalUrl 

teiswit:
Fix an issue for getCanonicalUrl in AMP when both canonical tag and `rootNode.location` are undefined by returning an empty string. This was a default behavior prior to https://github.com/subscriptions-project/swg-js/pull/2447

Internal tracking: b/260953276


ChrisAntaki:
Instead of the `main` branch, could you target the `amp` branch for AMP-specific fixes?

teiswit:
@ChrisAntaki Done - changed the target to `amp` branch.

jpettitt:
Looks like this picked up all the changes to main, we need to apply only the canonical fix to the AMP branch as it's diverging from main intentionally.

jpettitt (changes requested):
We need to untangle this change from the main branch changes before the PR and go ahead

teiswit:
Attempted to undo the commit in this PR, but was unable to untangle the change. Re-work this change by starting fresh from amp branch in #2562 and closing this PR.

PR #2544, opened 2022-12-01
Add EVENT_SURVEY_DATA_TRANSFER_COMPLETE action for when all callback…

justinchou-google:
…s are completed. Used to consolidate all data transfer actions into one for metrics dashboard.

b/261006300


src/runtime/audience-action-flow.js
```diff
@@ -344,6 +344,12 @@ export class AudienceActionFlow {
       };
       this.deps_.eventManager().logEvent(event, eventParams);
     });
+    this.deps_.eventManager().logEvent({
```
oyj9109, line 347: (Optional) What about logging this `handleSurveyDataTransferRequest_` where failure is logged to make it clear that either of them is logged?

oyj9109 (approved):

src/runtime/audience-action-flow.js
justinchou-google: Will we be able to distinguish between success and failures in the logs? Or are we not concerned with that?

oyj9109: I meant putting it in the `else` statement of `if (!gaLoggingSuccess)` so that either of failure or success is logged.

justinchou-google: Ah I see, yes I can do that.

ChrisAntaki (approved):

PR #2545, opened 2022-12-01
Demo site: Load qual JS for qual test pages

fikus:
For demo pages in the qual tree, load the qual JS file. Currently the qual pages load prod JS and do not work correctly.

Noticed as part of b/258726652.


ChrisAntaki (approved):

ChrisAntaki:
Nice catch!

PR #2550, opened 2022-12-01
Add mini prompt events

mhkawano:
Events needed for collecting analytics of mini prompts.


src/proto/api_messages.js
```diff
@@ -22,7 +22,7 @@
 /**
  * @interface
  */
-class Message {
+ class Message {
```
ChrisAntaki, line 25: Is this extra space necessary?

ChrisAntaki (approved):

src/proto/api_messages.js
mhkawano: No, removed. Not in the original file, which this should be copy-pasted from, so I'm not sure how it sneaked in.

PR #2560, opened 2022-12-02
Add position attribute to swg-disable-scroll class to prevent background scrolling on mobile

justinchou-google:
On mobile, the background publication is still scrollable when the user pinch-zooms. This change fixes this behavior and prevents zoom regardless of scroll on iOS Safari and iOS Chrome. Please see screen record artifact in b/259247789 and b/259252473.


oyj9109 (approved):

justinchou-google:
Setting `touch-action` on the body prevented scroll when the user scrolled on the background dom, but the user can still scroll when touching the iframe element. Updated change prevents background scrolling in both areas.

fikus (approved):

PR #2576, opened 2022-12-06
Fix async tests in gaa-test.js

ChrisAntaki:
Lets avoid `expect()` calls within `.then()` callbacks, since that can easily lead to the testing framework ignoring errors that happen after the test finishes. Async/await is a great way to ensure something happens before a test completes

### DIFF
--- a/src/utils/gaa-test.js
+++ b/src/utils/gaa-test.js
@@ -607,7 +607,7 @@ describes.realWin('GaaMeteringRegwall', {}, () => {
       });
     });
 
-    it('resolves with a gaaUser removes Regwall from DOM on click', () => {
+    it('resolves with a gaaUser removes Regwall from DOM on click', async () => {
       const gaaUserPromise =
         GaaMeteringRegwall.showWithNativeRegistrationButton({
           googleApiClientId: GOOGLE_API_CLIENT_ID,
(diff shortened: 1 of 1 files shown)


src/utils/gaa-test.js
```diff
@@ -2977,23 +2976,23 @@ describes.realWin('GaaMetering', {}, () => {
...
 
   describe('getLoginPromise', () => {
-    beforeEach(() => {
-      GaaMetering.loginPromiseResolve_ = undefined;
-    });
-    it('sets up the promise', () => {
-      GaaMetering.getLoginPromise().then(() => {
-        expect(GaaMetering.loginPromiseResolve_).to.have.been.called();
-      });
+    beforeEach(() => {});
```
fikus, line 2989: Remove?

src/utils/gaa-test.js
```diff
@@ -2977,23 +2976,23 @@ describes.realWin('GaaMetering', {}, () => {
...
+    it('sets up the promise', async () => {
+      const promise = GaaMetering.getLoginPromise();
+      sandbox.spy(GaaMetering, 'loginPromiseResolve_');
       GaaMetering.resolveLogin();
+      await promise;
+      expect(GaaMetering.loginPromiseResolve_).to.have.been.called;
```
fikus, line 2995: I think this should just expect `getLoginPromise.to.be.fulfilled`, rather than looking at the private field `loginPromiseResolve_` (existing issue, fine if you don't want to change it here).

fikus (approved):

src/utils/gaa-test.js
ChrisAntaki: Updated 👍

PR #2578, opened 2022-12-06
[SL] Allow specifying PPID on demo sites

fikus:
Allow specifying a PPID when invoking the subscription linking test flow. Populate the PPID to a random number by default.

Also moves shared logic and HTML structure to shared JS file to reduce duplication.

b/258726652

### DIFF
--- a/demos/public/subscription-linking.js
+++ b/demos/public/subscription-linking.js
@@ -1,0 +1,52 @@
+/**
+ * Copyright 2022 The Subscribe with Google Authors. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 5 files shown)


demos/public/subscription-linking.js
```diff
@@ -0,0 +1,53 @@
...
+  `;
+}
+
+(async () => {
+  createForm();
+  const ppid = randomPpid();
+  document.getElementById('ppid-input').value = ppid;
+})();
+
+window.linkSubscription = linkSubscription;
```
ChrisAntaki, line 53: This line could be removed, since `linkSubscription` would already be on the global scope

fikus: Removed. Without this I did need to add `eslint-disable no-unused-vars` for the function.

ChrisAntaki: Oh that's kinda annoying. Would `eslint-disable-next-line no-unused-vars` look a little cleaner?

PR #2579, opened 2022-12-06
Modernize async/await code in runtime.js

ChrisAntaki:


src/runtime/runtime-test.js
```diff
@@ -401,7 +401,7 @@ describes.realWin('Runtime', {}, (env) => {
       resolveStub = sandbox
         .stub(PageConfigResolver.prototype, 'resolveConfig')
         .callsFake(() => configPromise);
-      runtime.configuredPromise_.then((configuredRuntime) => {
+      runtime.configuredRuntimePromise_.then((configuredRuntime) => {
```
jpettitt, line 404: Missed a `then()`

jpettitt:
Thank you for commenting the flow

src/runtime/runtime-test.js
ChrisAntaki: Nice catch. Looks like the `analyticsMock` expectations weren't being verified either. Looking into fixing this...

ChrisAntaki: Fixed 👍

jpettitt (approved):

PR #2585, opened 2022-12-06
Modernize async/await code in toast.js

ChrisAntaki:


src/ui/toast.js
```diff
@@ -95,67 +95,65 @@ export class Toast {
...
+      this.src_,
+      this.args_
+    );
+    await port.whenReady();
+    resetStyles(this.iframe_, ['height']);
+
+    this.animating_ = this.animate_(() => {
+      setImportantStyles(this.iframe_, {
+        'transform': 'translateY(100%)',
+        'opactiy': 1,
```
fikus, line 111: Typo for `opacity` (already existing issue).

fikus (approved):

src/ui/toast.js
ChrisAntaki: Nice catch, fixed

PR #2588, opened 2022-12-06
Simplify unit test infra

ChrisAntaki:
- Removes unused `specs` param from `describes.realWin` and `describes.sandboxed` methods
- Removes unused fixtures code (`describes.fixtures`, `IntegrationFixture`, "test/fixtures/fixture.js", etc)
- Modernize async/await code
- Simplify code when possible (reducing nesting by terminating functions earlier when possible instead of having expansive if/else branches, moving from forEach() to for/of, filter() or map(), using arrow functions when possible, etc)
- Removes unused early prototypes from ~5 years ago


ChrisAntaki:
Good call, added some highlights

jpettitt (approved):

PR #2596, opened 2022-12-08
Support google analytics logging via gtm.

oyj9109:
Note that this will take effect for the specific events that publisher manually set up in their GTM account, which is describe in Step 1 in [News Tagging Guide (NTG)](https://newsinitiative.withgoogle.com/training/datatools/ntg?platform=gtm&id=UA-12-1&content=reader-revenue__membership
) and other events will be discarded by gtm. This change is to support Step 2.


src/runtime/google-analytics-event-listener-test.js
```diff
@@ -28,21 +28,28 @@ import {analyticsEventToGoogleAnalyticsEvent} from './event-type-mapping';
...
   let win;
   let winMock;
+  let dataLayerMock;
   let eventManager;
   let deps;
   let listener;
 
   beforeEach(() => {
     sandbox.stub(self.console, 'log');
+    dataLayerMock = sandbox.mock({
```
ChrisAntaki, line 38: Could you add `dataLayerMock.verify()` to the `afterEach` method? Without this, the tests won't verify the mock's expectations

fikus: I'd prefer any assertions to go into the individual tests rather than `afterEach`. This is widespread in the codebase but it's an anti-pattern to perform assertions in setup/teardown code.

ChrisAntaki: I view calling `verify()` on a mock different;y than defining a specific expectation within a test

ChrisAntaki: For instance, this expectation never gets checked if `verify()` isn't called on the mock
image

ChrisAntaki: In that case, a specific test creates the expectation

ChrisAntaki: I'm wondering if it's worth every test setting expectations on mocks to also call `.verify()` on the mock... perhaps?

fikus: In general I would prefer calling `verify` in each test. This has some advantages:

- Highlights the specific verification being performed in the test.
- Tests look more like arrange, act, assert instead of assert, arrange, act.
- The test runner will report failed expectations as a failure in the test case instead of an error in afterEach.

Given the style of the mock API we're using this may not always be practical, or could cause broken tests to pass because the author forgets to add `verify`. So, I guess I'd leave it up to the author.

ChrisAntaki: Good points, I'm liking this idea more now

oyj9109: Makes sense. Reorganized the tests to have `verify` where it's needed.

ChrisAntaki (approved):

PR #2603, opened 2022-12-09
Improve coverage in log-test.js

ChrisAntaki:
- Removes unecessary check that included an uncovered line
- Re-enables eslint

### DIFF
--- a/src/utils/log-test.js
+++ b/src/utils/log-test.js
@@ -20,7 +20,7 @@ describes.realWin('warn', () => {
   let warnFn;
 
   beforeEach(() => {
-    warnFn = sandbox.spy(console, 'warn');
+    warnFn = sandbox.spy(self.console, 'warn');
   });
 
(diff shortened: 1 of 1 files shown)


src/utils/log-test.js
```diff
@@ -31,15 +29,15 @@ describes.realWin('warn', () => {
 
   it('should log a warning', () => {
     warn('Hello World');
-    expect(console.warn.calledWith('Hello World')).to.be.true;
+    expect(self.console.warn.calledWith('Hello World')).to.be.true;
```
fikus, line 32: OOC can this be written `.to.be.calledWith(...)`? .to.be.true seems redundant.

fikus (approved):

src/utils/log-test.js
ChrisAntaki: Good call, updated

PR #2609, opened 2022-12-09
Improve coverage in callbacks.js

ChrisAntaki:
- Adds tests for warning messages

### DIFF
--- a/src/runtime/callbacks-test.js
+++ b/src/runtime/callbacks-test.js
@@ -133,6 +133,48 @@ describes.sandboxed('Callbacks', () => {
     expect(callbacks.hasOffersFlowRequestCallback()).to.be.true;
   });
 
+  describe('warnings', () => {
+    let warnFn;
+
+    beforeEach(() => {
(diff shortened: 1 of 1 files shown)


src/runtime/callbacks-test.js
```diff
@@ -133,6 +133,48 @@ describes.sandboxed('Callbacks', () => {
...
 
+  describe('warnings', () => {
+    let warnFn;
+
+    beforeEach(() => {
+      warnFn = sandbox.spy(self.console, 'warn');
+    });
+
+    afterEach(() => {
+      warnFn.restore();
```
fikus, line 144: I don't think this is needed, since we have a global teardown function that restores all fakes: https://github.com/subscriptions-project/swg-js/blob/main/test/describes.js#L196

fikus (approved):

src/runtime/callbacks-test.js
ChrisAntaki: Nice catch! That is great

PR #2612, opened 2022-12-09
Improve coverage in gaa-test.js

ChrisAntaki:
- Simplifies fakes that never execute


fikus (approved):
The change LGTM. Am I understanding correctly that this is meant to improve coverage of test code? Removing dead code is fine but I don't think we should be concerned with coverage of test code per se.

ChrisAntaki:
That's a great point. I just created a PR to exclude test files from coverage tracking: https://github.com/subscriptions-project/swg-js/pull/2614

PR #2627, opened 2022-12-12
Remove try/catch patterns within unit tests

ChrisAntaki:
- Removes try/catch patterns within unit tests, since they can hide silent failures
- Explicitly expects errors/rejects instead
- Adds eslint rule blocking try/catch within unit tests


ChrisAntaki:
I'll send this for review after #2628

src/runtime/basic-runtime-test.js
```diff
@@ -75,11 +75,7 @@ describes.realWin('installBasicRuntime', (env) => {
     expect(progress).to.equal('');
 
     // Install runtime and schedule few more dependencies.
-    try {
-      installBasicRuntime(win);
-    } catch (e) {
-      // Page doesn't have valid subscription and hence this function throws.
```
ChrisAntaki, line 81: Narrator: It didn't

jpettitt: LOL

jpettitt: But should it have?

ChrisAntaki: Haha, my best guess is no... at least from what's going on in the test and its name

PR #2630, opened 2022-12-12
Use arrow syntax for methods when possible

ChrisAntaki:
- Uses `() => {}` syntax for methods when possible
- Keeps using `function () {}` syntax when the method (or a method it contains) needs to reference a dynamic `this` value


ChrisAntaki:
Waiting for #2629

hartmannr76 (approved):

jpettitt (approved):

PR #2648, opened 2022-12-20
chore: get rid of custom css->js transform

hartmannr76:
Internal bug tracker: b/261875231

We had a custom process to transform CSS->JS in Gulp, which makes it hard to use internal tools. This cleans up our usage to just use tagged [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) so there is no custom transform and also enables us to have syntax highlighting with the use of a `css` tag and the [styled-components extension](https://marketplace.visualstudio.com/items?itemName=styled-components.vscode-styled-components) installed

image


build-system/tasks/compile.js
```diff
@@ -45,9 +42,6 @@ exports.compile = async (options = {}) => {
   mkdirSync('build/fake-module/src');
   mkdirSync('build/css');
 
-  // Compile CSS because we need the css files in compileJs step.
-  await compileCss('./src/', './build/css', options);
-
   // Optionally skip JavaScript compilation.
   if (options.skipJs) {
```
ChrisAntaki, line 46: We could remove this option too. Unit tests would no longer need to call this method

ChrisAntaki (approved):

PR #2653, opened 2022-12-21
Standardized compiler define usage

hartmannr76:
Internal bug tracker: b/261875231

Closure has a built-in ability to do replacements on a variable with the `@define` attribute in JSDoc ([reference](https://github.com/google/closure-compiler/wiki/Annotating-JavaScript-for-the-Closure-Compiler#define-type-description)).

This updates our code to stop using a string replace with variables like `$someVar$` and uses proper constructs.

There was no existing babel plugin to support this from what I could tell (closest was https://github.com/FormidableLabs/babel-plugin-transform-define, but it would only swap the `Identifier` type for a literal which doesn't work the same way.

The custom plugin is very very minimal and meant to add similar support to what we get from closure.

### DIFF
--- a/build-system/tasks/karma.conf.js
+++ b/build-system/tasks/karma.conf.js
@@ -16,7 +16,6 @@
 'use strict';
 
 const argv = require('minimist')(process.argv.slice(2));
-const through = require('through2');
 const {isCiBuild} = require('../ci');
 
 /**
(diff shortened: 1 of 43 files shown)


build-system/tasks/karma.conf.js
```diff
@@ -35,19 +35,26 @@ module.exports = {
...
-        through(function (buf, enc, next) {
+        through2(function (buf, enc, next) {
           this.push(
             buf
               .toString('utf8')
-              // Set Pay environment to indicate we're in a Karma test.
-              .replace(/\$payEnvironment\$/g, 'TEST')
-              // Some tests need a valid SwG server origin.
-              .replace(/\$frontend\$/g, 'https://news.google.com')
+              /**
```
hartmannr76, line 43: @ChrisAntaki so this is unfortunate. Apparently theres a bunch of plugin that were reporting issues with istanbul and they made hacks to get around it with no response on the issues for the past few years

ChrisAntaki: Years? Oof

ChrisAntaki (approved):

jpettitt (approved):

PR #2658, opened 2022-12-24
:recycle: chore(callbacks): add entitlement to unlockArticle()

hjanuschka:
Hey there 🙌,

great project, keep it up and take a 🍺  or ☕ 

we ran into the problem that we'd needed to get the entitlements, for server-side validation - in the "unlock" scenario for EA.

a workaround -  that doesn't feel future proof, that we came up with is:

```js
    function unlockArticle() {
      callSwg((subscriptions) => {
        subscriptions.getEntitlements().then((entitlements) => {
          console.log("Consumed Entitlements", entitlements);
        });
      });
    }
```

after the PR - it would be easier:

```js
    function unlockArticle(entitlements) {
      if(entitlements) {
          console.log("Consumed Entitlements", entitlements);
      }
   }
```

please let me know if you'd like to proceed with that PR and if you'd want me to change anything.

regards
helmut


ChrisAntaki:
Hey, thanks for suggesting this! I'll discuss it more with the team, then get back to you. This might take a few weeks due to the holidays. ❄

jpettitt:
+1 this looks like a good suggestion.

teiswit:
Hi @hjanuschka are you implementing a server-side or a client-side paywall solution of our Extended Access API? If you are implementing a server-side solution you would already be getting the entitlement object from the Showcase API on the server-side which you can validate there directly. To help us understand the problem you are trying to solve could you elaborate on the use case here? Thanks!

hjanuschka:
i know - i read about the server-side implemantation. but the proposed solutions all require to generate the entitlement on the server side. i need/want to do the fetching of the entitlement client side (static pages, caching and many other reasons) - and only validate the entitlement on the server side. swg generates a entitlement anyway so i don't see the benefits/hard-requirement of server-side-generated entitlement fetching.

ChrisAntaki (approved):

Asesjix:
+1 We do a client site check but we need a call in the backend afterwards to approve the user for the article. We would like to verify this call. For this it would be very helpful to have the entitlement.

ChrisAntaki:
This change should go live within 48 hours. Thank you @hjanuschka for the PR and @Asesjix for the feedback!

PR #2662, opened 2023-01-03
Simplify custom cancellation errors

ChrisAntaki:
- Creates `Error` with a modified `name` and a prefixed `message`
- Fixes Closure warning: JSC_CONFORMANCE_VIOLATION


src/utils/errors.js
```diff
@@ -14,30 +14,26 @@
...
-  isAbortError,
-} = require('web-activities/activity-ports');
-
 /**
  * Whether the specified error is an AbortError type.
  * See https://heycam.github.io/webidl/#aborterror.
  * @param {*} error
  * @return {boolean}
  */
 export function isCancelError(error) {
```
fikus, line 23: Could this function be called on an error generated by web-activities (in which case the logic would not be correct)? I'm not sure this simplification is worth it because it could lead to unexpected behavior, since there are now diverging implementations of this functionality here and in web-activities.

ChrisAntaki: Good news, it works with `web-activities` abort errors. I just added a test verifying this.
https://github.com/subscriptions-project/swg-js/pull/2662/files#diff-08d49f0abfc63df34bf11282639d901e155eda0210bf852990daeffa102be431R28

src/utils/errors.js
```diff
@@ -14,30 +14,26 @@
...
-  return isAbortError(error);
+  return error?.name === 'AbortError';
 }
 
 /**
- * Creates or emulates a DOMException of AbortError type.
+ * Creates an Error of AbortError type.
  * See https://heycam.github.io/webidl/#aborterror.
- * @param {!Window} win
  * @param {string=} message
```
fikus, line 30: nit (existing issue): the type should just be `string` since there is no default value.

fikus (approved):

src/utils/errors.js
fikus: Oh, OK, cool.

src/utils/errors.js
ChrisAntaki: Nice catch, updated

jpettitt (approved):

PR #2668, opened 2023-01-04
When experiment SurveyTriggeringPriority from article endpoint is enabled, trigger Survey prompt before Contribution prompt.

justinchou-google:
b/263402383

This implements the logic to serve a Survey prompt before a Contribution prompt given the SurveyTriggeringPriorityExperiment is enabled and passed from the article endpoint. The Survey prompt will be served first, even if there are other Audience Actions configured with a higher priority.

### DIFF
--- a/src/runtime/auto-prompt-manager-test.js
+++ b/src/runtime/auto-prompt-manager-test.js
@@ -1273,7 +1273,11 @@ describes.realWin('AutoPromptManager', (env) => {
       expect(autoPromptManager.promptDisplayed_).to.equal(
         'TYPE_REWARDED_SURVEY'
       );
-      await verifyOnCancelStores('contribution,TYPE_REWARDED_SURVEY');
+      await verifyOnCancelStores(
+        storageMock,
+        actionFlowSpy,
+        'contribution,TYPE_REWARDED_SURVEY'
+      );
     });
 
     it('should show the second Audience Action flow if the first was previously dismissed and is not the next Contribution prompt time', async () => {
@@ -1306,6 +1310,8 @@ describes.realWin('AutoPromptManager', (env) => {
         'TYPE_REGISTRATION_WALL'
       );
       await verifyOnCancelStores(
+        storageMock,
+        actionFlowSpy,
         'contribution,TYPE_REWARDED_SURVEY,TYPE_REGISTRATION_WALL'
       );
     });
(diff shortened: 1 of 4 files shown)


src/runtime/auto-prompt-manager-test.js
```diff
@@ -1526,24 +1548,303 @@ describes.realWin('AutoPromptManager', (env) => {
...
-    async function verifyOnCancelStores(setValue) {
-      storageMock
-        .expects('set')
-        .withExactArgs(
-          StorageKeys.DISMISSED_PROMPTS,
-          setValue,
-          /* useLocalStorage */ true
-        )
-        .resolves(null)
+  describe('Contribution Flows with Audience Actions and Experiments', () => {
```
justinchou-google, line 1559: One comment on these tests: these are almost entirely redundant of the original `Contribution Flows with Audience Actions` test, with some slight modifications -- primarily this suite checks for the main case (where the survey prompt should be triggered before the contribution prompt when both are valid to show), and the other tests assert the behavior is the same as with the experiment disabled in all other cases. 

There may be a more concise way to test this with parameters, but I think there is an argument for keeping an explicit suite for Experiments, which can easily be added to / ripped out as we launch/turn down experiments.

src/runtime/auto-prompt-manager.js
```diff
@@ -673,4 +691,22 @@ export class AutoPromptManager {
...
   }
+
+  /**
+   * Checks if provided ExperimentFlag is returned in article endpoint.
+   * @param {?./entitlements-manager.Article|undefined} article
+   * @param {string} experimentFlag
+   * @return {!Promise<boolean>}
+   */
+  async isExperimentEnabled_(article, experimentFlag) {
+    if (article === undefined) {
```
ChrisAntaki, line 702: Just curious, could we simplify this to return `false` if the `article` param is falsy?

src/runtime/auto-prompt-manager.js
```diff
@@ -419,14 +419,32 @@ export class AutoPromptManager {
       autoPromptType === AutoPromptType.CONTRIBUTION ||
       autoPromptType === AutoPromptType.CONTRIBUTION_LARGE
     ) {
-      if (!dismissedPrompts) {
+      const surveyTriggeringPriorityEnabled = await this.isExperimentEnabled_(
```
ChrisAntaki, line 422: I'm not quite clear what `surveyTriggeringPriorityEnabled` means. Could you describe it a bit?

justinchou-google: `surveyTriggeringPriority` refers to the experiment, perhaps it will be more clear if I name it as such? `surveyTrigginerPriorityExperimentIsEnabled`

ChrisAntaki: Would it make sense to call it `preferSurveyOverContributionPrompt`?

ChrisAntaki (approved):
Looks good!

PR #2676, opened 2023-01-06
Simplify loading of custom `eslint` rules to avoid stale cache issues

ChrisAntaki:
- Tells `eslint` about folder of custom rules, instead of treating them like an `eslint` plugin.
- Stops treating the folder like an `npm` dependency, since that can lead to stale cache issues if the rules are updated.


ChrisAntaki:
Waiting for #2675 to merge

hartmannr76 (approved):

PR #2681, opened 2023-01-09
Improve coverage in auto-prompt-manager.js

ChrisAntaki:
- Removes unreachable code

### DIFF
--- a/src/runtime/auto-prompt-manager.js
+++ b/src/runtime/auto-prompt-manager.js
@@ -695,15 +695,11 @@ export class AutoPromptManager {
 
   /**
    * Checks if provided ExperimentFlag is returned in article endpoint.
-   * @param {?./entitlements-manager.Article|undefined} article
+   * @param {!./entitlements-manager.Article} article
    * @param {string} experimentFlag
    * @return {!Promise<boolean>}
    */
   async isExperimentEnabled_(article, experimentFlag) {
-    if (!article) {
-      return false;
-    }
-
     const articleExpFlags =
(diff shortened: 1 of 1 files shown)


src/runtime/auto-prompt-manager.js
```diff
@@ -421,7 +421,7 @@ export class AutoPromptManager {
     ) {
       const preferSurveyOverContributionPrompt =
         await this.isExperimentEnabled_(
-          article,
+          /** @type {!./entitlements-manager.Article} */ (article),
```
fikus, line 424: How do you know article is defined here? It's treated as possibly undefined elsewhere in this method.

ChrisAntaki: The `getAudienceActionPromptType_` method ends on [line 410](
https://github.com/subscriptions-project/swg-js/blob/6e75b712b60354cb7b9c060305140cec200f10fc/src/runtime/auto-prompt-manager.js#L410) if the `article` param is falsy

fikus: Can we explicitly return undefined earlier if article is undefined?

ChrisAntaki: Sure, I updated the code to return `undefined` earlier if possible (cc45139d1513d73c2d2a9cce28692ab68185ba51). This required updating some tests (5174ba3ffeb8cebc588975e2a3f40e3e9f86644e) too

fikus (approved):

PR #2691, opened 2023-01-12
Remove linkSubscription from basic-runtime

fikus:
This method should only be invoked using the original runtime. See b/261874155.


ChrisAntaki (approved):

fikus:
Hi @jpettitt , can you please review for the api change?

hartmannr76 (approved):

PR #2696, opened 2023-01-13
Improve coverage in propensity.js

ChrisAntaki:
- Add a test
- Add to an existing test


src/runtime/propensity-test.js
```diff
@@ -182,6 +201,14 @@ describes.realWin('Propensity', (env) => {
       })
     ).to.throw('Event data must be an Object(all_offers)');
 
+    //ensure it rejects invalid `active` values
```
fikus, line 204: nits: Add space after //, use sentence case, end with period.

fikus (approved):

src/runtime/propensity-test.js
ChrisAntaki: Ah yes, that's more readable

PR #2707, opened 2023-01-18
Improve Extended Access CSS with syntax highlighting

ChrisAntaki:
- Improves Extended Access CSS 
  - Adds syntax highlight
  - Removes an extra `!important` statement (which VS Code flagged, after adding the syntax highlighting)
  - Formatting (which VS Code automatically applied, after adding the syntax highlighting)
- Fixes a typo in ui-css.js file

image

### DIFF
--- a/src/runtime/extended-access/html-templates.js
+++ b/src/runtime/extended-access/html-templates.js
@@ -54,14 +54,14 @@ export const REGISTRATION_WIDGET_IFRAME_HTML = `
 `;
 
 /**
- * HTML for the metering regwall dialog, where users can sign in with Google.
- * The script creates a dialog based on this HTML.
- *
- * The HTML includes an iframe that loads the Google Sign-In button.
- * This iframe can live on a different origin.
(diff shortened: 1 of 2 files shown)


hartmannr76 (approved):

src/runtime/extended-access/html-templates.js
```diff
@@ -190,6 +194,18 @@ export const REGWALL_HTML = `
...
+`;
+
+/**
+ * HTML for the metering regwall dialog, where users can sign in with Google.
+ * The script creates a dialog based on this HTML.
+ *
+ * The HTML includes an iframe that loads the Google Sign-In button.
+ * This iframe can live on a different origin.
+ */
+export const REGWALL_HTML = `
```
fikus, line 206: Would it also be beneficial to tag the HTML strings in this file as html?

ChrisAntaki: Looks like `styled-components` just covers CSS syntax highlighting

PR #2715, opened 2023-01-24
CodeCov: Fail check if patch has less than 99% coverage on changed lines

ChrisAntaki:
- Configures CodeCov check to fail if a patch's changed lines have less than 99% coverage
- Excludes Storybook stories from coverage tracking


fikus:
This threshold was intentionally relaxed in #2236 because this was being triggered too often. 0.05% is already a very low threshold and I don't think we need to lower it any further.

fikus (approved):

PR #2718, opened 2023-01-26
Update graypane.js

mhkawano:
Remove `pointer-events': 'none'`.

Issue found in b/266469160. The scrim/graypane does not block clicks on links in the article, which is an issue(?) for regwalls/newsletters.


mhkawano:
Working as intended.

PR #2731, opened 2023-01-30
Remove non-delayed redirection in 3P sign-in

teiswit:
A delay was added in #2380 to allow a transmission of analytics events to survive a redirection, however it seems that the original redirection (without a delay) which is supposed to be replaced has not been removed.

https://github.com/subscriptions-project/swg-js/pull/2380/files#diff-33c69a3765168e37e59b22dfbc8440c647eb980a87d5716e0d3fa3d664e6d84eR1391


teiswit:
Looks like this is intentional as the behavior is described in the unit test: 'it should open window twice when redirectMode is true'

ChrisAntaki:
Hey! Your initial comment makes sense. I just added that test after looking at the code and seeing the delayed redirect wasn't covered by tests. Please feel free to modify the test(s) as needed

teiswit:
Got it - I've removed and modify the test case to capture the correct behavior of opening a window once (but delayed).

PR #2732, opened 2023-01-30
E2E test new build system

ChrisAntaki:
- Add new CI task
- Preserve existing E2E CI task with the same name, since the name is part of the `main` branch's rules
- Remove Closure build step from `e2e` task

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -20,7 +20,7 @@ jobs:
       - name: Check
         run: gulp check
 
-  e2e-with-closure:
+  e2e:
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -34,6 +34,29 @@ jobs:
       - name: Install dependencies
         run: yarn
 
+      - name: Compile JS with Closure
+        run: gulp dist
+
+      - name: E2E Tests
+        run: gulp e2e
+
+  e2e-with-new-build-system:
```
fikus, line 43: Would `e2e-vite` be a little more clear?

fikus (approved):

.github/workflows/ci.yaml
ChrisAntaki: Yes, that's cleaner 🧼

PR #2736, opened 2023-02-01
Use NPM instead of Yarn

ChrisAntaki:
- Switches to the default Nodejs package manager


.github/workflows/ci.yaml
```diff
@@ -15,7 +15,7 @@ jobs:
           node-version: '18.x'
 
       - name: Install dependencies
-        run: yarn --ignore-optional
+        run: npm i --omit optional
```
hartmannr76, line 18: Should we be using things like `npm ci` instead? It enforces not overwriting the lock file so changes need to be updated in there

ChrisAntaki: Updated!

hartmannr76 (approved):

PR #2744, opened 2023-02-07
DO NOT MERGE: Lazy load gpay instead of baked in

hartmannr76:
FYI @fikus @ChrisAntaki @jpettitt 

This is meant more of a demo for what lazily loading the pay.js script would look like.

Essentially:
1) Load the script asynchronously as part of our script loading
2) We need to have hooks in place to let us know when it is ready, hence the use of the Promise

Results:
Before: `dist/basic-subscriptions.js  199.02 kB │ gzip: 59.10 kB`
After: `dist/basic-subscriptions.js  163.02 kB │ gzip: 48.90 kB`

Thats an 18% decrease in raw bundle size and a 17.26% decrease in zipped size. Thats fairly significant, plus it removes tech debt from our end on needing to maintain a copy of it

Complications:
If we want to just directly use it the way we currently do, the web-activities code in pay.js needs to expose its public methods since they are getting renamed at the moment
image

### DIFF
--- a/src/runtime/pay-client.js
+++ b/src/runtime/pay-client.js
@@ -15,7 +15,7 @@
  */
 
 import {ExperimentFlags} from './experiment-flags';
-import {PaymentsAsyncClient} from '../../third_party/gpay/src/payjs_async';
+// import {PaymentsAsyncClient} from '../../third_party/gpay/src/payjs_async';
 import {Preconnect} from '../utils/preconnect';
 import {StorageKeys} from '../utils/constants';
 import {bytesToString, stringToBytes} from '../utils/bytes';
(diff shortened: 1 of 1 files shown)


src/runtime/pay-client.js
```diff
@@ -104,11 +115,13 @@ export class PayClient {
...
    * @private
    */
-  createClient_(options, googleTransactionId, handler) {
+  async createClient_(options, googleTransactionId, handler) {
+    await PayReadyPromise;
     // Assign Google Transaction ID to PaymentsAsyncClient.googleTransactionId_
     // so it can be passed to gpay_async.js and stored in payment clearcut log.
-    PaymentsAsyncClient.googleTransactionId_ = googleTransactionId;
-    return new PaymentsAsyncClient(
+    google.payments.api.PaymentsAsyncClient.googleTransactionId_ =
```
ChrisAntaki, line 122: We might need to quote the `googleTransactionId_` property name, so Terser doesn't mangle it in the new Vite build system

PR #2750, opened 2023-02-09
When SecondPromptDelay experiment is enabled, allow default 2 free reads between first and second prompts

justinchou-google:
b/267650049

For SecondPromptDelay experiment, grant the user "free reads" (default 2 reads) and do not show the next prompt after the first prompt. Stores timestamp of initial and free reads into local storage to determine if the second prompt should be delayed/suppressed.

Refactoring of `showAutoPrompt_` to better isolate when to suppress prompt in https://github.com/subscriptions-project/swg-js/pull/2751


src/runtime/experiment-constants.js
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * @enum {number}
+ */
+export const ExperimentConstants = {
+  /**
+   * Number of free reads between prompts for Second Prompt Delay Experiment.
+   * For X reads, we expect prompt sequence to be: [Prompt 1, Read 1, ..., Read X, Prompt 2]
+   */
+  SECOND_PROMPT_DELAY_BY_NUMBER_OF_READS_DEFUALT: 2,
```
ChrisAntaki, line 25: Would `FREE_READS_PER_PROMPT` work as a name?

src/utils/constants.js
```diff
@@ -69,6 +69,12 @@ StorageKeys.IS_READY_TO_PAY = 'isreadytopay';
  */
 StorageKeys.REDIRECT = 'subscribe.google.com:rk';
 
+/**
+ * Local storage key for should show autoprompt timestamps.
+ * const {string}
+ */
+StorageKeys.SHOULD_SHOW_AUTOPROMPT = 'shouldshowautopromptexp';
```
oyj9109, line 76: I feel like the key name does not represent well what it stores. I thought something like 'free reads consumed' but realized we also want to store timestamp for the first prompt. 
Just an optional suggestion, what about "metered for second prompt"? Might be too specific to the experiment so we may want to come up with a more general one later...

src/runtime/auto-prompt-manager.js
```diff
@@ -700,6 +721,41 @@ export class AutoPromptManager {
...
+   * number of free reads X = 2, then
+   * Timestamps   Show Autoprompt   Store Timestamp
+   * []           YES (1st prompt)  YES
+   * [t1]         NO  (free read)   YES
+   * [t1, t2]     NO  (free read)   YES
+   * [t1, t2, t3] YES (2nd prompt)  NO
+   * @param {number} numFreeReads
+   * @return {!Promise<boolean>}
+   */
+  async checkFreeReadShouldSuppressAutoprompt_(numFreeReads) {
```
ChrisAntaki, line 738: I noticed this method is only used in one place, with the same argument every time (2). I'm wondering if we could simplify the code a bit. What if we...
- Removed `src/runtime/experiment-constants.js` and hardcoded the `2` in the method
- Named the method `shouldDelayPromptForSecondPromptDelayExperiment()` and removed the `numFreeReads` param

What do you think?

src/utils/constants.js
ChrisAntaki, line 76: Nit: This name seems more accurate to me, since the code is saving this event even when the prompt is suppressed
```suggestion
StorageKeys.SECOND_PROMPT_DELAY_COUNTER= 'secondpromptdelaycounter';
```

ChrisAntaki (changes requested):
Just a few comments

src/runtime/auto-prompt-manager.js
justinchou-google: SGTM, I thought it may be better practice to keep the constants separately, but I am also fine with minimizing the number of unnecessary files.

oyj9109 (approved):
One thing to note here is, it's designed for new visitors (starting from 1st prompt), however there are users who have already visited and may have seen 1st prompt, but are put in the experiment as if they are first visitor (timestamps are collected from n-th prompt). We will need to handle the transitions between before/after this change if that could lead to an unexpected triggering behavior. 
Fortunately, it doesn't seem like a big issue for now as we're going to selectively run the experiment for one publisher who have only 2 types of prompt and the possible issues I can think of is, the experiment doesn't take effect if 1st prompt was already shown.
That being said, the approach looks good to me. We will need to think about how we can distinguish it in the experiment data though.

src/utils/constants.js
justinchou-google: `SECOND_PROMPT_DELAY_COUNTER` was suggested, what do you think?

src/runtime/auto-prompt-manager.js
```diff
@@ -700,6 +721,41 @@ export class AutoPromptManager {
...
+      shouldShowAutopromptTimestamps.length <= numFreeReads;
+    const shouldStoreTimestamp =
+      shouldShowAutopromptTimestamps.length <= numFreeReads;
+
+    if (shouldStoreTimestamp) {
+      this.storage_.storeEvent(StorageKeys.SHOULD_SHOW_AUTOPROMPT);
+    }
+
+    if (shouldSuppressPrompt) {
+      this.promptDisplayed_ = null;
```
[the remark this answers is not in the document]
justinchou-google: good point, moved to `showAutoPrompt_`.

src/utils/constants.js
oyj9109: Just saw chris's suggestion earlier. That looks good to me too :)

justinchou-google: > One thing to note here is, it's designed for new visitors (starting from 1st prompt), however there are users who have already visited and may have seen 1st prompt, but are put in the experiment as if they are first visitor (timestamps are collected from n-th prompt). We will need to handle the transitions between before/after this change if that could lead to an unexpected triggering behavior. Fortunately, it doesn't seem like a big issue for now as we're going to selectively run the experiment for one publisher who have only 2 types of prompt and the possible issues I can think of is, the experiment doesn't take effect if 1st prompt was already shown. That being said, the approach looks good to me. We will need to think about how we can distinguish it in the experiment data though.

Good call out, perhaps a solution is to add extra checks to verify a user is a "new visitor" (at least to our best ability with anonymous readers), such as checking no other prompts were previously dismissed.

justinchou-google:
> One thing to note here is, it's designed for new visitors (starting from 1st prompt), however there are users who have already visited and may have seen 1st prompt, but are put in the experiment as if they are first visitor (timestamps are collected from n-th prompt). We will need to handle the transitions between before/after this change if that could lead to an unexpected triggering behavior. Fortunately, it doesn't seem like a big issue for now as we're going to selectively run the experiment for one publisher who have only 2 types of prompt and the possible issues I can think of is, the experiment doesn't take effect if 1st prompt was already shown. That being said, the approach looks good to me. We will need to think about how we can distinguish it in the experiment data though.

Good call out, perhaps a solution is to add extra checks to verify a user is a "new visitor" (at least to our best ability with anonymous readers), such as checking no other prompts were previously dismissed.

justinchou-google:
> Just a few comments

Thank you for the review!

src/utils/constants.js
```diff
@@ -69,6 +69,12 @@ StorageKeys.IS_READY_TO_PAY = 'isreadytopay';
  */
 StorageKeys.REDIRECT = 'subscribe.google.com:rk';
 
+/**
+ * Local storage key for should show autoprompt timestamps.
```
ChrisAntaki, line 73:
```suggestion
 * Local storage key for tracking when the SecondPromptDelay experiment should suppress prompts.
```

justinchou-google: oops, fixed.

ChrisAntaki (approved):

PR #2757, opened 2023-02-13
Fix Subscription tests for SecondPromptDelay Experiment

justinchou-google:
Tests were intermittently failing after https://github.com/subscriptions-project/swg-js/pull/2750, fixes test to use `AutoPromptType.SUBSCRIPTION`.

### DIFF
--- a/src/runtime/auto-prompt-manager-test.js
+++ b/src/runtime/auto-prompt-manager-test.js
@@ -1907,7 +1907,7 @@ describes.realWin('AutoPromptManager', (env) => {
         .once();
     });
 
-    it('should follow AudienceActionFlow without SecondPromptDelay experiment for Subscriptions', async () => {
+    it('should not delay second prompt for Subscriptions', async () => {
       await autoPromptManager.showAutoPrompt({
         autoPromptType: AutoPromptType.SUBSCRIPTION,
         alwaysShow: false,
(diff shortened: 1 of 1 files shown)


src/runtime/auto-prompt-manager-test.js
```diff
@@ -1909,20 +1909,16 @@ describes.realWin('AutoPromptManager', (env) => {
...
-      expect(startSpy).to.have.been.calledOnce;
-      expect(actionFlowSpy).to.have.been.calledWith(deps, {
-        action: 'TYPE_REWARDED_SURVEY',
-        onCancel: sandbox.match.any,
-        autoPromptType: AutoPromptType.SUBSCRIPTION_LARGE,
-      });
+      expect(startSpy).to.not.have.been.called;
       expect(alternatePromptSpy).to.not.have.been.called;
-      expect(autoPromptManager.getLastAudienceActionFlow()).to.not.equal(null);
+      expect(autoPromptManager.getLastAudienceActionFlow()).to.equal(null);
```
ChrisAntaki, line 1921: This might be a dumb question, but why is switching to `AutoPromptType.SUBSCRIPTION` causing this change in behavior?

justinchou-google: This should have been `.to.equal(null)` originally, not sure why it wasn't failing before :)

ChrisAntaki (approved):

PR #2763, opened 2023-02-15
Set num_impressions_between_prompts from client config SecondPromptDelayExperiment

justinchou-google:
b/268491550 
Read `num_impressions_between_prompts` from autoprompmanager client config and use that to determine number of impressions inbetween triggering of prompt for SecondPromptDelayExperiment. 
Currently, autoPromptConfig always sets value `2`.


src/runtime/auto-prompt-manager.js
```diff
@@ -730,18 +732,22 @@ export class AutoPromptManager {
...
    * @return {!Promise<boolean>}
    */
-  async secondPromptDelayExperimentSuppressesPrompt_() {
-    const numFreeReads = 2; // (b/267650049) 2 free reads
-    const shouldShowAutopromptTimestamps = await this.storage_.getEvent(
+  async secondPromptDelayExperimentSuppressesPrompt_(expConfigNumFreeReads) {
+    const numFreeReads =
+      expConfigNumFreeReads && expConfigNumFreeReads > 0
+        ? expConfigNumFreeReads
+        : 2; // (b/267650049) default 2 free reads
```
justinchou-google, line 742: Last PR, we explicitly removed this param to simplify this function, but here we need to read this from `article`.

src/runtime/auto-prompt-manager.js
```diff
@@ -730,18 +732,22 @@ export class AutoPromptManager {
    * [t1]         NO  (free read)   YES
    * [t1, t2]     NO  (free read)   YES
    * [t1, t2, t3] YES (2nd prompt)  NO
+   * @param {number|undefined} expConfigNumFreeReads
    * @return {!Promise<boolean>}
    */
-  async secondPromptDelayExperimentSuppressesPrompt_() {
-    const numFreeReads = 2; // (b/267650049) 2 free reads
-    const shouldShowAutopromptTimestamps = await this.storage_.getEvent(
+  async secondPromptDelayExperimentSuppressesPrompt_(expConfigNumFreeReads) {
```
ChrisAntaki, line 738: If we set a default number here, would we need the new logic on lines 739-742?
```suggestion
  async secondPromptDelayExperimentSuppressesPrompt_(numFreeReads  = 2) {
```

src/runtime/auto-prompt-manager-test.js
```diff
@@ -1894,20 +1894,14 @@ describes.realWin('AutoPromptManager', (env) => {
         .resolves(entitlements)
         .once();
       articleExpectation = entitlementsManagerMock.expects('getArticle');
```
ChrisAntaki, line 1896: Could we rename this to `getArticleExpectation`?

ChrisAntaki: I think this would make the connection to the `getArticle` method clearer

src/runtime/auto-prompt-manager-test.js
```diff
@@ -2162,6 +2193,32 @@ describes.realWin('AutoPromptManager', (env) => {
     await tick(2);
   }
 
+  function resolveArticleExpectationOnce(
```
ChrisAntaki, line 2196: Could we call the method `mockGetArticleResponse`? I think that would more clearly communicate the method's outcome
```suggestion
  function mockGetArticleResponse(
```

src/runtime/auto-prompt-manager-test.js
```diff
@@ -2162,6 +2193,32 @@ describes.realWin('AutoPromptManager', (env) => {
...
+    articleExpectation,
+    actionTypes,
+    experimentFlags,
+    numReadsBetweenPrompts = 0
+  ) {
+    const actions =
+      actionTypes.map((actionType) => {
```
ChrisAntaki, line 2203: Nit: This could be a one liner
```suggestion
      actionTypes.map((actionType) => ({ type: actionType })
```

src/runtime/auto-prompt-manager.js
justinchou-google: So the server-default value passed in from `experimentConfig` is `0` in the case where the `numFreeReads` is simply initialized, which we do not want to honor over the default of `2` free reads. As such, the server-default `0` would take precedence, which we don't want. Perhaps ideally we would want the default `2` to be passed down from the server.

ChrisAntaki: The suggested change would cover the `undefined` value. If the server is sending `2` then I don't think we need to cover the `0` case?

justinchou-google: Will add this suggested change.

src/runtime/auto-prompt-manager.js
justinchou-google: see https://github.com/subscriptions-project/swg-js/pull/2763#discussion_r1119167308

ChrisAntaki:
Looks good overall! I'll just wait until the team further discusses the Article object design before approving

justinchou-google:
Updated b/268491550, when this feature is available outside of this experiment, we will want the logic to be on the server side. For now, `experimentConfig` can serve this parameter, but we will enforce that this only be used for this experiment.

hartmannr76 (approved):

ChrisAntaki (approved):

PR #2767, opened 2023-02-16
Optimizing license comments

ChrisAntaki:
- Removes Google license comments from bundles, while preserving licenses from others (ex: Math.uuid.js), using an inline Vite plugin

## Bundle size diffs (gzipped)
| Bundle | Before | After | Diff |
| :--- | :---: | :---: | :---: |
| Swgjs Basic | 59,160 B | 58,620 B | -540 B |
| Swgjs Classic | 51,254 B | 50,753 B | -501 B |

### DIFF
--- a/vite.config.js
+++ b/vite.config.js
@@ -39,6 +39,23 @@ const plugins = [
     preventAssignment: false,
     values: replacementValues,
   }),
+  // Remove Google license comments.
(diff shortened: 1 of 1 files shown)


ChrisAntaki:
TODO: Add _one_ license for Google

PR #2775, opened 2023-02-21
Add useArticleEndpoint to config

mhkawano:
~~@hartmannr76 
We talked about using `subscriptions.configure` for setting experiments on this flag, but it appears that `configure` will run after the endpoint is already decided in `runtime.configured_`.  I think this means that we will need to use an experiment that's either compiled in, use a value in the markup, or add a new method to the runtime. I opted for the last option, and was wondering what you think. With this implementation you will need to add the following to the swg callback: `subscriptions.setExperimentValue('enable_enterprise_article', true);`.~~

Add new config field to set the use of the article endpoint. Needed as part of efforts to open audience actions for enterprise users.

### DIFF
--- a/src/runtime/runtime.js
+++ b/src/runtime/runtime.js
@@ -231,10 +231,7 @@ export class Runtime {
       pageConfig,
       /* integr */ {
         configPromise: this.configuredRuntimePromise_,
-        useArticleEndpoint: isExperimentOn(
-          this.win_,
-          ExperimentFlags.ENABLE_ENTERPRISE_ARTICLE
-        ),
+        useArticleEndpoint: this.config_.useArticleEndpoint || false,
       },
       this.config_
     );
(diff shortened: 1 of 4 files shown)


src/runtime/runtime.js
```diff
@@ -817,6 +820,11 @@ export class ConfiguredRuntime {
             error = 'publisherProvidedId must be a string, value: ' + value;
           }
           break;
+        case 'useArticleEndpoint':
+          if (!isBoolean(value)) {
+            error = 'Unknown useArticleEndpoint value: ' + value;
```
ChrisAntaki, line 825:
```suggestion
            error = 'useArticleEndpoint must be a boolean, type: ' + typeof value;
```

ChrisAntaki (changes requested):
Looks good overall! I just suggested a different error message

src/runtime/runtime.js
ChrisAntaki: This seems clearer, in case someone passes "true", like in your test case

mhkawano: Done. I wanted to keep the message consistent with other bool params, so I'll update the others.

ChrisAntaki (approved):

hartmannr76 (approved):

PR #2777, opened 2023-02-22
Call unlockArticle if provided when consuming showcaseEntitlement

teiswit:
Previously (#2771) we made `unlockArticle` optional but not really calling it in a server-side paywall scenario (i.e. when showcaseEntitlement is provided). This change will make `unlockArticle` effective when provided.

- Publishers who unlock articles on the server-side will not be impacted by this change. GAA will initialize a default `unlockArticle: () => {}`.
- Publishers who provide `unlockArticle` will have this method called when `showcaseEntitlement` is consumed.
 
Ref: b/234314583#5

### DIFF
--- a/src/runtime/extended-access/gaa-metering.js
+++ b/src/runtime/extended-access/gaa-metering.js
@@ -103,12 +103,6 @@ export class GaaMetering {
       rawJwt,
     } = params;
 
-    // Make unlockArticle optional when showcaseEntilement is provided.
-    const unlockArticle =
-      showcaseEntitlement && !params.unlockArticle
-        ? () => {}
-        : params.unlockArticle;
-
     // Set class variables
     GaaMetering.userState = userState;
     GaaMetering.publisherEntitlementPromise = publisherEntitlementPromise;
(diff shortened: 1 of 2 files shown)


src/runtime/extended-access/gaa-metering.js
```diff
@@ -145,7 +145,11 @@ export class GaaMetering {
         unlockArticleIfGranted();
       } else if (showcaseEntitlement) {
         debugLog(showcaseEntitlement);
-        subscriptions.consumeShowcaseEntitlementJwt(showcaseEntitlement);
+        subscriptions.consumeShowcaseEntitlementJwt(showcaseEntitlement, () => {
```
ChrisAntaki, line 148: Could you move the definition of the `unlockArticle` variable here? The new code below looks like the first usage of the variable

ChrisAntaki (changes requested):
Looking good, I just suggested moving a variable definition to improve clarity

src/runtime/extended-access/gaa-metering.js
teiswit: It's actually being used prior to this line in a `unlockArticleIfGranted`, so I've moved the definition of `unlockArticle` closest to where it's first used.

teiswit:
Please take another look. Thanks!

src/runtime/extended-access/gaa-metering.js
ChrisAntaki: Thanks! And sorry for missing that

ChrisAntaki (approved):

PR #2779, opened 2023-02-22
Funnel Prompts for Enterprise API

mhkawano:
This is the proposed API changes for making funnel prompts available to enterprise publishers.


src/api/subscriptions.js
```diff
@@ -351,9 +354,61 @@ export class Subscriptions {
    * @return {?}
    */
   setPublisherProvidedId(publisherProvidedId) {}
+
+  /**
+   * Starts the intervention audience action flow.
```
ChrisAntaki, line 359: Could this say more about what interventions are, and when publishers might want to call the method?

mhkawano: Done.

ChrisAntaki: Thanks!

ChrisAntaki (approved):

PR #2782, opened 2023-02-27
Optimization: Avoid unnecessary iframe redirects

ChrisAntaki:
- Removes unnecessary "_/" from iframe URLs. This used to be necessary, but now it just causes a redirect, which slows down UX.

### DIFF
--- a/src/runtime/services.js
+++ b/src/runtime/services.js
@@ -156,7 +156,7 @@ export function feUrl(
       ? `swg/${prefix}`
       : `${prefix}/swg`
     : 'swg';
-  url = feCached(`${getSwgMode().frontEnd}/${prefixed}/_/ui/v1${url}`);
+  url = feCached(`${getSwgMode().frontEnd}/${prefixed}/ui/v1${url}`);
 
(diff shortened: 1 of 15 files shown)


src/runtime/services.js
```diff
@@ -156,7 +156,7 @@ export function feUrl(
       ? `swg/${prefix}`
       : `${prefix}/swg`
     : 'swg';
-  url = feCached(`${getSwgMode().frontEnd}/${prefixed}/_/ui/v1${url}`);
+  url = feCached(`${getSwgMode().frontEnd}/${prefixed}/ui/v1${url}`);
```
hartmannr76, line 159: Wanna just collapse 154-159 since its no longer behind a flag?

ChrisAntaki: Is it cool if I tackle this in a followup PR?

hartmannr76: sgtm

hartmannr76 (approved):

alin04 (approved):

PR #2788, opened 2023-03-01
Move autoprompt code into autopromptmanager file

kristenwang:
b/270947002, we will use value from audienceActions to determine the autoprompt type, therefore the displayLargePromptFn should also be defined based on audienceActions. So moving this code into autopromptmanager file.


src/runtime/basic-runtime.js
```diff
@@ -367,7 +366,10 @@ export class ConfiguredBasicRuntime {
     }
 
     /** @private @const {!AutoPromptManager} */
-    this.autoPromptManager_ = new AutoPromptManager(this);
+    this.autoPromptManager_ = new AutoPromptManager(
+      this,
+      Promise.resolve(this.configuredClassicRuntime_)
```
ChrisAntaki, line 371: Just curious, why does this need to be a promise?

kristenwang: Remove the promise. I initially used it cuz the buttonApi is using it.

ChrisAntaki (approved):

PR #2792, opened 2023-03-02
Get value from audienceActions if autoPromptType is undefined

kristenwang:
b/270947002, If autoPromptType is undefined in the snippet, we will set it to be the autoPromptType gotten from audienceActions.


src/runtime/auto-prompt-manager.js
```diff
@@ -187,6 +189,15 @@ export class AutoPromptManager {
     dismissedPrompts,
     params
   ) {
+    const audienceActionsPromptType = article
+      ? this.getAutoPromptType_(article)
```
ChrisAntaki, line 193: Could you pass the `article.audienceActions` directly here?

src/runtime/auto-prompt-manager.js
```diff
@@ -187,6 +189,15 @@ export class AutoPromptManager {
...
+      : undefined;
+
+    // Override autoPromptType if it is undefined.
+    params.autoPromptType = params.autoPromptType
```
ChrisAntaki, line 197: This would be a little simpler
```suggestion
    params.autoPromptType = params.autoPromptType || audienceActionsPromptType;
```

kristenwang: Thanks for the suggestion!

src/runtime/auto-prompt-manager.js
```diff
@@ -410,6 +419,33 @@ export class AutoPromptManager {
...
+   * Show the first AutoPromptType passed in from Audience Actions.
+   * @param {{
+   *   article: (!./entitlements-manager.Article)
+   * }} article
+   * @return {!AutoPromptType|undefined}
+   */
+  getAutoPromptType_(article) {
+    const audienceActions = article.audienceActions?.actions || [];
+
+    const potentialActions = audienceActions.filter(
```
ChrisAntaki, line 434: Since we just need one match, we could use the `find()` method, which could simplify the code a bit
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

src/runtime/auto-prompt-manager.js
kristenwang: Seeing https://github.com/subscriptions-project/swg-js/blob/main/src/runtime/entitlements-manager.js#L53, the audienceActions isn't really a defined type, so in this file, it's usually the "article" value that gets passed around. If we were to pass audienceActions, what should I put here as the param type? https://github.com/subscriptions-project/swg-js/blob/3c29a8084d4ce843d2ff887fd89eb39a15da0f84/src/runtime/auto-prompt-manager.js#L429

ChrisAntaki: Oh, good call. Could you pass `article.audienceActions?.actions` as a `string[]`?

src/runtime/auto-prompt-manager.js
kristenwang: Done.

src/runtime/auto-prompt-manager.js
```diff
@@ -410,6 +420,29 @@ export class AutoPromptManager {
...
   }
 
+  /**
+   * Determines what Audience Action prompt type should be shown.
+   *
+   * Show the first AutoPromptType passed in from Audience Actions.
+   * @param {string[]|undefined} actions
+   * @return {!AutoPromptType|undefined}
+   */
+  getAutoPromptType_(actions) {
```
ChrisAntaki, line 430: This would ensure the `actions.find()` method available
```suggestion
  getAutoPromptType_(actions = []) {
```

src/runtime/auto-prompt-manager.js
```diff
@@ -187,6 +189,14 @@ export class AutoPromptManager {
     dismissedPrompts,
     params
   ) {
+    const audienceActionsPromptType =
+      article && article.audienceActions && article.audienceActions.actions
+        ? this.getAutoPromptType_(article.audienceActions.actions)
+        : undefined;
+
+    // Override autoPromptType if it is undefined.
+    params.autoPromptType = params.autoPromptType || audienceActionsPromptType;
```
ChrisAntaki, line 198: With this change, you could remove lines 192-195
```suggestion
    params.autoPromptType = params.autoPromptType || this.getAutoPromptType_(article?.audienceActions?.actions);
```

kristenwang: Done.

src/runtime/auto-prompt-manager.js
kristenwang: Done.

ChrisAntaki (approved):
Looks good to me, thanks for making those changes

PR #2796, opened 2023-03-03
Migrate `src/utils/log` to TypeScript

ChrisAntaki:
- Migrates file to TypeScript
- Expands test to match codecov diff target


src/utils/log.js
```diff
@@ -1,80 +0,0 @@
...
- * Throws an error if the first argument isn't trueish.
- *
- * Supports argument substitution into the message via %s placeholders.
- *
- * @param {T} shouldBeTrueish The value to assert. The assert fails if it does
- *     not evaluate to true.
- * @param {string=} message The assertion message
- * @param {...*} var_args Arguments substituted into %s in the message.
- * @return {T} The value of shouldBeTrueish.
- * @template T
```
fikus, line 56: Generally we should preserve API documentation. When migrating to TS we can just remove the Closure type information.

ChrisAntaki: Good call, it's back now

fikus (approved):

PR #2799, opened 2023-03-03
Migrate `src/utils/i18n` to TypeScript

ChrisAntaki:


src/utils/i18n.ts
```diff
@@ -19,11 +19,11 @@ const DEFAULT_LANGUAGE_CODE = 'en';
 
 /**
  * Gets a message for a given language code, from a map of messages.
- * @param {!Object<string, string>} map
- * @param {?string|?Element} languageCodeOrElement
- * @return {?string}
  */
-export function msg(map, languageCodeOrElement) {
+export function msg(
+  map: {[key: string]: string},
```
fikus, line 24: I think the value type should be `string?` since in general `map[key]` can be undefined.

src/utils/i18n.ts
```diff
@@ -19,11 +19,11 @@ const DEFAULT_LANGUAGE_CODE = 'en';
...
+  languageCodeOrElement: string | HTMLElement
+): string {
```
fikus, line 26: Similarly this can return `string?`.

fikus (approved):

src/utils/i18n.ts
ChrisAntaki: Makes sense

PR #2800, opened 2023-03-03
Migrate `src/utils/json` to TypeScript

ChrisAntaki:
- Migrates json util to TypeScript
- Removes unnecessary `parseJson` method, which just existed for Closure typecasting
- Replaces `parseJson` with `JSON.parse` across several files


src/utils/json.ts
```diff
@@ -53,11 +40,11 @@ export function tryParseJson(json, onFailed) {
...
- * @param {string} propertyName
- * @return {*}
  */
-export function getPropertyFromJsonString(jsonString, propertyName) {
+export function getPropertyFromJsonString(
+  jsonString: string,
+  propertyName: string
+): any {
   const json = tryParseJson(jsonString);
   return (json && json[propertyName]) || null;
```
fikus, line 49: Doesn't look like this handles falsey values correctly. For example, it will return `null` for `getPropertyFromJsonString('{"value": false}', 'value')`. Can you add a TODO to fix this (separately from TS migration)?

fikus (approved):

src/utils/json.ts
ChrisAntaki: Added 👍

PR #2804, opened 2023-03-03
getPropertyFromJsonString: Handle falsy values

ChrisAntaki:


src/utils/json.ts
```diff
@@ -46,5 +46,5 @@ export function getPropertyFromJsonString(
   propertyName: string
 ): any {
   const json = tryParseJson(jsonString);
-  return (json && json[propertyName]) || null; // TODO: Handle falsy values.
+  return (json && json[propertyName]) ?? null;
```
fikus, line 49: The updated code returns the value of the property, or null if the input string is not valid JSON or the given property is not present in the JSON. If that's the desired behavior it should be specifically documented. It might make more sense to write this as

```
return json?.[propertyName];
```

so the method returns `undefined` if the property is missing or the string is not valid JSON, and only returns `null` if the value is the JSON `null` literal.

fikus: Either way the behavior should be verified with unit tests.

ChrisAntaki: Good call, I updated the logic and added a test

src/utils/json-test.js
```diff
@@ -58,9 +58,14 @@ describe('json', () => {
...
+    it('returns undefined if value is missing', () => {
       const value = getPropertyFromJsonString('{}', 'x');
-      expect(value).to.equal(null);
+      expect(value).to.be.undefined;
+    });
+
+    it('returns undefined if json is invalid', () => {
+      const value = getPropertyFromJsonString('...', 'x');
+      expect(value).to.be.undefined;
     });
```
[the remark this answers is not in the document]
ChrisAntaki: For sure, I just added a "returns false if value is false" test

fikus (approved):

PR #2806, opened 2023-03-04
Simplify `src/utils/style`

ChrisAntaki:
- Removes unused features (ex: vendor prefixing)
- Uses `element.style.setProperty` and `element.style.getPropertyValue`, instead of accessing `element.style` directly (will help with TypeScript migration)

### DIFF
--- a/src/components/friendly-iframe-test.js
+++ b/src/components/friendly-iframe-test.js
@@ -61,7 +61,6 @@ describes.realWin('FriendlyIframe', (env) => {
       expect(friendlyIframe.isConnected()).to.be.true;
 
       setImportantStyles(iframe, importantStyles);
-      setStyles(iframe, {});
 
       expect(getStyle(iframe, 'opacity')).to.equal('1');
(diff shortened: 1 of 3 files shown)


src/components/friendly-iframe-test.js
```diff
@@ -61,7 +61,6 @@ describes.realWin('FriendlyIframe', (env) => {
       expect(friendlyIframe.isConnected()).to.be.true;
 
       setImportantStyles(iframe, importantStyles);
-      setStyles(iframe);
```
ChrisAntaki, line 64: This was a noop

src/utils/style.js
```diff
@@ -14,12 +14,6 @@
  * limitations under the License.
  */
 
-/** @type {Object<string, string>} */
-let propertyNameCache;
-
-/** @const {!Array<string>} */
-const vendorPrefixes = ['Webkit', 'webkit', 'Moz', 'moz', 'O', 'o'];
```
ChrisAntaki, line 21: Never actually used

fikus (approved):

PR #2812, opened 2023-03-07
Migrate `src/utils/jwt` to TypeScript

ChrisAntaki:


src/utils/jwt.ts
```diff
@@ -35,19 +32,12 @@ export class JwtHelper {
 
   /**
    * Decodes JWT token and returns its payload.
-   * @param {string} encodedToken
-   * @return {?JsonObject|undefined}
    */
-  decode(encodedToken) {
+  decode(encodedToken: string): any {
```
fikus, line 36: We should minimize use of `any`, especially in public APIs. If this is intended to be opaque, can you update the documentation to say so and change the return type to `{}`?

src/utils/jwt.ts
```diff
@@ -35,19 +32,12 @@ export class JwtHelper {
...
     return this.decodeInternal_(encodedToken).payload;
   }
 
-  /**
-   * @param {string} encodedToken
-   * @return {!JwtTokenInternalDef}
-   * @private
-   */
-  decodeInternal_(encodedToken) {
+  decodeInternal_(encodedToken: string): JwtTokenInternalDef {
```
fikus, line 40: Should be marked `private`.

src/utils/jwt.ts
ChrisAntaki: Updated!

src/utils/jwt.ts
ChrisAntaki: Updated!

src/utils/jwt.ts
ChrisAntaki: In this case, I think `any` is appropriate since a JWT could have anything in its payload

ChrisAntaki: As a general utility, this needs to support a variety of JWTs. Perhaps the code calling this method could cast the value to a more specific type?

fikus (approved):

src/utils/jwt.ts
ChrisAntaki: Cool, I think `any` works well here. When we migrate JavaScript files that expects a certain kind of JSON in the payload, I think that could be a good time to define an `interface` for the expected JSON response. As for the no-any rule, maybe we could discuss that more offline?

ChrisAntaki: Updated to use `unknown` instead of `any`

PR #2815, opened 2023-03-07
Migrate `src/model/attribution-params` to TypeScript

ChrisAntaki:


fikus (approved):

src/model/attribution-params.ts
```diff
@@ -18,15 +18,11 @@
...
-   * @param {string} avatarUrl
-   */
-  constructor(displayName, avatarUrl) {
-    /** @const {string} */
-    this.displayName = displayName;
+  displayName: string;
+  avatarUrl: string;
 
-    /** @const {string} */
+  constructor(displayName: string, avatarUrl: string) {
```
hartmannr76, line 24: Nit: Internal style guide recommends using [Parameter Properties](https://www.typescriptlang.org/docs/handbook/2/classes.html#parameter-properties) "Rather than plumbing an obvious initializer through to a class member"

This class can just be
```ts
export class AttributionParams {
  constructor(public readonly displayName: string, public readonly avatarUrl: string) {
}
```

ChrisAntaki: Nice, I totally forgot about this feature 🔥

PR #2819, opened 2023-03-07
Use configurationId from getArticle in audienceActionFlow

justinchou-google:
b/243704873
ConfigurationId is already being passed through article endpoint. Updates AudienceActionFlow to pass this configurationId along with the actionType during the triggering flow.


src/runtime/entitlements-manager.js
```diff
@@ -92,6 +92,7 @@ export class AvailableIntervention extends Intervention {
     const flow = new AudienceActionFlow(this.deps_, {
       isClosable: params.isClosable,
       action: this.type,
+      configurationId: this.configurationId,
```
justinchou-google, line 95: @mhkawano can you confirm this change is required?

mhkawano: Yes, this looks right. We will also need the config Id.

mhkawano (approved):
LGTM

src/runtime/auto-prompt-manager.js
```diff
@@ -531,8 +532,8 @@ export class AutoPromptManager {
       );
 
       if (contributionIndex > 0) {
-        actionToUse = potentialActions[0].type;
-        this.promptDisplayed_ = actionToUse;
+        actionToUse = potentialActions[0];
+        this.promptDisplayed_ = actionToUse.type;
```
oyj9109, line 536: (Not actionable) FYI
One thing to think about suppression logic for frequency-based prompts is how we're going to handle it for multi-configs type (e.g., survey). We may want to distinguish each survey for suppression then we should log impression/dismissals per configid, not per type.
Just leaving a note as it's out of scope and needs more discussion with PM/UX.

src/runtime/auto-prompt-manager.js
```diff
@@ -531,8 +532,8 @@ export class AutoPromptManager {
       );
```
oyj9109, line 532: Comment on line 527: `return TYPE_REWARDED_SURVEY;`
Should we convert it to an `Intervention` instance? 
(I think it would've caught by tests. Do you know why it wasn't?)

Not sure why I can't comment on the line :'(

src/runtime/auto-prompt-manager.js
justinchou-google: I agree, I made a note of it on b/243704873 talking about prompt dismissals too. I will add this case to the bug as well.

src/runtime/auto-prompt-manager-test.js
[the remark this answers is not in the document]
justinchou-google: Added constants for the three audience action interventions.

src/runtime/auto-prompt-manager.js
justinchou-google: expanded the field `interventionDisplayed_` to include the configurationId. updated the survey tests to also assert equality on this field.

src/runtime/auto-prompt-manager.js
```diff
@@ -546,7 +550,7 @@ export class AutoPromptManager {
           previouslyShownPrompts.includes(AutoPromptType.CONTRIBUTION_LARGE)
         )
       ) {
-        this.promptDisplayed_ = AutoPromptType.CONTRIBUTION;
+        this.interventionDisplayed_ = {type: AutoPromptType.CONTRIBUTION};
```
oyj9109, line 553: One thing we should be careful in the future change here.
We're planning to change audience action engine to serve monet prompt type (subscription/contribution) as a part of pre-monet work. It means we'll need to handle audience actions of `AudienceActionType.TYPE_CONTRIBUTION`  in the flow while we used to have a clear distinction between auto-prompts and non-monet action prompts.
We should make it clear that what could be broken with the change. For instance, what if we replace `AutoPromptType.CONTRIBUTION` with `AudienceActionType.TYPE_CONTRIBUTION` here without knowing they are mapped to different string values? Suppression logic can be broken by mistake.

Sorry for long explanation. Can you put a comment to prevent such a situation? (like AutoPromptType.CONTRIBUTION is different from AudienceActionType.TYPE_CONTRIBUTION so should be careful on change)

src/runtime/auto-prompt-manager.js
oyj9109: I like how they are renamed! Thanks

oyj9109 (approved):
LGTM!

src/runtime/auto-prompt-manager.js
```diff
@@ -515,24 +515,28 @@ export class AutoPromptManager {
...
       if (
         prioritizeSurvey &&
         potentialActions
           .map((action) => action.type)
           .includes(TYPE_REWARDED_SURVEY)
       ) {
-        this.promptDisplayed_ = TYPE_REWARDED_SURVEY;
-        return TYPE_REWARDED_SURVEY;
+        const surveyAction = potentialActions.find(
+          (action) => action.type === TYPE_REWARDED_SURVEY
```
ChrisAntaki, line 525:
```suggestion
          ({type}) => type === TYPE_REWARDED_SURVEY
```

ChrisAntaki (approved):

src/runtime/auto-prompt-manager.js
justinchou-google: Added a warning regarding difference between `AutoPromptType` and `AudienceActionType`

PR #2820, opened 2023-03-08
Use audienceActions to determine prompt type

kristenwang:
b/270965538. Currently we use the experiment flag to determine whether to show survey first or contribution prompt first, and we want to deprecated the flag in the future and use the order of actions from audienceActions to determine the prompt to show here.


ChrisAntaki (approved):

kristenwang:
After further discussion with Ricky, I changed the logic because we want the following to happen:
1. If the SURVEY_TRIGGERING_PRIORITY experiment is turned on, we will show survey prompt first if it's not previously dismissed.
2. If SURVEY_TRIGGERING_PRIORITY experiment is turned off, and we have contribution in the audience actions, then we will show the first prompt (not dismissed before) from audience actions.
3. If SURVEY_TRIGGERING_PRIORITY experiment is turned off, and we don't have the contribution in the audience actions, we will follow the current behavior.

src/runtime/auto-prompt-manager.js
```diff
@@ -502,37 +502,40 @@ export class AutoPromptManager {
...
         ExperimentFlags.SURVEY_TRIGGERING_PRIORITY
       );
 
-      // If audienceActions contains both contribution and survey actions,
-      // we will use the order from audienceActions to determine the preference.
-      const contributionIndex = potentialActions.findIndex(
+      let previousPrompts = [];
+      if (dismissedPrompts) {
+        previousPrompts = dismissedPrompts.split(',');
+        previousPrompts = previousPrompts.map((action) =>
```
ChrisAntaki, line 514: Could we combine this line with the line above?
```suggestion
         previousPrompts.map((action) =>
```

ChrisAntaki: Ex: `previousPrompts = dismissedPrompts.split(',').map(...`

src/runtime/auto-prompt-manager.js
```diff
@@ -502,26 +502,40 @@ export class AutoPromptManager {
...
-        await this.isExperimentEnabled_(
-          article,
-          ExperimentFlags.SURVEY_TRIGGERING_PRIORITY
+      // Survery take highest priority if this flag is enabled.
+      const prioritizeSurvey = await this.isExperimentEnabled_(
+        article,
+        ExperimentFlags.SURVEY_TRIGGERING_PRIORITY
+      );
+
+      let previousPrompts = [];
```
ChrisAntaki, line 511: Nit: This name could be clearer
```suggestion
      let previouslyShownPrompts = [];
```

src/runtime/auto-prompt-manager.js
```diff
@@ -502,26 +502,40 @@ export class AutoPromptManager {
       autoPromptType === AutoPromptType.CONTRIBUTION ||
       autoPromptType === AutoPromptType.CONTRIBUTION_LARGE
     ) {
-      const preferSurveyOverContributionPrompt =
-        await this.isExperimentEnabled_(
-          article,
-          ExperimentFlags.SURVEY_TRIGGERING_PRIORITY
+      // Survery take highest priority if this flag is enabled.
+      const prioritizeSurvey = await this.isExperimentEnabled_(
```
ChrisAntaki, line 506: We don't need this variable until later on in the code. Could you move the definition closer to where it's first used?

src/runtime/auto-prompt-manager.js
kristenwang: Done

src/runtime/auto-prompt-manager.js
kristenwang: Done.

ChrisAntaki:
Looks good 👍

src/runtime/auto-prompt-manager.js
kristenwang: Done.

PR #2824, opened 2023-03-08
Migrate `src/model/page-config` to TypeScript

ChrisAntaki:


src/model/page-config.ts
```diff
@@ -14,14 +14,13 @@
...
 export class PageConfig {
-  /**
-   * @param {string} productOrPublicationId
-   * @param {boolean} locked
-   */
-  constructor(productOrPublicationId, locked) {
+  publicationId_: string;
+  productId_: string | null;
+  label_: string | null;
+  locked_: boolean;
```
fikus, line 21: Looks like these should be private, and as noted in the other PR we can optionally remove the underscore from the names.

fikus (approved):

src/model/page-config.ts
ChrisAntaki: For now, the `_` suffixes are improving minification

PR #2833, opened 2023-03-10
Respect ordering from audienceActions to determine prompts

kristenwang:
b/270965538.


src/runtime/auto-prompt-manager.js
```diff
@@ -545,20 +526,40 @@ export class AutoPromptManager {
...
+        (action) => action.type === TYPE_CONTRIBUTION
+      );
+
+      if (contributionIndex > 0) {
+        actionToUse = potentialActions[0].type;
+        this.promptDisplayed_ = actionToUse;
+        return actionToUse;
+      }
+
+      // If the first potential action is contribution, and it has never been
```
ChrisAntaki, line 539: What if contribution isn't a potential action? (Ex: contributionIndex was -1)

kristenwang: When contributionIndex is -1, it will be treated the same as when contributionIndex === 0.

ChrisAntaki: In that case, should the comment on line 539 change?

ChrisAntaki: Thanks!

ChrisAntaki (approved):

PR #2834, opened 2023-03-10
Migrate `src/api/offer` to TypeScript

ChrisAntaki:
- Migrates to TypeScript
- Updates an import and references in `src/runtime/offers-api.js`


src/runtime/offers-api.js
```diff
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+/* eslint-disable-next-line @typescript-eslint/no-unused-vars */
```
ChrisAntaki, line 17: Note: Only necessary during the TypeScript migration. Once the `src/runtime/offers-api.js` file is TypeScript, we can remove this eslint comment

fikus (approved):

PR #2836, opened 2023-03-10
Migrate `src/api/logger-api` and `src/api/propensity-api` TypeScript

ChrisAntaki:
- Migrates to TypeScript
- Updates `src/runtime/event-type-mapping-test.js` test
  - Updates an import
  - Simplifies two tests
  - Removes a redundant test


ChrisAntaki:
Oops, this isn't quite ready for review, but should be by the end of the day

ChrisAntaki:
It's ready now, I had to improve a test 🧹

src/api/logger-api.ts
```diff
@@ -212,30 +211,23 @@ export const Event = {
...
    * Send a buy-flow event that occurred on the publisher's site to Google.  The
    * ultimate destination is controlled by configuration settings.  Publisher
    * configuration available:
    *   enablePropensity - Sends data to the Propensity to Subscribe ads server.
    *   enableSwgAnalytics - Sends data to Google's analytics server for buy-flow
    *     comparison purposes.
-   * @param {!PublisherEvent} userEvent
    */
-  sendEvent(userEvent) {}
+  sendEvent(_userEvent: PublisherEvent): void;
```
fikus, line 230: Why does _userEvent have an underscore prefix in this interface? Same question for the other interfaces.

ChrisAntaki: This is the standard way to mark a parameter as unused in TypeScript

fikus: Interface parameters should not need to be marked unused, there is no implementation so all parameters are unused.

ChrisAntaki: Good catch, I just removed the leading `_`s

src/api/logger-api.ts
```diff
@@ -247,9 +239,9 @@ export class LoggerApi {
    *     {'product': ['product1', 'product2']}
    * Each call to this API should have the first argument
    * as a valid string from the enum SubscriptionState.
-   * @param {SubscriptionState} state
-   * @param {?JsonObject} jsonProducts
    */
-  sendSubscriptionState(state, jsonProducts) {}
+  sendSubscriptionState(
+    state: SubscriptionState,
+    jsonProducts: {[key: string]: string[]}
```
fikus, line 245: Can this be typed more precisely with {product: string[]}?

fikus (approved):

src/api/logger-api.ts
ChrisAntaki: Updated!

PR #2843, opened 2023-03-14
Implement first part of new interventions api.

mhkawano:
This is the first part of the planned "Funnel Prompts for Enterprise" API. This implements the basic classes and methods for us to be able to begin dogfooding the API internally without including it in `subscriptions.js`. The current implementation does not have support for the `onResult` param yet.


src/runtime/entitlements-manager-test.js
```diff
@@ -1937,6 +1941,55 @@ describes.realWin('EntitlementsManager', (env) => {
...
+        pageConfig,
+        fetcher,
+        deps,
+        /* useArticleEndpoint */ false
+      );
+      sandbox.stub(manager, 'getArticle').resolves(null);
+
+      expect(await manager.getAvailableInterventions()).to.equal(
+        null,
+        'getAvailableInterventions should return null'
```
ChrisAntaki, line 1957: Could we log an error that advises a dev on how to add support for this API?

mhkawano: Done.

src/runtime/entitlements-manager-test.js
```diff
@@ -1937,6 +1941,59 @@ describes.realWin('EntitlementsManager', (env) => {
...
+      );
+
+      expect(self.console.warn).to.have.been.calledWithExactly(
+        '[swg.js:getAvailableInterventions] Article is null. Make sure you have enabled it in the client ready callback with: `subscriptions.configure({enableArticleEndpoint: true})`'
+      );
+    });
+
+    it('should correct AvailableInterventions', async () => {
```
ChrisAntaki, line 1965:
```suggestion
    it('should return correct AvailableInterventions', async () => {
```

src/runtime/entitlements-manager.js
```diff
@@ -50,16 +51,61 @@ import {warn} from '../utils/log';
...
+  constructor(original, deps) {
+    super();
+    Object.assign(this, original);
+    /** @private @const {!./deps.DepsDef} */
+    this.deps_ = deps;
+  }
+  /**
+   * Starts the intervention flow.
+   * @param {!ShowInterventionParams=} params
+   * @return {?}
```
ChrisAntaki, line 89: Does it return a Promise?

src/runtime/entitlements-manager.js
```diff
@@ -50,16 +51,61 @@ import {warn} from '../utils/log';
...
+ *                intervention should be marked complete.
+ *
+ * @typedef {{
+ *   isClosable: (boolean|undefined),
+ *   onResult: ((function(!Object):(Promise<Boolean>|Boolean))|undefined),
+ * }}
+ */
+export let ShowInterventionParams;
+
+class Intervention {
```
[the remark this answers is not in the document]
ChrisAntaki: Perhaps not with Closure Compiler. When we migrate to TypeScript, I think this will be possible though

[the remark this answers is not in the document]
mhkawano: I think the appropriate way to do it in TypeScript would be an `interface`, which we can extend on for a class.

src/runtime/entitlements-manager.js
mhkawano: In this case it would always be a promise. I left it generic because there might be situations in the future where a publisher provided intervention would not be async. Maybe they would add an element(banner/ad) to the page and don't care to wait on it?

ChrisAntaki: For now, I think Promise makes more sense. We could always change the type later if needed?

mhkawano: Sure. Do we want to keep it that way when adding the interface to the API?

ChrisAntaki: That sounds good to me 👍

mhkawano: Done. Copied the return type of `AudienceActionFlow.start`.

src/runtime/entitlements-manager.js
```diff
@@ -50,16 +51,61 @@ import {warn} from '../utils/log';
...
  * Article response object.
  *
  * @typedef {{
  *  entitlements: (../api/entitlements.Entitlements),
  *  clientConfig: (../model/client-config.ClientConfig),
  *  audienceActions: ({
- *    actions: Array<{
- *      type: (string)
- *    }>,
+ *    actions: Array<Intervention>,
```
ChrisAntaki, line 108: Just checking, can the entries be `null`?
```suggestion
 *    actions: Array<!Intervention>,
```

ChrisAntaki (approved):

PR #2847, opened 2023-03-15
Provide an option to bypass SwG init in GaaMetering

teiswit:
This PR provides an alternative solution to #2784.
Publishers require an option to bypass `subscriptions.init` when the entitlement checking flow is operated by 3P.

Internal tracking: b/271063736

### DIFF
--- a/src/runtime/extended-access/gaa-metering.js
+++ b/src/runtime/extended-access/gaa-metering.js
@@ -118,8 +118,15 @@ export class GaaMetering {
         ? () => {}
         : params.unlockArticle;
 
+    // Provide an option to bypass SwG init for 3P integrations.
+    const shouldInitializeSwG = params.shouldInitializeSwG
+      ? params.shouldInitializeSwG
+      : true;
+
     callSwg(async (subscriptions) => {
-      subscriptions.init(productId);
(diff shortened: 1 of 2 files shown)


src/runtime/extended-access/gaa-metering.js
```diff
@@ -118,8 +118,14 @@ export class GaaMetering {
         ? () => {}
         : params.unlockArticle;
 
+    // Provide an option to bypass SwG init for 3P integrations.
+    const shouldInitializeSwG =
+      'shouldInitializeSwG' in params ? params.shouldInitializeSwG : true;
```
ChrisAntaki, line 123:
```suggestion
      params.shouldInitializeSwG ?? true;
```

src/runtime/extended-access/gaa-metering.js
```diff
@@ -451,6 +457,16 @@ export class GaaMetering {
       }
     }
 
+    if (
+      'shouldInitializeSwG' in params &&
+      typeof params.shouldInitializeSwG != 'boolean'
```
ChrisAntaki, line 462:
```suggestion
      typeof params.shouldInitializeSwG !== 'boolean'
```

PR #2848, opened 2023-03-15
Add onResult to AvailableIntervention.

mhkawano:
Adds `onResult` support in `AvailableIntervention`. This replaces `setOnInterventionComplete` which was originally planned to be used for "Funnel Prompts for Enterprise". This was chosen to simplify the API for third parties.

### DIFF
--- a/src/runtime/audience-action-flow-test.js
+++ b/src/runtime/audience-action-flow-test.js
@@ -34,6 +34,7 @@ import {Constants} from '../utils/constants';
 import {PageConfig} from '../model/page-config';
 import {ProductType} from '../api/subscriptions';
 import {Toast} from '../ui/toast';
+import {tick} from '../../test/tick';
 
 const WINDOW_LOCATION_DOMAIN = 'https://www.test.com';
 const WINDOW_INNER_HEIGHT = 424242;
@@ -124,6 +125,10 @@ describes.realWin('AudienceActionFlow', (env) => {
     sandbox.stub(runtime, 'win').returns(win);
     onCancelSpy = sandbox.spy();
     sandbox.useFakeTimers(CURRENT_TIME);
+  });
+
+  afterEach(() => {
+    eventManagerMock.verify();
   });
 
   function setWinWithoutGtag() {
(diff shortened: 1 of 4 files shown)


src/runtime/audience-action-flow.js
```diff
@@ -290,11 +291,14 @@ export class AudienceActionFlow {
...
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
-  handleSurveyDataTransferRequest_(request) {
+  async handleSurveyDataTransferRequest_(request) {
     // @TODO(justinchou): execute callback with setOnInterventionComplete
     // then check for success
-    const gaLoggingSuccess = this.logSurveyDataToGoogleAnalytics(request);
-    if (gaLoggingSuccess) {
+    const {onResult} = this.params_;
+    const loggingSuccess = onResult
+      ? await this.params_.onResult(request)
```
ChrisAntaki, line 299:
```suggestion
      ? await onResult(request)
```

ChrisAntaki: We might want to gracefully support rejects/throws from the `onResult` callback

ChrisAntaki: We could `await onResult(request)` within a `try/catch` block, and if it fails, we could set `loggingSuccess` to false

src/runtime/audience-action-flow.js
```diff
@@ -290,11 +291,14 @@ export class AudienceActionFlow {
...
    */
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
-  handleSurveyDataTransferRequest_(request) {
+  async handleSurveyDataTransferRequest_(request) {
     // @TODO(justinchou): execute callback with setOnInterventionComplete
     // then check for success
-    const gaLoggingSuccess = this.logSurveyDataToGoogleAnalytics(request);
-    if (gaLoggingSuccess) {
+    const {onResult} = this.params_;
+    const loggingSuccess = onResult
```
oyj9109, line 298: nit: what about 'dataTransferSuccess'?

src/runtime/audience-action-flow-test.js
oyj9109, line 1: nit: can you rename the existing test cases to be specfiic to Google Analytics?

src/runtime/audience-action-flow.js
oyj9109: Good catch! +1 to error handling

oyj9109 (approved):

src/runtime/audience-action-flow.js
mhkawano: Moved this area into a new function `attemptSurveyDataTransfer` and wrapped `onResult` in a try/catch with a warning if it catches an exception.

src/runtime/audience-action-flow.js
mhkawano: Done.

src/runtime/audience-action-flow-test.js
mhkawano: Done.

mhkawano:
>BTW, setOnInterventionComplete is not limited to RRM:E. We will need to provide an API for RRM:S to be able to trasnfer data to their system. I agree getting a callback onResult as a parameter of show function for RRM:E would be more convenient for them, but are we okay with supporting setOnInterventionComplete later for RRM:S and maintaining two different paths for setting callbacks?
I didn't mean to imply that it would replace the use cases for RRM:S. The consensus in the after review discussion seemed to be that we were okay with two paths.

src/runtime/audience-action-flow.js
```diff
@@ -311,10 +311,31 @@ export class AudienceActionFlow {
...
+   * @return {boolean}
+   * @private
+   */
+  async attemptSurveyDataTransfer(request) {
+    // @TODO(justinchou): execute callback with setOnInterventionComplete
+    // then check for success
+    const {onResult} = this.params_;
+    if (onResult) {
+      try {
+        return onResult(request);
```
ChrisAntaki, line 330: I think this should `await` the onResult call, in case it returns a promise

ChrisAntaki: It could return a promise that eventually rejects

mhkawano: Done.

src/runtime/audience-action-flow-test.js
```diff
@@ -713,7 +722,147 @@ describes.realWin('AudienceActionFlow', (env) => {
...
+    const messageCallback = messageMap[TEST_SURVEYDATATRANSFERREQUEST.label()];
+    messageCallback(TEST_SURVEYDATATRANSFERREQUEST);
+
+    await tick(10);
+
+    activityIframeViewMock.verify();
+    onResultMock.verify();
+  });
+
+  it(`handles a SurveyDataTransferRequest with onResult logging exception`, async () => {
```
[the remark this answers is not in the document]
mhkawano: Done.

src/runtime/audience-action-flow-test.js
```diff
@@ -713,7 +722,196 @@ describes.realWin('AudienceActionFlow', (env) => {
...
+    );
+    activityIframeViewMock.verify();
+    onResultMock.verify();
+  });
+
+  it(`handles a SurveyDataTransferRequest with onResult logging rejection`, async () => {
+    const onResultMock = sandbox
+      .mock()
+      .withExactArgs(TEST_SURVEYDATATRANSFERREQUEST)
+      .resolves(Promise.reject(new Error('Test Callback Exception')))
```
ChrisAntaki, line 872: Would this work?
```suggestion
      .rejects(new Error('Test Callback Exception'))
```

ChrisAntaki (approved):
Looks good 👍

PR #2874, opened 2023-03-24
Migrate `src/components/activities` to TypeScript

ChrisAntaki:
- Add `d.ts` file for the `web-activities/activity-ports` module
- Migrate `src/components/activities` to TypeScript
- Migrate `src/runtime/deps` to TypeScript, creating TODOs for types we don't have yet.
- Create new mocks (`MockDeps` and `MockActivityPort`) for unit tests. Many unit tests were instantiating interfaces, which is no longer possible with TypeScript.


src/runtime/deps.ts
```diff
@@ -0,0 +1,70 @@
...
+ */
+
+import {ActivityPorts} from '../components/activities';
+import {Config} from '../api/subscriptions';
+import {DialogManager} from '../components/dialog-manager';
+import {Doc} from '../model/doc';
+import {JsError} from './jserror';
+import {PageConfig} from '../model/page-config';
+
+export interface DepsDef {
```
fikus, line 24: Should we rename this to Deps? I think elsewhere we've been renaming Def from the name since we no longer need to bypass the jslint for unused types.

test/mock-activity-port.js
```diff
@@ -0,0 +1,33 @@
+/**
```
fikus, line 1: Can we not write this in TypeScript?

src/@types/web-activities/activity-ports.d.ts
```diff
@@ -0,0 +1,336 @@
...
+    private readonly connectedPromise_: Promise<void>;
+    private readyResolver_: (() => void) | null;
+    private readonly readyPromise_: Promise<void>;
+    private resultResolver_:
+      | ((result: ActivityResult | Promise<ActivityResult>) => void)
+      | null;
+    private readonly resultPromise_: Promise<ActivityResult>;
+    private onResizeRequest_: ((size: number) => void) | null;
+    private requestedHeight_: number | null;
+    private readonly messenger_: Messenger;
```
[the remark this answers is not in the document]
ChrisAntaki: Good point, I just removed those private fields

test/mock-activity-port.js
ChrisAntaki: That works, updated

src/runtime/deps.ts
ChrisAntaki: For sure, updated

test/mock-activity-port.ts
```diff
@@ -0,0 +1,30 @@
...
+ * distributed under the License is distributed on an "AS-IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * This mock exists for unit tests.
+ */
+export class MockActivityPort {
```
fikus, line 20: Should we make these new mock classes implement the interface they're mocking?

ChrisAntaki: Since the tests don't need that, I don't think it would really be worth it

ChrisAntaki: The current JavaScript code just returns `undefined` for all methods

fikus: Can we back each of these methods with a publicly settable field? They can default to `undefined` but let users assign an override for convenience in testing.

fikus: This is what I would expect from the test implementation, but we can defer that as a future enhancement instead of adding that to this migration.

fikus (approved):

PR #2875, opened 2023-03-24
Improve coverage in `src/api/subscribe-response`

ChrisAntaki:
- Fully covers a partially covered line
image


ChrisAntaki:
That makes sense, I'll add more tests

ChrisAntaki:
Tests are added now, and this is ready for another review

src/api/subscribe-response-test.js
```diff
@@ -0,0 +1,107 @@
...
+      expect(mockCompleteHandler).not.to.be.called;
+      subscribeResponse.complete();
+      expect(mockCompleteHandler).to.be.called;
+    });
+  });
+});
+
+describes.realWin('PurchaseData', () => {
+  describe('json', () => {
+    it('returns JSON', () => {
```
fikus, line 90: nit: It does not return JSON (which would be a string). It returns an object.

fikus (approved):

src/api/subscribe-response-test.js
ChrisAntaki: Nice catch, I updated the description to "returns a serializable object"

PR #2877, opened 2023-03-27
Migrate `src/ui/activity-iframe-view` and `src/utils/activity-utils` to TypeScript

ChrisAntaki:


src/@types/web-activities/activity-ports.d.ts
```diff
@@ -69,16 +69,22 @@ declare module 'web-activities/activity-ports' {
...
+    readonly data: unknown;
+    readonly mode: ActivityMode;
+    readonly origin: string;
+    readonly originVerified: boolean;
+    readonly secureChannel: boolean;
     readonly ok: boolean;
     readonly error: Error | null;
 
     constructor(
-      readonly code: ActivityResultCode,
```
ChrisAntaki, line 76: Note: This shorthand doesn't work in `.d.ts` files 😭

fikus: Should we clean up the other places we're using that syntax, like lines 113-115? (Note, private fields should not need to be declared in the types file, so those only need to specify the constructor interface).

src/runtime/meter-toast-api-test.js
```diff
@@ -341,6 +342,7 @@ describes.realWin('MeterToastApi', (env) => {
     activitiesMock.expects('openIframe').resolves(port);
     const messageStub = sandbox.stub(port, 'execute');
     await meterToastApi.start();
+    await tick(1);
```
fikus, line 345: I'm curious, why did the test need to be updated?

src/ui/activity-iframe-view.ts
```diff
@@ -0,0 +1,169 @@
...
+
+    return this.port_.whenReady();
+  }
+
+  private getPortPromise_(): Promise<ActivityIframePort> {
+    return this.portPromise_;
+  }
+
+  async on<T extends Message>(
+    message: new (data?: unknown[], includesLabel?: boolean) => T,
```
fikus, line 109: Should this use the `MessageConstructor` type?

ChrisAntaki: I think this template approach could work better.

We don't export the `MessageConstructor` type currently. I tried recreating it and calling it from an example method, and ran into an issue (screenshots below).
image
image


Whereas if I stick with the template approach, the example method works (screenshot below).
image

src/runtime/meter-toast-api-test.js
ChrisAntaki: I had updated the `ActivityIframeView`'s `whenComplete` method to be `async` and return `Promise<void>`, since nothing was using its result. By making that function `async` though, it meant a rejection was happening an event loop later. I modified the tests so they would one more event loop for the error behavior to happen.

I just updated the code to remove the `async` modifier, which means the tests don't need to be updated anymore. It does mean the `whenComplete` method returns something that the app code isn't using though. I'm cool with either approach. What do you think?

fikus: Ah, I see, this is because of the type of `whenComplete` in `View`. I don't think the abstract method should be declared to return `ActivityResult`, `View` is meant to be general and should not be specific to activity results. Can we make ActivityIframeView return Promise<void> to match the interface, but not change it to async to avoid changing any behavior?

ChrisAntaki: In this case, I think using `async` would improve the code. If we're just `await`ing the result without returning, then async/await would express that pretty clearly. End users wouldn't notice a difference. The tests would just need those `tick()`s

ChrisAntaki: I reverted the code so that `whenComplete` has an `async` modifier and return `Promise<void>`. This is the best approach I can think of currently, and it doesn't change the behavior for end users. If you have a specific suggestion that would make the code more readable, I'm definitely interested in hearing it.

fikus: OK, SG, thanks.

fikus (approved):

PR #2882, opened 2023-03-28
Use audience actions to determine isClosable.

mhkawano:


src/runtime/basic-runtime.js
```diff
@@ -228,6 +228,7 @@ export class BasicRuntime {
     this.setupAndShowAutoPrompt({
       autoPromptType,
       alwaysShow,
+      isAccessibleForFree,
```
ChrisAntaki, line 231: Could you update `src/api/basic-subscriptions.ts` too?

src/runtime/auto-prompt-manager.js
```diff
@@ -132,6 +132,7 @@ export class AutoPromptManager {
    *   autoPromptType: (AutoPromptType|undefined),
    *   alwaysShow: (boolean|undefined),
    *   displayLargePromptFn: (function()|undefined),
+   *   isAccessibleForFree: (boolean|undefined),
```
ChrisAntaki, line 135: Could you update the `showAutoPrompt_` method's comment too? Or possibly create a @typedef ShowAutoPromptParams both could share?

src/runtime/auto-prompt-manager.js
```diff
@@ -190,17 +192,23 @@ export class AutoPromptManager {
...
-      params.autoPromptType ||
-      this.getAutoPromptType_(article?.audienceActions?.actions);
+    params.autoPromptType ??= this.getAutoPromptType_(
+      article?.audienceActions?.actions
+    );
+
+    // Override isClosable if isAccessibleForFree is defined.
+    const isClosable =
+      !!params.isAccessibleForFree ||
+      (params.autoPromptType != AutoPromptType.SUBSCRIPTION &&
```
ChrisAntaki, line 202: Could you use identity operators when possible? Ex: `!==` instead of `!=` https://google.github.io/styleguide/jsguide.html#features-equality-checks

src/runtime/auto-prompt-manager.js
```diff
@@ -190,17 +192,23 @@ export class AutoPromptManager {
...
   ) {
     // Override autoPromptType if it is undefined.
-    params.autoPromptType =
-      params.autoPromptType ||
-      this.getAutoPromptType_(article?.audienceActions?.actions);
+    params.autoPromptType ??= this.getAutoPromptType_(
+      article?.audienceActions?.actions
+    );
+
+    // Override isClosable if isAccessibleForFree is defined.
```
ChrisAntaki, line 199: Is this code checking whether `isAccessibleForFree` is defined? It seems like values of `false` and `undefined` would be treated the same

mhkawano: In this case I'm casting it to a bool with `!!`, so it should convert undefined to false. That should be fine here, but maybe we would be better off just checking `isLocked`... I'll look at it again.

src/runtime/basic-runtime.js
mhkawano: Done.

src/runtime/auto-prompt-manager.js
mhkawano: Done.

src/runtime/auto-prompt-manager.js
mhkawano: Looked at it again, pretty sure I'm wrong. Properly checking if it's undefined first.

src/runtime/auto-prompt-manager.js
ChrisAntaki: Cool, maybe I missed it

src/runtime/auto-prompt-manager.js
ChrisAntaki: Sweet, I'm taking another look

src/runtime/auto-prompt-manager.js
```diff
@@ -190,17 +192,23 @@ export class AutoPromptManager {
...
+    params.autoPromptType ??= this.getAutoPromptType_(
+      article?.audienceActions?.actions
+    );
 
+    // Override isClosable if isAccessibleForFree is defined.
+    const isClosable =
+      params.isAccessibleForFree !== undefined
+        ? params.isAccessibleForFree
+        : params.autoPromptType !== AutoPromptType.SUBSCRIPTION &&
+          params.autoPromptType !== AutoPromptType.SUBSCRIPTION_LARGE;
```
oyj9109, line 204: nit: seems like there are lots of duplicate logic to define subscription/contribution based on `AutoPromptType`. Would it be useful to create a function like `isSubscriptionPrompt`?

src/runtime/auto-prompt-manager.js
```diff
@@ -190,17 +192,23 @@ export class AutoPromptManager {
...
   ) {
     // Override autoPromptType if it is undefined.
-    params.autoPromptType =
-      params.autoPromptType ||
-      this.getAutoPromptType_(article?.audienceActions?.actions);
+    params.autoPromptType ??= this.getAutoPromptType_(
+      article?.audienceActions?.actions
+    );
 
+    // Override isClosable if isAccessibleForFree is defined.
```
oyj9109, line 199: Can we make the comment clearer to understand how we determine closability? Something like..
```
If isAccessibleForFree is set in page config, ...
Otherwise, for a publication with subscription revenue model selected, prompts are blocking
for a publication with contribution revenue model selected or of pre-monetization (no revenue model selected), prompts are dismissible.
```

We're using `AutoPromptType` to determine revenue model type and I just realized that it can be confusing because it doesn't necessarily represent the prompt type of what we will actually end up showing. Not asking to fix it right now but something we should think about renaming in the long term.

oyj9109 (approved):
Left a few comments for nit, but overall LGTM!

src/runtime/auto-prompt-manager.js
mhkawano: Added some helper functions.

src/runtime/auto-prompt-manager.js
mhkawano: Updated the comment.

I think it make sense to rename `AutoPromptType` to something like `RevenueModel`.

src/runtime/auto-prompt-manager.js
```diff
@@ -190,26 +192,27 @@ export class AutoPromptManager {
...
+    );
 
-    if (
-      params.autoPromptType === AutoPromptType.SUBSCRIPTION ||
-      params.autoPromptType === AutoPromptType.SUBSCRIPTION_LARGE
-    ) {
+    // Override isClosable if isAccessibleForFree is set in the page config.
+    // Otherwise, for publications with a subscription revenue model the
+    // prompt is blocking, while all others can be dismissed.
+    const isClosable =
```
ChrisAntaki, line 202: Nit: Could this work?
```
const isClosable = params.isAccessibleForFree ?? !this.isSubscription_(params);
```

PR #2886, opened 2023-03-28
Migrate `src/runtime/event-type-mapping` to TypeScript

ChrisAntaki:


src/runtime/event-type-mapping.ts
```diff
@@ -225,49 +229,53 @@ export const ContributionSpecificAnalyticsEventToGoogleAnalyticsEvent = {
...
-      gaEvent = SubscriptionSpecificAnalyticsEventToGoogleAnalyticsEvent[event];
-    } else if (subscriptionFlow == SubscriptionFlows.CONTRIBUTE) {
-      gaEvent = ContributionSpecificAnalyticsEventToGoogleAnalyticsEvent[event];
-    }
+export function analyticsEventToGoogleAnalyticsEvent(
+  event: AnalyticsEvent | null,
+  subscriptionFlow: string
+): GoogleAnalyticsEvent | void {
+  if (!event) {
+    return;
```
fikus, line 271: I think the original code would return null in this case, can we preserve that behavior? It's unusual for code to conditionally return void.

src/runtime/event-type-mapping.ts
```diff
@@ -32,28 +31,27 @@ const PublisherEventToAnalyticsEvent = {
...
-  [AnalyticsEvent.ACTION_ACCOUNT_ACKNOWLEDGED]: null,
-  [AnalyticsEvent.ACTION_SUBSCRIPTIONS_LANDING_PAGE]:
-    Event.ACTION_SUBSCRIPTIONS_LANDING_PAGE,
-  [AnalyticsEvent.ACTION_PAYMENT_FLOW_STARTED]:
-    Event.ACTION_PAYMENT_FLOW_STARTED,
-  [AnalyticsEvent.ACTION_OFFER_SELECTED]: Event.ACTION_OFFER_SELECTED,
-  [AnalyticsEvent.EVENT_PAYMENT_FAILED]: null,
-  [AnalyticsEvent.EVENT_CUSTOM]: Event.EVENT_CUSTOM,
-};
+const AnalyticsEventToPublisherEvent: {[key in AnalyticsEvent]?: Event | null} =
```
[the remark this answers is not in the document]
ChrisAntaki: Here's the error message I see when removing the `?`.
image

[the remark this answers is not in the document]
ChrisAntaki: Looks like TypeScript is complaining that not every `AnalyticsEvent` is covered

src/runtime/event-type-mapping.ts
ChrisAntaki: In this case, I think returning `null` makes the code more complicated (adds branches) and less readable.
image

I think returning `void` works well here. The app only calls the method in one place, and checks whether the result is falsy.
image

ChrisAntaki: I think the JS could return `undefined`/`void` sometimes too. Ex: if `subscriptionFlow` was defined and `event` was `null`

fikus (approved):

PR #2893, opened 2023-03-31
Migrate Extended Access to TypeScript

ChrisAntaki:


src/runtime/extended-access/gaa-metering-test.js
```diff
@@ -246,18 +246,6 @@ describes.realWin('GaaMetering', () => {
     }
   }
 
-  describe('constructor', () => {
-    it('sets class variable', () => {
-      const gaaMeteringInstance = new GaaMetering();
```
ChrisAntaki, line 251: Note: There was no reason to instantiate this class, since it just contains static methods.

fikus (approved):

PR #2896, opened 2023-03-31
Clean up use-article-endpoint experiment flag

fikus:
This flag was used for switching the basic runtime to the article endpoint, and has been fully launched since March 2022.

Tests were updated to stub getEntitlements before creating the ConfiguredBasicRuntime instance, as with this change getEntitlements is always called by the ConfiguredBasicRuntime constructor.

b/276458173


ChrisAntaki (approved):

ChrisAntaki:
Thanks for cleaning this up!

PR #2906, opened 2023-04-04
Migrate `src/runtime/logger` to TypeScript

ChrisAntaki:


src/runtime/google-analytics-event-listener.ts
```diff
@@ -79,8 +80,9 @@ export class GoogleAnalyticsEventListener {
 
     // additionalParameters isn't strongly typed so checking for both object and class notation.
     const subscriptionFlow =
-      event.additionalParameters?.subscriptionFlow ||
-      event.additionalParameters?.getSubscriptionFlow?.() ||
+      (event.additionalParameters as {[key: string]: string})
+        ?.subscriptionFlow ||
```
fikus, line 84: ['subscriptionFlow'] since the expression is cast to an indexed type?

fikus (approved):

src/runtime/google-analytics-event-listener.ts
ChrisAntaki: Nice catch, updated

PR #2909, opened 2023-04-05
Migrate `src/runtime/entitlements-manager` to TypeScript

ChrisAntaki:


src/api/entitlements.ts
```diff
@@ -231,9 +231,11 @@ export class Entitlements {
    * When a metering entitlement is consumed, SwG shows the user a metering dialog.
    * When the user closes the dialog, SwG depletes one of the user's remaining
    * "free reads".
-   * @param {?Function=} onCloseDialog Called after the user closes the dialog.
    */
-  consume(onCloseDialog?: Function | null) {
+  consume(
+    /** Called after the user closes the dialog. */
```
fikus, line 236: This comment should be a param in the JSDoc.

src/runtime/entitlements-manager.ts
```diff
@@ -824,24 +732,19 @@ export class EntitlementsManager {
...
 
-  /**
-   * @param {!Entitlements} entitlements
-   * @param {?Function=} onCloseDialog Called after the user closes the dialog.
-   * @private
-   */
-  consume_(entitlements, onCloseDialog) {
+  private consume_(
+    entitlements: Entitlements,
+    /** Called after the user closes the dialog. */
```
fikus, line 743: Should be JSDoc param.

src/runtime/entitlements-manager.ts
```diff
@@ -850,25 +753,30 @@ export class EntitlementsManager {
...
         // Ignore decoding errors. Don't show a toast, and return
         // onConsumeCallback directly.
         return onConsumeCallback();
       }
 
-      if (
-        entitlement.subscriptionTokenContents['metering'] &&
-        entitlement.subscriptionTokenContents['metering']['showToast'] === true
-      ) {
+      if (subscriptionTokenContents['metering']['showToast'] === true) {
```
fikus, line 773: This is losing the check that ['metering'] exists, should we use optional chaining here?

fikus (approved):

src/api/entitlements.ts
ChrisAntaki: Updated

src/runtime/entitlements-manager.ts
ChrisAntaki: Nice catch, fixed

PR #2915, opened 2023-04-07
Implement handler for storing PPS in localStorage upon SurveyDataTransferRequest

caroljli:
Stores PPS values of AnswerChoices in QuestionCatalog upon survey submission request to send as populated GPT to GAM. The format for sending the PPS values via tag follows this example:

```
{
      [googletag.enums.Taxonomy.IAB_AUDIENCE_1_1]: 
        {values: ['12', '13', '14']},
      [googletag.enums.Taxonomy.IAB_CONTENT_2_2]:
        {values: ['22', '25', '57']},
}
```


src/proto/api_messages.ts
justinchou-google, line 1: Seems like the formatter didn't run, can we restore these changes?

src/runtime/audience-action-flow.js
```diff
@@ -340,6 +351,64 @@ export class AudienceActionFlow {
   }
 
   /**
+   * Populates localStorage with PPS configuration parameters based on
+   * SurveyDataTransferRequest.
+   * @param {SurveyDataTransferRequest} request
+   * @return {boolean}
+   * @private
+   **/
+
```
justinchou-google, line 360: nit: extra line

src/runtime/audience-action-flow.js
```diff
@@ -340,6 +351,64 @@ export class AudienceActionFlow {
...
+      let existingIabTaxonomyMap;
+      if (!existingIabTaxonomy) {
+        existingIabTaxonomyMap = {
+          '[googletag.enums.Taxonomy.IAB_AUDIENCE_1_1]': ppsConfigParams,
+        };
+      } else {
+        const existingIabMap = [
+          ...existingIabTaxonomy
+            .substring(existingIabTaxonomy.indexOf(':'))
+            .replace(/\D/g, ''),
```
justinchou-google, line 386: Can we make this into a function for clarity and unit testing?

src/runtime/audience-action-flow.js
```diff
@@ -340,6 +351,64 @@ export class AudienceActionFlow {
...
+        )
+      );
+      const result = Promise.resolve(this.storage_.get(iabAudienceKey, true));
+      return await !!Promise.resolve(
+        this.storage_.get(iabAudienceKey, /* useLocalStorage= */ true)
+      );
+    } else {
+      // Answers with no PPS parameters should always evaluate to true with
+      // the feature flag enabled.
+      return true;
```
justinchou-google, line 407: nit: can we move this null check to the beginning to simplify this control sequence?

src/proto/api_messages.ts
justinchou-google: seems like formatting was applied during migration: https://github.com/subscriptions-project/swg-js/commit/29658e96bac4141afb40d3f55494558843cc7c98#diff-ec8d982c9d32304c84c4f0f37b705b5b977b52215f9ad8095856126a6f217ee2R391
@ChrisAntaki , do we want to preserve these formatting changes where previously `api_messages.js` was just the raw output from the generator?

src/runtime/audience-action-flow.ts
```diff
@@ -321,6 +330,69 @@ export class AudienceActionFlow {
   }
 
   /**
+   * Populates localStorage with PPS configuration parameters based on
+   * SurveyDataTransferRequest.
+   * @param {SurveyDataTransferRequest} request
+   * @return {boolean}
+   * @private
```
fikus, line 337: Since this is TypeScript, types don't need to be specified in JSDoc. The `request` param can be removed from the doc (the type is in the param declaration). The return type and visibility (private) should be specified on the method itself.

src/runtime/audience-action-flow.ts
```diff
@@ -321,6 +330,69 @@ export class AudienceActionFlow {
...
+      )
+    );
+    return await !!Promise.resolve(
+      this.storage_.get(iabAudienceKey, /* useLocalStorage= */ true)
+    );
+  }
+
+  /**
+   * Filters array to contain only an array of digits after the identifier key
+   * in localStorage.
```
fikus, line 387: This looks inaccurate, the input is a string.

src/runtime/audience-action-flow.ts
```diff
@@ -321,6 +330,69 @@ export class AudienceActionFlow {
...
+   * @param {Array} list
+   * @returns
+   */
```
fikus, line 390: As above, the type information should be specified using TS syntax and can be removed from the JSDoc. Looks like this method is also meant to be private.

src/runtime/audience-action-flow.ts
caroljli: sorry about that! the migration happened while i was still in the process of writing this PR which is why i had the annotations.

src/runtime/audience-action-flow.ts
caroljli: sorry about that! migration happened mid-PR but i'll fix all the js->ts instances

src/runtime/audience-action-flow.ts
```diff
@@ -296,7 +296,16 @@ export class AudienceActionFlow {
       this.storage_.storeEvent(StorageKeys.SURVEY_DATA_TRANSFER_FAILED);
     }
     const surveyDataTransferResponse = new SurveyDataTransferResponse();
-    surveyDataTransferResponse.setSuccess(dataTransferSuccess);
+    const isPpsEligible = request.getStorePpsInLocalStorage();
+
+    if (isPpsEligible) {
+      const configurePpsSuccess = await this.configureAnswerPpsData(request);
+      surveyDataTransferResponse.setSuccess(
+        configurePpsSuccess && dataTransferSuccess
```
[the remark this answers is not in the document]
caroljli: hmm ok that makes sense. i can put a client log in order to handle this which will be more useful but since it doesn't affect data transfer it makes sense that iterating on both tese variables is not wise.

src/runtime/audience-action-flow.ts
```diff
@@ -321,6 +330,69 @@ export class AudienceActionFlow {
...
+      iabAudienceKey,
+      /* useLocalStorage= */ true
+    );
+    let existingIabTaxonomyMap;
+    if (!existingIabTaxonomy) {
+      existingIabTaxonomyMap = {
+        '[googletag.enums.Taxonomy.IAB_AUDIENCE_1_1]': ppsConfigParams,
+      };
+    } else {
+      const existingIabMap = this.filterDigits_(existingIabTaxonomy);
```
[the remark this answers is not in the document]
caroljli: it's because the key is 'ppstaxonomies' but the value itself is stored in localstorage as one string `'[googletag.enums.Taxonomy.IAB_AUDIENCE_1_1]': ppsConfigParams`, where only `ppsConfigParams` is the useful stringified array of numbers for us.

src/runtime/audience-action-flow.ts
```diff
@@ -321,6 +330,69 @@ export class AudienceActionFlow {
...
+      };
+    }
+    await Promise.resolve(
+      this.storage_.set(
+        iabAudienceKey,
+        JSON.stringify(existingIabTaxonomyMap),
+        /* useLocalStorage= */ true
+      )
+    );
+    return await !!Promise.resolve(
```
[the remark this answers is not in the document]
caroljli: sounds good! will get of all return types.

src/runtime/audience-action-flow.ts
```diff
@@ -321,6 +330,69 @@ export class AudienceActionFlow {
...
+    );
+    let existingIabTaxonomyMap;
+    if (!existingIabTaxonomy) {
+      existingIabTaxonomyMap = {
+        '[googletag.enums.Taxonomy.IAB_AUDIENCE_1_1]': ppsConfigParams,
+      };
+    } else {
+      const existingIabMap = this.filterDigits_(existingIabTaxonomy);
+      existingIabTaxonomyMap = {
+        '[googletag.enums.Taxonomy.IAB_AUDIENCE_1_1]': existingIabMap.concat(
```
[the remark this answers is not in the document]
caroljli: makes sense. added logic to remove.

src/runtime/audience-action-flow.ts
```diff
@@ -321,6 +330,69 @@ export class AudienceActionFlow {
...
+    }
+
+    const existingIabTaxonomy = await this.storage_.get(
+      iabAudienceKey,
+      /* useLocalStorage= */ true
+    );
+    let existingIabTaxonomyMap;
+    if (!existingIabTaxonomy) {
+      existingIabTaxonomyMap = {
+        '[googletag.enums.Taxonomy.IAB_AUDIENCE_1_1]': ppsConfigParams,
```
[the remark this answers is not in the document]
caroljli: yes, since the format takes on: 
```
googletag.setConfig({   pps: {     taxonomies: {      [googletag.enums.Taxonomy.IAB_AUDIENCE_1_1]:         {values: ['12', '13', '14']},      [googletag.enums.Taxonomy.IAB_CONTENT_2_2]:        {values: ['22', '25', '57']},    }  } });
--
```
i updated the key to accommodate the key for the values as defined the API.

src/runtime/audience-action-flow.ts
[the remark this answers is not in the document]
oyj9109: If we're storing something in a stringified object format, we should be able to parse it and get the value of object using the same key. In other words, for
```
itemToBeStored = {[PPS_KEY]: pps_values};
localStorage.setItem(STORAGE_KEY, JSON.stringify(itemToBeStored));
```
where `PPS_KEY` and `STORAGE_KEY` are string constants.

`JSON.parse(localStorage.getItem(STORAGE_KEY))` should be equal to `itemToBeStored` and thus `itemToBeStored[PPS_KEY]` is equal to `pps_values` which is an array of PPS string values.

src/runtime/audience-action-flow.ts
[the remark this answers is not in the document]
oyj9109: I think `concat` doesn't deal with de-duplication.

src/runtime/audience-action-flow.ts
```diff
@@ -321,6 +327,68 @@ export class AudienceActionFlow {
...
+    if (ppsConfigParams.length === 0 || !ppsConfigParams) {
+      log(`[swg.js] No PPS values found in survey answer choices.`);
+    }
+
+    const existingIabTaxonomy = await this.storage_.get(
+      iabAudienceKey,
+      /* useLocalStorage= */ true
+    );
+    let ppsConfigParamsValues;
+    if (!existingIabTaxonomy) {
```
oyj9109, line 354: Can we define a variable for the existing pps values (an array of string, which can be empty in case there's nothing) and append the current values to the array to construct a new array to override stored value without having to deal with if/else statements?

src/runtime/audience-action-flow.ts
```diff
@@ -321,6 +327,68 @@ export class AudienceActionFlow {
...
+      ppsConfigParamsValues = {values: ppsConfigParams};
+    } else {
+      const existingIabMap = this.filterDigits(existingIabTaxonomy);
+      ppsConfigParamsValues = {
+        values: existingIabMap.concat(ppsConfigParams as string[]),
+      };
+      await Promise.resolve(
+        this.storage_.remove(iabAudienceKey, /* useLocalStorage= */ true)
```
oyj9109, line 362: Do we need to remove the existing item? Can we just execute `this.storage_.set` to overwrite?

src/runtime/audience-action-flow.ts
caroljli: makes sense. will do that

src/runtime/audience-action-flow.ts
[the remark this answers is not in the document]
caroljli: This makes sense if we were calling the localStorage.getItem() methods (verified in console for InsideNova) but the storage_ API returns a string and therefore returns a parse error when this is fetched.

src/runtime/audience-action-flow.ts
```diff
@@ -321,6 +327,68 @@ export class AudienceActionFlow {
...
+  private async configureAnswerPpsData(
+    request: SurveyDataTransferRequest
+  ): Promise<void> {
+    const iabAudienceKey = StorageKeys.PPS_TAXONOMIES;
+    // PPS value field is optional and category may not be populated
+    // in accordance to IAB taxonomies.
+    const ppsConfigParams = request
+      .getSurveyQuestionsList()!
+      .flatMap((question) => question.getSurveyAnswersList())
+      .map((answer) => answer?.getPpsValue())
```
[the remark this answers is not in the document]
caroljli: the linter wouldn't compile without the `?` operator

src/runtime/audience-action-flow.ts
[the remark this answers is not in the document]
caroljli: ah I see. didn't know wrapping in square brackets would omit the default variable value. thanks!

src/runtime/audience-action-flow.ts
```diff
@@ -321,6 +327,47 @@ export class AudienceActionFlow {
...
+      .filter((ppsValue) => ppsValue !== null);
+
+    const existingIabTaxonomy = await this.storage_.get(
+      iabAudienceKey,
+      /* useLocalStorage= */ true
+    );
+    const existingIabTaxonomyValues = existingIabTaxonomy
+      ? JSON.parse(existingIabTaxonomy)[Constants.PPS_AUDIENCE_TAXONOMY_KEY]
```
oyj9109, line 350: nit: JSON.parse should be wrapped by try-catch for error handling.

Parsing an item stored in localStorage can throw a runtime error if the item is not in the expected format. If it fails to parse, we can just fallback the existing values to an empty array and move forward with the current values.

src/runtime/audience-action-flow.ts
```diff
@@ -321,6 +327,47 @@ export class AudienceActionFlow {
...
+      .getSurveyQuestionsList()!
+      .flatMap((question) => question.getSurveyAnswersList())
+      .map((answer) => answer?.getPpsValue())
+      .filter((ppsValue) => ppsValue !== null);
+
+    const existingIabTaxonomy = await this.storage_.get(
+      iabAudienceKey,
+      /* useLocalStorage= */ true
+    );
+    const existingIabTaxonomyValues = existingIabTaxonomy
```
oyj9109, line 349: nit: use optional chaining to avoid a runtime error when the expected key doesn't exist.

`JSON.parse(existingIabTaxonomy)?.[Constants.PPS_AUDIENCE_TAXONOMY_KEY]?.values` || []

(You don't need to nullness of `existingIabTaxonomy` as `JSON.parse(null)` evalues to `null` and it will be nicely handled by optional chaining.)

src/runtime/audience-action-flow.ts
```diff
@@ -321,6 +327,47 @@ export class AudienceActionFlow {
...
+      ? JSON.parse(existingIabTaxonomy)[Constants.PPS_AUDIENCE_TAXONOMY_KEY]
+          .values
+      : [];
+    const iabTaxonomyValues = Array.from(
+      new Set(ppsConfigParams.concat(existingIabTaxonomyValues))
+    );
+    const existingIabTaxonomyMap = {
```
oyj9109, line 356: nit: it's not existing, but something new to be store. What about `iabTaxonomy` or `ppsTaxonomy`?

src/runtime/audience-action-flow.ts
```diff
@@ -321,6 +327,47 @@ export class AudienceActionFlow {
   }
 
   /**
+   * Populates localStorage with PPS configuration parameters based on
+   * SurveyDataTransferRequest.
+   **/
+  private async configureAnswerPpsData(
```
oyj9109, line 333: nit: what about `storePpsValuesFromSurveyAnswers`?

src/runtime/audience-action-flow-test.js
```diff
@@ -1010,6 +1046,105 @@ describes.realWin('AudienceActionFlow', (env) => {
...
+    const audienceActionFlow = new AudienceActionFlow(runtime, {
+      action: 'TYPE_REWARDED_SURVEY',
+      configurationId: 'configId',
+      onCancel: onCancelSpy,
+      autoPromptType: AutoPromptType.CONTRIBUTION,
+    });
+    activitiesMock.expects('openIframe').resolves(port);
+
+    const existingIabTaxonomyMap = {
+      [Constants.PPS_AUDIENCE_TAXONOMY_KEY]: {values: ['3', '4']},
```
oyj9109, line 1092: Can you add a value that's not in new to existing so that the test covers adding (de-duped) values to existing one?

oyj9109 (commented):
Left a few more comments for nit. Overall looks very good and I really like how neat it ended up.

src/runtime/audience-action-flow.ts
caroljli: ahh that makes sense. would be super helpful, added :)

src/runtime/audience-action-flow.ts
caroljli: makes sense. forgot about optional chaining since the file migrated to ts haha

src/runtime/audience-action-flow.ts
```diff
@@ -321,6 +327,50 @@ export class AudienceActionFlow {
...
+        JSON.parse(existingIabTaxonomy)?.[Constants.PPS_AUDIENCE_TAXONOMY_KEY]
+          ?.values || [];
+      const iabTaxonomyValues = Array.from(
+        new Set(ppsConfigParams.concat(existingIabTaxonomyValues))
+      );
+      const iabTaxonomy = {
+        [Constants.PPS_AUDIENCE_TAXONOMY_KEY]: {values: iabTaxonomyValues},
+      };
+
+      await Promise.resolve(
```
oyj9109, line 360: Can we take it out of try/catch so that the current pps values are still stored even if reading/parsing existing signals fails?

caroljli: yeah, ended up making it fall back to a default value

caroljli:
fixing unit tests

src/runtime/audience-action-flow-test.js
oyj9109: To be more specific with an example, I meant something like
- existing : 3, 4, 5
- new: 1, 2, 3
- expect: 1, 2, 3, 4, 5

src/runtime/audience-action-flow-test.js
```diff
@@ -148,6 +174,16 @@ describes.realWin('AudienceActionFlow', (env) => {
     sandbox.stub(runtime, 'win').returns(winWithNoGtag);
   }
 
+  async function addToLocalStorage(key, value) {
+    await runtime.storage().set(key, value, true);
```
oyj9109, line 178: Can we set up an expectation of the return value instead of actually interacting a local storage instance which is suppoed to be mocked?

```
      storageMock
        .expects('get')
        .withExactArgs(...)
        .resolves(...)
        .once();
```

src/runtime/audience-action-flow.ts
```diff
@@ -321,6 +327,55 @@ export class AudienceActionFlow {
...
+
+    const existingIabTaxonomy = await this.storage_.get(
+      iabAudienceKey,
+      /* useLocalStorage= */ true
+    );
+    let existingIabTaxonomyValues: string[] = [];
+    try {
+      const parsedExistingIabTaxonomyValues =
+        JSON.parse(existingIabTaxonomy)?.[Constants.PPS_AUDIENCE_TAXONOMY_KEY]
+          ?.values || [];
```
oyj9109, line 353: nit: `|| []` part is unnecessary as type is checked below.

package-lock.json
oyj9109, line 1: Can you revert this change if unnecessary?

oyj9109 (approved):
LGTM, Great work!

PR #2926, opened 2023-04-10
Migrate `src/runtime/callbacks` to TypeScript

ChrisAntaki:


src/runtime/callbacks.ts
```diff
@@ -0,0 +1,271 @@
...
+    );
+    return !!this.callbacks_[CallbackId.PAYMENT_RESPONSE];
+  }
+
+  hasPaymentResponsePending(): boolean {
+    return !!this.resultBuffer_[CallbackId.PAYMENT_RESPONSE];
+  }
+
+  setOnFlowStarted(
+    callback: ({flow, data}: {flow: string; data: unknown}) => void
```
justinchou-google, line 201: Should we be using type `object` here over `unknown`?

src/runtime/callbacks.ts
```diff
@@ -0,0 +1,271 @@
...
+  ): void {
+    this.setCallback_(CallbackId.FLOW_STARTED, callback as Callback);
+  }
+
+  /**
+   * @return Whether the callback has been found.
+   */
+  triggerFlowStarted(flow: string, data = {}): boolean {
```
justinchou-google, line 209: similarly here, not sure if type for `data` needs to be explicitly defined.

src/runtime/callbacks.ts
ChrisAntaki: Should I update `setOnFlowStarted` in `src/api/subscriptions.ts` to match?

src/runtime/callbacks.ts
ChrisAntaki: In this case, the default value implies `{}`. What do you think?
image

justinchou-google:
```
Object has similarities to both {} and object, but is slightly looser. Use {} for a type that include everything except null and undefined
```

Seems good to me!

src/runtime/callbacks.ts
justinchou-google: Yes please!

src/runtime/callbacks.ts
```diff
@@ -0,0 +1,271 @@
...
+    return this.trigger_(CallbackId.FLOW_STARTED, {
+      flow,
+      data,
+    });
+  }
+
+  setOnFlowCanceled(
+    callback: (params: {flow: string; data: unknown}) => void
```
justinchou-google, line 217: `data` can be updated here, too, and similarly for `subscriptions.ts`

ChrisAntaki: Updated!

src/runtime/callbacks.ts
ChrisAntaki: Updated!

justinchou-google (approved):

PR #2929, opened 2023-04-11
Extend AudienceAction Triggering and Suppression to Non-Monetary Revenue Models

justinchou-google:
b/277567112 
Specifically, when AutoPromptType is undefined, and not overwritten by the presence of a Contribution or Subscription action in audience actions.


ChrisAntaki (approved):

src/runtime/auto-prompt-manager.js
```diff
@@ -518,44 +515,47 @@ export class AutoPromptManager {
     // Default to the first recommended action.
     let actionToUse = potentialActions[0];
 
-    // Contribution prompts should appear before recommended actions, so we'll need
-    // to check if we have shown it before.
+    // For subscriptions, skip triggering checks and use the first potential action
+    if (this.isSubscription_({autoPromptType})) {
+      return actionToUse;
```
oyj9109, line 520: Q. Is it possible `actionToUse` equals to subscription? I'm wondering if it's safer to filter out contribution/subscription type from `potentialActions` as it's not supported in `AudienceActionFlow`. WDYT?

justinchou-google: My understanding is, as part of https://github.com/subscriptions-project/swg-js/pull/2833, we are planning on migrating contribution and subscription types into the article response as `potentialActions`. @kristenwang can you confirm?

kristenwang (approved):

src/runtime/auto-prompt-manager.js
oyj9109: Let's follow up this as a larger refactoring as I left comments in the other PR.

oyj9109 (approved):
LGTM. Thanks for covering a lot of different test cases!

PR #2934, opened 2023-04-12
Fix Prioritized Audience Action Prompts to not use Mini Prompt

justinchou-google:
b/277963129
There was an edge case where audience actions that take precedence over auto prompts were being launched via the `MiniPromptAPI` rather than the `AudienceActionFlow`. This fix resolves some testing around recording dismissals, but should not have had any effect on the reader experience. This also consolidates the flow for audience actions into one, as intended. The intervention displayed is also recorded for all autoprompts, rather than just for the contributions flow.


src/runtime/auto-prompt-manager.js
```diff
@@ -272,7 +272,12 @@ export class AutoPromptManager {
       (clientConfig?.autoPromptConfig?.clientDisplayTrigger
         ?.displayDelaySeconds || 0) * SECOND_IN_MILLIS;
 
-    if (shouldShowAutoPrompt) {
+    const interventionDisplayedIsAutoPromptType =
+      this.interventionDisplayed_ === null ||
+      // Only need to check contributions because Subsciptions interventions
+      // will not be saved.
+      this.isContribution_({autoPromptType: this.interventionDisplayed_.type});
```
justinchou-google, line 279: @kristenwang adding this check here due to edge case: https://github.com/subscriptions-project/swg-js/blob/main/src/runtime/auto-prompt-manager.js#L579

Do we plan to support this long term, or is this only maintained while we transition contributions to audience actions?

src/runtime/auto-prompt-manager.js
```diff
@@ -272,7 +272,12 @@ export class AutoPromptManager {
       (clientConfig?.autoPromptConfig?.clientDisplayTrigger
         ?.displayDelaySeconds || 0) * SECOND_IN_MILLIS;
 
-    if (shouldShowAutoPrompt) {
+    const interventionDisplayedIsAutoPromptType =
+      this.interventionDisplayed_ === null ||
+      // Only need to check contributions because Subsciptions interventions
```
oyj9109, line 277: I feel like `interventionDisplayed_` excluding subscription seems wrong. It can be unexpected as it's unclear in the name that this is only used for storing dismissals (for frequency-based triggering) and that's unneeded for subscription as there should be no suppression logic for subscription flow.

I think right thing to do is,
1) allow `interventionDisplayed_` to track any prompt type (remove the assumption that it's not subscription) 
2) allow dismissals for subscription prompt to be stored storeLastDismissal_.
3) Explicitly exclude subscription from the suppression logic that uses dismissed prompts.

src/runtime/auto-prompt-manager.js
```diff
@@ -272,7 +272,12 @@ export class AutoPromptManager {
...
+      // will not be saved.
+      this.isContribution_({autoPromptType: this.interventionDisplayed_.type});
+    if (shouldShowAutoPrompt && interventionDisplayedIsAutoPromptType) {
```
oyj9109, line 280: Can you elaborate more on the edge case?
For example, is it possible `shouldShowAutoPrompt` is `true` and `interventionDisplayed_` is survey? if so, what's expected to be but what's broken?

src/runtime/auto-prompt-manager.js
```diff
@@ -272,7 +272,12 @@ export class AutoPromptManager {
       (clientConfig?.autoPromptConfig?.clientDisplayTrigger
         ?.displayDelaySeconds || 0) * SECOND_IN_MILLIS;
 
-    if (shouldShowAutoPrompt) {
+    const interventionDisplayedIsAutoPromptType =
```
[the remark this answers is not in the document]
justinchou-google: Sure, created b/278069444

src/runtime/auto-prompt-manager.js
justinchou-google: Updated the subscription flow to record `interventionDisplayed_`.

src/runtime/auto-prompt-manager.js
oyj9109: `interventionDisplayedIsAutoPromptType` can evaluates to true when there's no intervention to be displayed. When `shouldShowAutoPrompt` is `false`, I think it should evaluate to `false` as well.
I think we should move `shouldShowAutoPrompt ` up in the part of `interventionDisplayedIsAutoPromptType`?

justinchou-google: If no intervention is displayed at all, the control sequence exits before it reaches this line:
https://github.com/subscriptions-project/swg-js/pull/2934/files#diff-c74361e6f3e0a3fc5042519e0ddf35b67dc861e52b588913d6ebcf696ff17f12R248

Are you suggesting the following?
```
const interventionDisplayedIsAutoPromptType = shouldShowAutoPromptType && (
    this.interventionDisplayed_ === null ||this.isAutoPromptType_(this.interventionDisplayed_.type));
if (interventionDisplayedIsAutoPromptType) {
```

src/runtime/auto-prompt-manager.js
```diff
@@ -517,6 +535,14 @@ export class AutoPromptManager {
...
     // For subscriptions, skip triggering checks and use the first potential action
     if (this.isSubscription_({autoPromptType})) {
+      this.interventionDisplayed_ = actionToUse;
+      if (shouldShowAutoPrompt || actionToUse.type === TYPE_SUBSCRIPTION) {
+        // WARNING: Refers explicity to the Subscription AutoPromptType,
+        // which CANNOT be a potential audience action. This is not to be
+        // confused with the AudienceActionType TYPE_SUBSCRIPTION, which
+        // is part of the pre-monetization effort.
+        this.interventionDisplayed_ = {type: AutoPromptType.SUBSCRIPTION};
+      }
```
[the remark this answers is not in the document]
justinchou-google: Spoke offline. Will update `getAudienceActionPromptType_` to filter out autoprompt types in audience actions.

src/runtime/auto-prompt-manager.js
justinchou-google: Spoke offline. The decision from `getAudienceActionType_` and the prompt set in `this.interventionDisplayed_` both honor the decision to display autoprompt type.

src/runtime/auto-prompt-manager-test.js
```diff
@@ -1789,10 +1814,10 @@ describes.realWin('AutoPromptManager', (env) => {
...
       setupPreviousImpressionAndDismissals(storageMock, {
-        dismissedPromptGetCallCount: 1,
+        dismissedPromptGetCallCount: 2,
         getUserToken: true,
       });
-      miniPromptApiMock.expects('create').once();
+      miniPromptApiMock.expects('create').never();
 
       await autoPromptManager.showAutoPrompt({
         autoPromptType: AutoPromptType.CONTRIBUTION,
```
oyj9109, line 1823: (FYI, Not actionable)
AFAIK, the "mini" contribution UI has been deprecated, and now AutoPromptType is set as `CONTRIBUTION_LARGE` by default in the code snippet provided in publisher center. So this edge case (for contribution) should not exist in practice.

src/runtime/auto-prompt-manager.js
```diff
@@ -436,6 +457,22 @@ export class AutoPromptManager {
...
+      ? TYPE_SUBSCRIPTION
+      : this.isContribution_({autoPromptType})
+      ? TYPE_CONTRIBUTION
+      : null;
+    if (autopromptActionType) {
+      const autopromptActionIndex = actions.findIndex(
+        (action) => action.type === autopromptActionType
+      );
+      if (autopromptActionIndex > 0) {
+        return false;
```
oyj9109, line 472: What if the non-autoprompt action ended up suppressed in `getAudienceActionPromptType_`? I think it can result in `shouldShowAutoPrompt = false` even when contribution should be shown. I have a concern that it could be misleading for future reference, so would rather prefer not covering this case as there's no defined behavior for when an audience action can have explicit higher priority.
I feel like respecting priority (for autoprompt type) from audience engine should be handled within `getAudienceActionPromptType_` once we can remove the customized logic (e.g., show contribution first whenever it's availalbe) and delegate triggering decision to the BE.

src/runtime/auto-prompt-manager.js
oyj9109: Q. Can we use `potentialAction` to determine triggering decision from 2 flows is an non-autoprompttype audience action?

src/runtime/auto-prompt-manager-test.js
justinchou-google: +1, created b/278874256 to cleanup miniPromptApi.

src/runtime/auto-prompt-manager.js
justinchou-google: Good point, though we currently do cover the case where an audience action has an explicit higher priority than contribution: https://github.com/subscriptions-project/swg-js/blob/main/src/runtime/auto-prompt-manager-test.js#L1916

autoprompt suppression logic:
https://github.com/subscriptions-project/swg-js/pull/2934/files/9d4e5bd1be0b9c64169d45bec880407f5601f00f#diff-c74361e6f3e0a3fc5042519e0ddf35b67dc861e52b588913d6ebcf696ff17f12L559-L581

This was added as a part of b/270965538, https://github.com/subscriptions-project/swg-js/pull/2820/files

If we want to maintain this, then `getAudienceActionType_` should not filter out autoprompttypes from potential actions.

src/runtime/auto-prompt-manager.js
```diff
@@ -503,21 +549,31 @@ export class AutoPromptManager {
...
-    let potentialActions = audienceActions.filter((action) =>
+    potentialActions = potentialActions.filter((action) =>
       this.checkActionEligibility_(action.type, isSurveyEligible)
     );
 
-    // No audience actions means use the default prompt.
+    // No audience actions means use the default prompt, if it should be shown.
     if (potentialActions.length === 0) {
+      if (shouldShowAutoPrompt) {
+        this.interventionDisplayed_ = this.isSubscription_({autoPromptType})
```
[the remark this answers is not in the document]
justinchou-google: Updated `this.interventionDisplayed_` to be consistent with audience actions, and prefer `TYPE_CONTRIBUTION` over `AutoPromptType.CONTRIBUTION`.

src/runtime/auto-prompt-manager.js
justinchou-google: Spoke offline, `getAudienceActionType_` will revert to check for contribution action.

oyj9109 (approved):
LGTM. Sorry for delay reviews and thank you so much for your patience!

ChrisAntaki (approved):

PR #2935, opened 2023-04-12
Migrate some runtime APIs to TypeScript

ChrisAntaki:


src/runtime/mini-prompt-api.ts
```diff
@@ -102,11 +103,9 @@ export class MiniPromptApi {
     const lang = this.clientConfigManager_.getLanguage();
     let textContent = '';
     if (options.autoPromptType === AutoPromptType.CONTRIBUTION) {
-      textContent =
-        msg(SWG_I18N_STRINGS.CONTRIBUTION_TITLE_LANG_MAP, lang) || '';
```
ChrisAntaki, line 106: Note: The language map would have to be missing an entry for `en` for the `|| ''` branch to hit

justinchou-google (approved):

PR #2936, opened 2023-04-13
Migrate runtime propensity scripts to TypeScript

ChrisAntaki:


src/runtime/propensity-server.ts
```diff
@@ -18,74 +18,78 @@ import {
...
+  constructor(
+    private readonly win_: Window,
+    private readonly deps_: Deps,
+    private readonly fetcher_: Fetcher
+  ) {
     this.publicationId_ = this.deps_.pageConfig().getPublicationId();
-    /** @private {?string} */
-    this.clientId_ = null;
-    /** @private @const {!./fetcher.Fetcher} */
-    this.fetcher_ = fetcher;
```
justinchou-google, line 48: Does `this.fetcher_` need to be set here?

src/runtime/propensity.ts
```diff
@@ -69,22 +71,20 @@ export class Propensity {
...
+  getPropensity(type?: PropensityType): Promise<PropensityScore | null> {
+    if (type && !Object.values(PropensityType).includes(type)) {
       throw new Error('Invalid propensity type requested');
     }
     if (!type) {
-      type = PropensityApi.PropensityType.GENERAL;
+      type = PropensityType.GENERAL;
     }
     return this.propensityServer_.getPropensity(
       this.win_.document.referrer,
```
justinchou-google, line 82: Similarly, I think this requires `this._win` to be set?

src/runtime/propensity-server.ts
ChrisAntaki: Oh, the new code uses (`parameter properties`)[https://google.github.io/styleguide/tsguide.html#parameter-property-comments] to simplify the code. So `this.fetcher_` basically gets set on line 63 now

src/runtime/propensity.ts
ChrisAntaki: Yep, it's a `parameter property`

justinchou-google (approved):

PR #2937, opened 2023-04-14
Migrate `src/runtime/pay client.ts` to type script

ChrisAntaki:


src/runtime/pay-client-test.js
```diff
@@ -467,11 +467,6 @@ describes.sandboxed('RedirectVerifierHelper', () => {
       expect(useVerifierSync()).to.be.null;
     });
 
-    it('should default to null if crypto random not supported', () => {
-      delete crypto.getRandomValues;
```
ChrisAntaki, line 471: Note: This check wasn't necessary anymore

src/runtime/pay-client.ts
```diff
@@ -416,17 +373,12 @@ export class RedirectVerifierHelper {
     return !!(
       supportsLocalStorage &&
       crypto &&
-      crypto.getRandomValues &&
```
justinchou-google, line 419: is this intended?

justinchou-google (approved):

src/runtime/pay-client.ts
ChrisAntaki: Yep! TypeScript was complaining that the check was unnecessary

PR #2942, opened 2023-04-17
Migrate `src/runtime/pay-flow` to TypeScript

ChrisAntaki:


src/runtime/pay-flow-test.js
```diff
@@ -1054,6 +1054,80 @@ describes.realWin('PayCompleteFlow', (env) => {
     await flow.readyPromise_;
   });
 
+  it('constructs valid flow w/ virtual gift url params even without order id', async () => {
```
fikus, line 1057: virtual gifts is TWG, so there's probably no need for a new test.

src/runtime/pay-flow.ts
```diff
@@ -586,59 +541,49 @@ export function parseSubscriptionResponse(deps, data, completeHandler) {
...
 /**
- * @param {!Object} swgData
- * @return {?UserData}
- * @package Visible for testing.
+ * Visible for testing.
  */
-export function parseUserData(swgData) {
-  const idToken = swgData['idToken'];
+export function parseUserData(swgData: object): UserData | null {
+  const idToken = (swgData as {idToken?: string})['idToken'];
```
fikus, line 554: Can you type `swgData` in the function signature instead, or does that break any callers?

src/runtime/pay-flow.ts
```diff
@@ -586,59 +541,49 @@ export function parseSubscriptionResponse(deps, data, completeHandler) {
...
- * @package Visible for testing.
+ * Visible for testing.
  */
-export function parseEntitlements(deps, swgData) {
-  if (swgData['signedEntitlements']) {
+export function parseEntitlements(
+  deps: Deps,
+  swgData: object
+): Entitlements | null {
+  if ((swgData as {signedEntitlements?: string})['signedEntitlements']) {
```
fikus, line 569: Same question here about swgData.

fikus (approved):

src/runtime/pay-flow-test.js
ChrisAntaki: The TypeScript migration is exposing branches that were missing coverage. If code relating to `VIRTUAL_GIFT` purchases is still necessary, I think testing all the branches makes sense. If it's dead code though, we can remove it in a followup PR.

src/runtime/pay-flow.ts
ChrisAntaki: Good call, updated

src/runtime/pay-flow.ts
ChrisAntaki: Updated

PR #2943, opened 2023-04-17
Migrate `src/runtime/button-api` to TypeScript

ChrisAntaki:


src/api/subscriptions.ts
```diff
@@ -538,7 +538,7 @@ export interface ButtonOptions {
   /** "Light" or "dark". Default is "light". */
   lang?: string;
   /** Whether to grey out the button. */
-  disable?: boolean;
+  enable?: boolean;
```
fikus, line 541: This is public API, renaming the field could be a breaking change for integrators. Was this an accidental change? I don't see any other updated references to the field.

src/runtime/button-api.ts
```diff
@@ -198,31 +191,30 @@ export class ButtonApi {
...
-      const elements = this.doc_
-        .getRootNode()
-        .querySelectorAll(`[${attribute}="${attributeValue}"]`);
-      for (let i = 0; i < elements.length; i++) {
+      const elements: HTMLElement[] = Array.from(
+        this.doc_
+          .getRootNode()
+          .querySelectorAll(`[${attribute}="${attributeValue}"]`)
+      );
+      // for (let i = 0; i < elements.length; i++) {
```
fikus, line 207: Remove commented code.

src/api/subscriptions.ts
ChrisAntaki: The typedef seemed inaccurate:
https://github.com/subscriptions-project/swg-js/blob/dd225930aaa7a938ede15190d90103722d2f6699/src/api/subscriptions.js#L613

The code actually checks for "enable" instead of "disable":
https://github.com/subscriptions-project/swg-js/blob/7d5617cc077a47f8867ba362699c0ab98a350940/src/runtime/button-api.js#L147

fikus: Ah, gotcha. In that case, can you update the comment?

fikus (approved):

src/api/subscriptions.ts
ChrisAntaki: Updated!

PR #2948, opened 2023-04-18
Fix cache keys in operating modes

ChrisAntaki:
b/236934384

### DIFF
--- a/src/runtime/services.ts
+++ b/src/runtime/services.ts
@@ -24,6 +24,7 @@ import {
 } from '../constants';
 import {addQueryParam, parseQueryString, parseUrl} from '../utils/url';
 
+/** Possible cache keys which can influence how Swgjs busts caches for iframes. */
 type CacheKey = 'zero' | 'nocache' | 'hr1' | 'hr12';
 
 /**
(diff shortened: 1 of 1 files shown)


src/runtime/services.ts
```diff
@@ -80,7 +83,7 @@ const QUAL: OperatingMode = {
   frontEnd: 'https://subscribe-qual.sandbox.google.com',
   payEnv: 'SANDBOX',
   playEnv: 'STAGING',
-  feCache: CACHE_KEYS.hr1,
+  feCache: 'hr1',
```
jpettitt, line 86: Why did you leave it caching ion Qual?

fikus:
Can you add detail or reference a bug describing this change? I'm not sure what the intention of the change is.

jpettitt:
> Can you add detail or reference a bug describing this change? I'm not sure what the intention of the change is.

b/236934384

src/runtime/services.ts
ChrisAntaki: Updated!

jpettitt (approved):

PR #2960, opened 2023-04-21
Implement paywallType option for server-side optimization

teiswit:
The current GAA always runs entitlement checks on the client-side even though a publisher implements a server-side paywall. This change will enable an option for a publisher to specify that their implementation is server-side through `paywallType` param.

- When `userState.granted` is `true`, a page is unlocked as usual regardless of `paywallType`.
- When `showcaseEntitlement` is provided, a page is treated as having `paywallType`  set to SERVER_SIDE. `showcaseEntitlement` will be consumed as usual.
- When `paywallType` is set to SERVER_SIDE, no entitlement checks will be done on the client-side i.e. no getEntitlements call. 

Internal tracking: b/263350663


src/runtime/extended-access/gaa-metering.ts
```diff
@@ -149,13 +151,27 @@ export class GaaMetering {
...
-        subscriptions.consumeShowcaseEntitlementJwt(showcaseEntitlement, () => {
-          // Consume the entitlement and trigger a dialog that lets the user
-          // know Google provided them with a free read.
-          unlockArticle();
-        });
+      } else if (paywallType == PaywallType.SERVER_SIDE) {
+        // For server-side paywall, publisher checks both their own and Google entitlements from their server.
+        // If a user has access from publisher, publisher sets userState.granted to true and provides grantReason.
+        // If a user has access from Google, publisher sets showcaseEntitlement.
+        // Otherwise, publisher ensures paywallType is set to SERVER_SIDE to avoid repeated entitlement checks
```
ChrisAntaki, line 158: Nice comments

ChrisAntaki (approved):

teiswit:
Thanks Chris!

PR #2963, opened 2023-04-21
Migrate `src/runtime/auto-prompt-manager` to TypeScript

ChrisAntaki:


src/runtime/auto-prompt-manager.ts
```diff
@@ -29,104 +45,82 @@ const TYPE_SUBSCRIPTION = 'TYPE_SUBSCRIPTION';
...
   /**
    * Returns an instance of MiniPromptApi. Can be overwridden by subclasses,
    * such as in order to instantiate a different implementation of
    * MiniPromptApi.
-   * @param {!./deps.Deps} deps
-   * @return {!MiniPromptApi}
-   * @protected
    */
-  getMiniPromptApi(deps) {
+  getMiniPromptApi(deps: Deps): MiniPromptApi {
```
fikus, line 123: Should this method still be marked `protected`?

src/runtime/auto-prompt-manager.ts
```diff
@@ -413,14 +384,16 @@ export class AutoPromptManager {
...
-    ) {
+    const userReachedMaxImpressions =
+      impressions.length >= autoPromptConfig.impressionConfig.maxImpressions;
+    const timeSinceLastImpression = Date.now() - lastImpression;
+    const timeToWaitAfterMaxImpressionsInSeconds =
+      autoPromptConfig.impressionConfig.maxImpressionsResultingHideSeconds || 0;
+    const timeToWaitAfterMaxImpressions =
+      timeToWaitAfterMaxImpressionsInSeconds * SECOND_IN_MILLIS;
+    const userWaitedLongEnough =
+      timeSinceLastImpression < timeToWaitAfterMaxImpressions;
```
fikus, line 395: Since the condition is `timeSinceLastImpression < timeToWaitAfterMaxImpressions`, `userWaitedLongEnough` does not seem like an accurate name. The user has not waited long enough for the time to wait to expire.

src/runtime/auto-prompt-manager.ts
```diff
@@ -779,9 +746,9 @@ export class AutoPromptManager {
       StorageKeys.DISMISSED_PROMPTS,
       /* useLocalStorage */ true
     );
-    const intervention = /** @type {./entitlements-manager/Intervention} */ (
-      this.interventionDisplayed_
-    );
+    const intervention =
+      /** @type {./entitlements-manager/Intervention} */ this
```
fikus, line 750: Update syntax.

src/runtime/auto-prompt-manager.ts
ChrisAntaki: It can even be `private`

src/runtime/auto-prompt-manager.ts
ChrisAntaki: Nice catch! Renamed to `waitingAfterMaxImpressions`

src/runtime/auto-prompt-manager.ts
ChrisAntaki: Good catch, updated

PR #2982, opened 2023-05-01
Populate navigation timing

henel677:
Populates the navigation timing fields: runtime creation and load event start delay

### DIFF
--- a/src/runtime/basic-runtime.ts
+++ b/src/runtime/basic-runtime.ts
@@ -289,7 +289,6 @@ export class BasicRuntime implements BasicSubscriptions {
 export class ConfiguredBasicRuntime implements Deps, BasicSubscriptions {
   private audienceActivityEventListener_?: AudienceActivityEventListener;
 
-  private readonly creationTimestamp_: number;
   private readonly doc_: Doc;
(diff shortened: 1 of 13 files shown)


src/runtime/runtime.ts
```diff
@@ -592,13 +598,16 @@ export class ConfiguredRuntime implements Deps, SubscriptionsInterface {
     clientOptions?: {
       lang?: string;
       theme?: ClientTheme;
-    }
+    },
+    creationTimestamp?: number
```
ChrisAntaki, line 602:
```suggestion
    private readonly creationTimestamp_ = 0
```

This change should remove the need for lines 609 and 563

ChrisAntaki (approved):

src/runtime/basic-runtime.ts
```diff
@@ -329,7 +327,7 @@ export class ConfiguredBasicRuntime implements Deps, BasicSubscriptions {
       integr,
       config,
       clientOptions,
-      creationTimestamp
+      this.creationTimestamp_
```
ChrisAntaki, line 330: This could be just `creationTimestamp_` since it's a method param too

PR #3008, opened 2023-05-16
Log events

henel677:
Adds generation of log event EVENT_RUNTIME_IS_READY


src/runtime/client-event-manager.ts
```diff
@@ -113,11 +113,15 @@ export class ClientEventManager implements ClientEventManagerApi {
...
+  logEvent(
+    event: ClientEvent,
+    eventParams?: ClientEventParams,
+    eventTime?: number
+  ) {
     validateEvent(event);
 
     // Always use current timestamp.
-    event.timestamp = Date.now();
+    event.timestamp = eventTime == null ? Date.now() : eventTime;
```
ChrisAntaki, line 124:
```suggestion
    event.timestamp = eventTime ?? Date.now();
```

src/runtime/client-event-manager.ts
```diff
@@ -113,11 +113,15 @@ export class ClientEventManager implements ClientEventManagerApi {
...
 
-  logEvent(event: ClientEvent, eventParams?: ClientEventParams) {
+  logEvent(
+    event: ClientEvent,
+    eventParams?: ClientEventParams,
+    eventTime?: number
+  ) {
     validateEvent(event);
 
     // Always use current timestamp.
```
ChrisAntaki, line 123: Should this comment change?

src/runtime/client-event-manager.ts
henel677: done

src/runtime/client-event-manager.ts
henel677: done

PR #3010, opened 2023-05-16
Remove some TWG logic from pay-flow

fikus:


src/runtime/pay-flow.ts
```diff
@@ -193,8 +193,8 @@ export class PayStartFlow {
       {
         forceRedirect:
           this.deps_.config().windowOpenMode == WindowOpenMode.REDIRECT,
-        // SwG basic and TwG flows do not support native.
-        forceDisableNative: paySwgVersion == '2' || paySwgVersion == '3',
+        // SwG basic does not support native.
+        forceDisableNative: paySwgVersion == '2',
```
ChrisAntaki, line 197:
```suggestion
        forceDisableNative: paySwgVersion === '2',
```

ChrisAntaki (approved):

src/runtime/pay-flow.ts
fikus: Thought about this, but this could break if paySwgVersion is ever passed as a number. Left this out to make sure the chagne is safe.

ChrisAntaki: Good call

PR #3018, opened 2023-05-19
Fix Subscription Action Path Mapping, clarify when a monetization prompt is triggered as a paywall vs. audience action

justinchou-google:
b/283310066

Fixes bug where the URL for fetching the Subscription prompt as an audience action was being incorrectly mapped.

Renames `shouldShowMonetizationPromptFromFrequencyCap` to clarify the check is based on frequency, and that an Autoprompt can still potentially trigger in some conditions. In the current behavior, Contribution prompts honor this, whereas Subscription prompts may override. See below comment.

Uses terminology "Monetization Prompt" as an initial effort to migrate away from "AutoPrompt", but does not replace all instances.


ChrisAntaki (approved):

src/runtime/auto-prompt-manager.ts
```diff
@@ -478,7 +479,10 @@ export class AutoPromptManager {
 
     // For subscriptions, skip triggering checks and use the first potential action
     if (this.isSubscription_({autoPromptType})) {
-      if (shouldShowAutoPrompt) {
```
oyj9109, line 481: Can we combine this with contribution check below? Ideally, I think we can refactor and simplify the overall flow like

1. Custom suppresion logic (filter out survey based on completiion/failures)
2. General suppresion logic (filter out previous dismissed prompts)
3. Pick the first one (return undefined if monetization prompt type so that `promptFn` gets assign with `displayLargePromptFn`

src/runtime/auto-prompt-manager.ts
```diff
@@ -271,9 +272,9 @@ export class AutoPromptManager {
 
   /**
    * Determines whether a mini prompt for contributions or subscriptions should
-   * be shown.
+   * be shown based on the frequency cap.
    */
-  async shouldShowAutoPrompt_(
+  async shouldShowMonetizationPromptFromFrequencyCap(
```
oyj9109, line 277: I think those two conditions should be generally applied regardless of it's frequency-based / metered flow
1) Don't show the prompt if publication predicate is false
2) Don't show the prompt if we found a valid entitlement

Should we break it down into 2 functions and early return if one of the two conditions above evaluates to `true`?

src/runtime/auto-prompt-manager.ts
justinchou-google: I agree this would be more consistent, however I believe we have to wait until all current contribution publishers have a `TYPE_CONTRIBUTION` action.

For contribution publishers without this action, the `shouldShowMonetizationPromptFromFrequencyCap` boolean overrides the potential action: https://github.com/subscriptions-project/swg-js/blob/main/src/runtime/auto-prompt-manager.ts#L504

I think it makes sense to hold off on this refactoring until after all publishers have their monetization action within audience actions, thoughts?

src/runtime/auto-prompt-manager-test.js
```diff
@@ -1362,6 +1375,86 @@ describes.realWin('AutoPromptManager', (env) => {
...
+      expect(startSpy).to.not.have.been.called;
+      expect(actionFlowSpy).to.not.have.been.called;
+      expect(contributionPromptFnSpy).to.not.have.been.called;
+      expect(subscriptionPromptFnSpy).to.have.been.calledOnce;
+      expect(
+        autoPromptManager.wasAutoPromptDisplayedUncappedByFrequency_
+      ).to.equal(false);
       expect(autoPromptManager.getLastAudienceActionFlow()).to.equal(null);
       expect(autoPromptManager.interventionDisplayed_.type).to.equal(
         'TYPE_SUBSCRIPTION'
```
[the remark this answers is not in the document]
justinchou-google: oops, forgot to publish this comment from last week.

src/runtime/auto-prompt-manager.ts
justinchou-google: Are you suggesting we refactor the `uiPredicates` and `autoPromptType` checks out of this function, into, let's say, `shouldShowMonetizationPromptFromPredicate()`? And define boolean `shouldShowMonetizationPrompt = shouldShowMonetizationPromptFromPredicate() && shouldShowMonetizationPromptFromFrequencyCap()`?

src/runtime/auto-prompt-manager-test.js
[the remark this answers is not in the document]
justinchou-google: oops forgot to publish this comment from last week.

src/runtime/auto-prompt-manager.ts
oyj9109: Ahh good catch. I think I missed when `contributionIndex = -1`. Holding off on the refactoring sounds good to me. Thanks for explaining!

src/runtime/auto-prompt-manager.ts
oyj9109: Yeah I think we can leave them in `shouldShowAutoPrompt_` and seperate frequency capping related logic into a new function. And then in `showAutoPrompt_`, early exit the flow if `shouldShowAutoPrompt_` evaluates to `false` and proceed with frequency capping and the rest otherwise. Does it make sense?

justinchou-google: I refactored the `clientConfig` and `entitlements` out of the `shouldShowMonetizationPromptFromFrequencyCap` function. Perhaps we can keep the scope of this PR to this, and in another PR investigate if we can terminate early. WDYT?

src/runtime/auto-prompt-manager.ts
```diff
@@ -200,18 +201,23 @@ export class AutoPromptManager {
...
-      clientConfig,
-      entitlements,
-      params.autoPromptType
-    );
+    const shouldShowMonetizationPrompt =
+      this.shouldShowMonetizationPromptFromClientConfig(
+        clientConfig,
+        params.autoPromptType
+      ) &&
+      !entitlements.enablesThis() &&
```
oyj9109, line 209: I'm wondering if we can check entitlement at the top of `showAutoPrompt_` and exit the flow if there's a valid entitlement?

src/runtime/auto-prompt-manager.ts
```diff
@@ -200,18 +201,23 @@ export class AutoPromptManager {
       };
     }
 
-    const shouldShowAutoPrompt = await this.shouldShowAutoPrompt_(
-      clientConfig,
-      entitlements,
-      params.autoPromptType
-    );
+    const shouldShowMonetizationPrompt =
```
oyj9109, line 204: nit: canDisplayMonetizationPrompt?
It seems like `clientConfig.uiPredicates.canDisplayAutoPrompt` is used as `canDisplayMonetizationPrompt` but the name is pretty misleading to me.

src/runtime/auto-prompt-manager.ts
justinchou-google: Moved to the top and no tests fail. Also removed a redundant entitlements check from `shouldShowBlockingPrompt`. Which, do we want to rename to `canDisplayBlockingPrompt`?

src/runtime/auto-prompt-manager.ts
justinchou-google: Updated. One thing to note, the class variable is still kept as `wasMonetizationPromptDisplayedUncappedByFrequency_`, to differentiate from the case where a monetization prompt is displayed through the audience action flow. Is this ok to leave?

src/runtime/auto-prompt-manager.ts
```diff
@@ -271,22 +280,28 @@ export class AutoPromptManager {
...
+
+  /**
+   * Determines whether a mini prompt for contributions or subscriptions should
+   * be shown based on the frequency cap.
+   */
+  async shouldShowMonetizationPromptFromFrequencyCap(
+    autoPromptType?: AutoPromptType,
+    autoPromptConfig?: AutoPromptConfig
+  ): Promise<boolean> {
     // If the auto prompt type is not supported, don't show the prompt.
```
oyj9109, line 305: Would you update the comment with
1) `undefined` `autoPromptType` can be a valid publication state with premonetization.
2) We don't apply 'monetization' prompt frequency capping for 1) (for now)

src/runtime/auto-prompt-manager.ts
```diff
@@ -271,22 +280,28 @@ export class AutoPromptManager {
...
    * Determines whether a mini prompt for contributions or subscriptions should
-   * be shown.
+   * be shown based on the UI Predicates.
    */
-  async shouldShowAutoPrompt_(
-    clientConfig: ClientConfig,
-    entitlements: Entitlements,
-    autoPromptType?: AutoPromptType
-  ): Promise<boolean> {
+  shouldShowMonetizationPromptFromUiPredicates(
```
oyj9109, line 285: nit: would 'canDisplay' be more accurate name?

src/runtime/auto-prompt-manager.ts
```diff
@@ -271,22 +280,28 @@ export class AutoPromptManager {
...
     }
 
+    return true;
+  }
+
+  /**
+   * Determines whether a mini prompt for contributions or subscriptions should
+   * be shown based on the frequency cap.
+   */
+  async shouldShowMonetizationPromptFromFrequencyCap(
```
oyj9109, line 301: ditto. what about `canDisplayMonetizationPromptAfterFrequencyCapping`?

src/runtime/auto-prompt-manager.ts
```diff
@@ -295,11 +310,6 @@ export class AutoPromptManager {
       return Promise.resolve(false);
     }
 
-    // If we found a valid entitlement, don't show the prompt.
-    if (entitlements.enablesThis()) {
-      return Promise.resolve(false);
-    }
-
     // The auto prompt is only for non-paygated content.
```
oyj9109, line 313: nit: Frequency capping is only for?

src/runtime/auto-prompt-manager.ts
```diff
@@ -478,7 +483,10 @@ export class AutoPromptManager {
 
     // For subscriptions, skip triggering checks and use the first potential action
     if (this.isSubscription_({autoPromptType})) {
-      if (shouldShowAutoPrompt) {
+      if (
+        canDisplayMonetizationPrompt ||
```
oyj9109, line 487: The name `canDisplayMonetizationPrompt` seems not accurate in terms that it can be `false` when "we can display (blocking) subscription prompt".
Can we make the name more specific to frequency capping flow or dismissible prompt?

src/runtime/auto-prompt-manager.ts
justinchou-google: updated to `canDisplayMonetizationPromptAfterFrequencyCap_`

src/runtime/auto-prompt-manager.ts
justinchou-google: I believe this one actually means `frequency cap should suppress monetization prompt on non-paygated content`

justinchou-google: I think we should actually approach this from a different perspective, rather than `frequency cap`. `Frequency cap` implies that we _should_ be regularly showing some monetization prompt, but the frequency cap may occasionally suppress this (hence the current naming preference for `can` versus `should`).

It seems bit more accurate to describe the situation as, we should be showing whatever recommended audience action is first, but this suggestion suggests we should instead show a monetization prompt. Hence, for subscription flow, we usually have surveys and regwall first, and these functions say we _should_ display the subscription prompt. Similarly, for contributions, after the contribution prompt is dismissed initially, the recommended audience actions may be surveys and newsletter, but sometimes this function suggests we _should_ display the contribution propmt. Thoughts?

justinchou-google: (I think the `uiPredicates` can remain as `canDisplay`)

oyj9109: I see. Yeah I think that makes sense. (and still +1 to `canDisplay` for ui predicate.

oyj9109 (approved):

src/runtime/auto-prompt-manager.ts
```diff
@@ -270,76 +282,71 @@ export class AutoPromptManager {
...
+    return true;
+  }
 
+  /**
+   * Determines whether a monetization prompt should be shown as a paywall,
+   * meaning with the explicit intent to hard- or soft-restrict access to the
+   * page. Does not prevent a monetization prompt from displaying as an
+   * eligible audience action.
+   */
+  async shouldShowMonetizationPromptAsPaywall(
```
justinchou-google, line 305: Hi @ChrisAntaki , wanted to get your take on the naming of this function and boolean `monetizationPromptWasDisplayedAsPaywall_`.

TLDR: The term `Paywall` is used to convey the intent to restrict a reader's access to some content [old]. This is different from swg.js displaying a monetization prompt as an audience action provided from the article endpoint [newly added].

Previously, we computed whether to show a monetization prompt ONLY from a few factors: client config, entitlements, and any frequency caps (explicitly distinct from the audience actions). With our upcoming June launch, we are expanding monetization prompts to also be served as audience actions. As such, we want to now distinguish these two serving pathways; the idea being that the monetization prompts can be shown with the intent to paywall content either as a hard-restriction (non-dismissible prompt) or soft-restriction (dismissible prompt), vs. when a monetization prompt is the next eligible audience action. Note: all Paywalls are Monetization Prompts, but not all Monetization Prompts are Paywalls.

Thoughts on this terminology?

src/runtime/auto-prompt-manager.ts
```diff
@@ -270,76 +282,71 @@ export class AutoPromptManager {
...
       return Promise.resolve(false);
     }
 
-    // If we found a valid entitlement, don't show the prompt.
-    if (entitlements.enablesThis()) {
-      return Promise.resolve(false);
-    }
-
-    // The auto prompt is only for non-paygated content.
+    // For non-paygated content, a paywall should not restrict access.
```
oyj9109, line 319: nit: should it start with "For pay-gated content"?
E.g., 
// For pay-gated content, we may show blocking prompts rather than soft paywall.

The original comment is identical to "For paygated content, we dont' show auto prompt.

src/runtime/auto-prompt-manager.ts
oyj9109: Thanks for the great summary. I like the new name and distinction from audience action serving path in the comment.

Would it be more accurate to narrow down the name to "soft paywall"?
One of the condition is returning `false` when the page is locked, meaning it doesn't cover showing "hard-" paywall.

src/runtime/auto-prompt-manager.ts
justinchou-google: Yes, good catch. Fixed!

src/runtime/auto-prompt-manager.ts
ChrisAntaki: I'm liking this clear naming

src/runtime/auto-prompt-manager.ts
justinchou-google: Resolved all offline.

PR #3022, opened 2023-05-19
Basic events are now firing for Subscription Linking

everyplace:
This is a proposal for firing events for Subscription Linking.

To test this code, I modified publisher.js as in the following screenshot, but I didn't check this part in.

image

### DIFF
--- a/src/runtime/subscription-linking-flow.ts
+++ b/src/runtime/subscription-linking-flow.ts
@@ -16,6 +16,9 @@
 
 import {ActivityIframeView} from '../ui/activity-iframe-view';
 import {ActivityPorts} from '../components/activities';
+import {
+  AnalyticsEvent
+} from '../proto/api_messages';
 import {Deps} from './deps';
 import {DialogManager} from '../components/dialog-manager';
 import {
@@ -23,7 +26,9 @@ import {
   LinkSubscriptionResult,
 } from '../api/subscriptions';
 import {PageConfig} from '../model/page-config';
-import {SubscriptionLinkingCompleteResponse} from '../proto/api_messages';
+import {
+  SubscriptionLinkingCompleteResponse
+} from '../proto/api_messages';
 import {feArgs, feUrl} from './services';
 
 export class SubscriptionLinkingFlow {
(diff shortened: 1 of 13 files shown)


src/runtime/subscription-linking-flow.ts
```diff
@@ -16,21 +16,27 @@
...
+  SubscriptionLinkingCompleteResponse
+} from '../proto/api_messages';
 import {feArgs, feUrl} from './services';
 
 export class SubscriptionLinkingFlow {
   private readonly activityPorts_: ActivityPorts;
   private readonly win_: Window;
   private readonly pageConfig_: PageConfig;
   private readonly dialogManager_: DialogManager;
+  private readonly deps_: Deps;
```
ChrisAntaki, line 39: Instead of the new lines 39, 51, and 52, we can just update line 43 to make `deps_` [property parameter](https://www.typescriptlang.org/docs/handbook/2/classes.html#parameter-properties)

ChrisAntaki:
```
constructor(private readonly deps_: Deps) {
```

everyplace: Thanks for the pointer @ChrisAntaki, I'll try it out Monday.

everyplace: Updated.

src/runtime/subscription-linking-flow.ts
```diff
@@ -81,15 +88,27 @@ export class SubscriptionLinkingFlow {
...
+        })
+        .then(() => {
+          this.deps_
+            .eventManager()
+            .logSwgEvent(
+              AnalyticsEvent.IMPRESSION_SUBSCRIPTION_LINKING_COMPLETE
+            );
+        });
+    } catch (e) {
+      this.deps_
```
ChrisAntaki, line 108: Could you add a test to cover this?

everyplace: Done in `896db24`

ChrisAntaki (approved):

PR #3029, opened 2023-05-23
Add documentation for configured_ method in runtime.ts

fikus:
b/283811907

### DIFF
--- a/src/runtime/runtime.ts
+++ b/src/runtime/runtime.ts
@@ -210,11 +210,24 @@ export class Runtime implements SubscriptionsInterface {
     this.buttonApi_.init(); // Injects swg-button stylesheet.
   }
 
+  /**
+   * Returns a promise for the configured runtime. If requested, also begins
+   * configuring the runtime.
+   *
+   * NOTE: Begininning runtime configuration prematurely can interfere with
+   * publisher integrations (b/283281215). This method should only be called
+   * with `startConfiguringRuntime=true` when immediate configuration is
+   * required, after integrators have had an opportunity to begin configuration
+   * using the `init` method.
+   *
+   * @param startConfiguringRuntime Whether to begin configuring the runtime.
+   * @return A promise resolving to the configured runtime.
+   */
   private async configured_(
     startConfiguringRuntime: boolean
   ): Promise<ConfiguredRuntime> {
     if (!startConfiguringRuntime) {
-      // Configuration isn't necessary yet, so lets wait.
+      // Configuration isn't necessary yet, so let's wait.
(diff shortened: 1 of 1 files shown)


src/runtime/runtime.ts
```diff
@@ -168,10 +168,23 @@ export class Runtime implements SubscriptionsInterface {
...
+   * configuring the runtime.
+   *
+   * NOTE: Begininning runtime configuration prematurely can interfere with
+   * publisher integrations (b/283281215). This method should only be called
+   * with `startConfiguringRuntime=true` when immediate configuration is
+   * required, after integratros have had an opportunity to begin configuration
+   * using the the `init` method.
+   *
+   * @param startConfiguringRuntime Whether to begin configuring the runtime.
+   * @returns A promise resolving to the configured runtime.
```
ChrisAntaki, line 182: This line seems a little redundant since the method description starts with the same info

src/runtime/runtime.ts
```diff
@@ -168,10 +168,23 @@ export class Runtime implements SubscriptionsInterface {
...
   }
 
+  /**
+   * Returns a promise for the configured runtime. If requested, also begins
+   * configuring the runtime.
+   *
+   * NOTE: Begininning runtime configuration prematurely can interfere with
+   * publisher integrations (b/283281215). This method should only be called
+   * with `startConfiguringRuntime=true` when immediate configuration is
+   * required, after integratros have had an opportunity to begin configuration
```
ChrisAntaki, line 178: integratros > integrators

src/runtime/runtime.ts
```diff
@@ -168,10 +168,23 @@ export class Runtime implements SubscriptionsInterface {
...
+   * using the the `init` method.
```
ChrisAntaki, line 179: the the > the

fikus:
Undid the formatting changes, my editor was misconfigured.

src/runtime/runtime.ts
ChrisAntaki: I think the TypeScript style guide suggests using `@return` instead of `@returns`, but only when it adds information. In this case, is it adding information though?
https://google.github.io/styleguide/tsguide.html#redundant-comments

ChrisAntaki: I see what you mean about there being other details in the description though

ChrisAntaki (approved):

src/runtime/runtime.ts
fikus: You're right, updated to `@return`. The style guide says it _may_ be omitted where it is obvious, I don't think it hurts clarity to be explicit here.

PR #3033, opened 2023-05-24
Add e2e test configuration enabling all experiment flags

fikus:
Add nightwatch env that runs test with all flags enabled. It can be run in isolation with `npx gulp e2e --env=all_experiments_enabled` and is included in the e2e npm script.

The new test configuration currently fails, as expected, due to b/282759151. Will wait to merge until this is fixed.

b/283467774

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
   },
   "scripts": {
     "preinstall": "node build-system/check-package-manager.js",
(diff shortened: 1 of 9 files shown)


ChrisAntaki (approved):
Nice!

ChrisAntaki:
If the test names ended with `(+experiments)` when the `all_experiments_enabled` env was chosen, that could help make the logs more readable

fikus:
Tests pass after merging #3039.

PR #3035, opened 2023-05-25
Simplify usage of largePromptFn in AutoPromptManager

fikus:
No-op refactor to simplify AutoPromptManager:
- Move determination of largePromptFn to separate method to make showAutoPrompt_ slightly smaller.
- Remove displayLargePromptFn from ShowAutoPromptParams. This was only used by test paths.

### DIFF
--- a/src/runtime/auto-prompt-manager-test.js
+++ b/src/runtime/auto-prompt-manager-test.js
@@ -65,7 +65,6 @@ describes.realWin('AutoPromptManager', (env) => {
   let clientConfigManager;
   let clientConfigManagerMock;
   let storageMock;
-  let alternatePromptSpy;
   let miniPromptApiMock;
   let actionFlowSpy;
   let startSpy;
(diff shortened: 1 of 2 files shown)


src/runtime/auto-prompt-manager-test.js
```diff
@@ -1164,59 +1136,67 @@ describes.realWin('AutoPromptManager', (env) => {
...
-    });
-    expect(alternatePromptSpy).to.be.calledOnce;
-  });
-
-  it('should replace the subscription miniprompt with a large prompt if DISABLE_DESKTOP_MINIPROMPT is enabled and viewport is wider than 480px', async () => {
-    win./*OK*/ innerWidth = 500;
-    setExperiment(win, ExperimentFlags.DISABLE_DESKTOP_MINIPROMPT, true);
-    miniPromptApiMock.expects('create').never();
+  it(
+    'should replace the contribution miniprompt with a large prompt if ' +
```
ChrisAntaki, line 1144: Could we avoid splitting the names up like this? I noticed it leads to more indentation for the body of the tests
image

src/runtime/auto-prompt-manager.ts
```diff
@@ -284,11 +273,32 @@ export class AutoPromptManager {
     );
   }
 
+  /**
+   * Returns a function to show the appropriate monetization prompt,
+   * or undefined if the type of prompt cannot be determined.
+   */
+  private getLargePromptFn_(
```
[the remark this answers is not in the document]
fikus: I think that's a good suggestion, thanks. Updated.

src/runtime/auto-prompt-manager-test.js
fikus: I had thought the line length was part of the style guide but I guess that isn't the case. Reverting these changes so the formatting is unchanged.

justinchou-google (approved):
LGTM

PR #3036, opened 2023-05-25
AutoPromptManager: Remove getMiniPromptApi_

fikus:
Remove this method. It was intended to be overridden by subclasses but is now marked private. AutoPromptManager is not otherwise designed for subclassing and is never subclassed. Plus this removes a typo.


ChrisAntaki (approved):
Nice catch

PR #3081, opened 2023-06-30
Ensure GAA regwall event handlers are only added once

fikus:
Add check so multiple calls to GAAMeteringRegwall.show do not install duplicate event handlers. Speculative fix for b/281908268.

### DIFF
--- a/src/runtime/extended-access/gaa-metering-regwall.ts
+++ b/src/runtime/extended-access/gaa-metering-regwall.ts
@@ -437,6 +437,7 @@ export class GaaMeteringRegwall {
    * Logs button click events.
    */
   private static logButtonClickEvents_(): void {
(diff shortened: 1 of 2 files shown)


src/runtime/extended-access/gaa-metering-regwall.ts
```diff
@@ -435,6 +437,10 @@ export class GaaMeteringRegwall {
    * Logs button click events.
    */
   private static logButtonClickEvents_(): void {
+    if (loggingEventHandlersAdded) {
```
ChrisAntaki, line 440:
```suggestion
    // Only add events once.
    if (loggingEventHandlersAdded) {
```

ChrisAntaki (approved):

PR #3125, opened 2023-08-08
Set up Nightwatch VRT testing and add test for Classic buyflow

fikus:
Add Nightwatch VRT plugin for visual regression tests (screenshot tests). Add an initial test case for the Classic buyflow.

b/294885021, b/294868068

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,6 @@
     "@babel/core": "7.22.9",
     "@babel/eslint-parser": "7.22.9",
     "@babel/preset-env": "7.22.9",
-    "@nightwatch/vrt": "^3.0.0",
     "@typescript-eslint/eslint-plugin": "5.62.0",
     "@typescript-eslint/parser": "5.62.0",
     "ansi-colors": "4.1.3",
@@ -124,6 +123,7 @@
     "is-running": "2.1.0",
     "jsonwebtoken": "9.0.1",
     "nightwatch": "2.6.21",
+    "@nightwatch/vrt": "^3.0.0",
     "rollup-plugin-visualizer": "5.9.2",
     "vite": "4.4.7"
   },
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -34,6 +34,7 @@
     "@babel/core": "7.22.9",
     "@babel/eslint-parser": "7.22.9",
     "@babel/preset-env": "7.22.9",
+    "@nightwatch/vrt": "^3.0.0",
```
ChrisAntaki, line 37: Could this be in the `optionalDependencies` list, where the `nightwatch` dependency resides? That would keep the dependencies lighter for the unit tests GitHub action, and a normal `npm install` command would still install it

test/e2e/nightwatch.conf.js
```diff
@@ -57,6 +63,11 @@ module.exports = {
           'populate-client-config-classic',
         ],
       },
+      '@nightwatch/vrt': {
+        baseline_suffix: '_all_experiments_on',
```
ChrisAntaki, line 67: Can this also test with experiments off?

ChrisAntaki: Oh I see it does, nice

package.json
fikus: Hm, I added it to `devDependencies` based on the installation instructions. Test framework dependencies are a commonly cited example of when to use dev dependencies. IIUC, optionalDependencies are for dependencies that can be used by app code but are not strictly required.

https://docs.npmjs.com/cli/v9/configuring-npm/package-json

Our devDependencies list includes many other test packages like chai, karma, etc. I'm not sure why nightwatch and others would be in optionalDependences.

fikus: Got it, thanks. Updated.

ChrisAntaki: Thanks! Just noticed the `check` omits the optional deps too https://github.com/subscriptions-project/swg-js/blob/65ccb7d85cd8a0080fe90808a1941de3cf083263/.github/workflows/ci.yaml#L22

ChrisAntaki (approved):
Thanks for adding this

PR #3131, opened 2023-08-16
Add VRT tests for Basic and Classic contribution pages

fikus:
Also, set window size in test configuration.

b/294869126


ChrisAntaki (approved):

fikus:
Note: also configured a diff threshold of 0.5%. Differences in text antialiasing were causing CI runs to fail.

PR #3132, opened 2023-08-17
Upload VRT report for GitHub CI runs

fikus:
Upload VRT artifacts in GitHub CI runs so they are available for debugging.

The uploaded HTML will not be directly usable until https://github.com/nightwatchjs/nightwatch-vrt/pull/4 is released, but in the meantime the uploaded images can show why a test failed.

b/296120391

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -44,6 +44,15 @@ jobs:
       - name: E2E Tests
         run: npm run e2e
 
+      - name: Upload VRT report
+        if: always()
+        uses: actions/upload-artifact@v3
+        with:
+          name: vrt-report
+          path: |
+            .vrt
+            vrt-report
+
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -44,6 +44,15 @@ jobs:
       - name: E2E Tests
         run: npm run e2e
 
+      - name: Upload VRT report
+        if: always()
```
ChrisAntaki, line 48: Just curious, is this line necessary?

ChrisAntaki (approved):

.github/workflows/ci.yaml
fikus: I included it because it was in the sample config I found in the docs, my guess is that it causes the upload to run even if the previous test step failed.

PR #3133, opened 2023-08-17
Fix exit status of e2e command if default run fails

fikus:
Fix e2e command so it reports failure if default config fails, but all_experiments_enabled passes.

b/296404968

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
   },
   "scripts": {
     "preinstall": "node build-system/check-package-manager.js",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -16,7 +16,7 @@
   },
   "scripts": {
     "preinstall": "node build-system/check-package-manager.js",
-    "e2e": "npx vite build -- --target=classic && npx vite build -- --target=gaa && npx vite build -- --target=basic && npx gulp e2e ; npx gulp e2e --env=all_experiments_enabled",
```
ChrisAntaki, line 19: Why not just an && instead of a ;   ?

ChrisAntaki (approved):

PR #3157, opened 2023-08-24
Determines moentization prompt eligibility using audience action instead of UiPredicate

oyj9109:
Context: b/288962599#comment6

We need to remove reference to `canDisplayAutoPrompt` in order to correct its conditions. Replacing it with checking the precense of audienc actions is identical and it's actually what's meant to be checked there.


src/runtime/auto-prompt-manager-test.js
```diff
@@ -1320,15 +1320,13 @@ describes.realWin('AutoPromptManager', (env) => {
...
-      actionType: 'TYPE_CONTRIBUTION',
       autoPromptType: AutoPromptType.CONTRIBUTION,
     },
     {
-      actionType: 'TYPE_SUBSCRIPTION',
       autoPromptType: AutoPromptType.SUBSCRIPTION,
     },
-  ].forEach(({actionType, autoPromptType}) => {
-    it(`should not display any prompt if UI predicate is false and page is locked for autoPromptType: ${autoPromptType}`, async () => {
+  ].forEach(({autoPromptType}) => {
```
fikus, line 1328: Would this test be more specific to the change being made if lines 1351 and 1352 were also changed to true?

src/runtime/auto-prompt-manager.ts
```diff
@@ -308,18 +305,16 @@ export class AutoPromptManager {
   }
 
   /**
-   * Determines whether a mini prompt for contributions or subscriptions can
-   * be shown based on the UI Predicates.
+   * Determines whether moentization prompt can be shown based audience actions
```
fikus, line 308: nit: "based on"?

fikus (approved):

src/runtime/auto-prompt-manager-test.js
oyj9109: Yes, good suggestion.

PR #3177, opened 2023-09-08
Allow AutoPromptManager to branch into rewarded ads

mhkawano:
This change adds a flag for rewarded ads, filters the rewarded ads based on the existence of the flag and gpt.js, and then adds a branch to call the new rewarded ads flow.

### DIFF
--- a/src/runtime/auto-prompt-manager.ts
+++ b/src/runtime/auto-prompt-manager.ts
@@ -25,6 +25,7 @@ import {
   AudienceActionIframeFlow,
   AudienceActionIframeParams,
 } from './audience-action-flow';
+import {AudienceActionLocalFlow} from './audience-action-local-flow';
 import {AutoPromptConfig} from '../model/auto-prompt-config';
 import {AutoPromptType} from '../api/basic-subscriptions';
 import {ClientConfig} from '../model/client-config';
@@ -48,6 +49,7 @@ import {isExperimentOn} from './experiments';
 const TYPE_CONTRIBUTION = 'TYPE_CONTRIBUTION';
 const TYPE_SUBSCRIPTION = 'TYPE_SUBSCRIPTION';
 const TYPE_REWARDED_SURVEY = 'TYPE_REWARDED_SURVEY';
+const TYPE_REWARDED_ADS = 'TYPE_REWARDED_AD';
 const SECOND_IN_MILLIS = 1000;
 
 const impressionEvents = [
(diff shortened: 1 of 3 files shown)


src/runtime/experiment-flags.ts
```diff
@@ -36,4 +36,9 @@ export enum ExperimentFlags {
    * Experiment flag to enable populating the client config in the Classic runtime.
    */
   POPULATE_CLIENT_CONFIG_CLASSIC = 'populate-client-config-classic',
+
+  /**
+   * Experiment for allowing rewarded ads to be rendered.
+   */
+  REWARDED_ADS_ENABLED = 'rewarded_ads_enabled',
```
fikus, line 43: Can this be controlled by a backend flag instead?

src/runtime/auto-prompt-manager.ts
```diff
@@ -550,21 +552,39 @@ export class AutoPromptManager {
...
     isClosable?: boolean;
   }): () => void {
-    return () => {
-      const params: AudienceActionIframeParams = {
-        action,
-        configurationId,
-        autoPromptType,
-        onCancel: () => this.storeLastDismissal_(),
-        isClosable,
+    if (action == TYPE_REWARDED_ADS) {
```
fikus, line 555: nit: in context, it appears `actionType` or similar would be a clearer name for this variable.

fikus, line 555: nit: ===

src/runtime/auto-prompt-manager.ts
```diff
@@ -550,21 +552,39 @@ export class AutoPromptManager {
...
+          onCancel: () => this.storeLastDismissal_(),
+          isClosable,
+        };
+        const lastAudienceActionFlow = new AudienceActionIframeFlow(
+          this.deps_,
+          params
+        );
+        this.setLastAudienceActionFlow(lastAudienceActionFlow);
+        lastAudienceActionFlow.start();
+      };
```
fikus, line 586: Consider extracting the common logic between the two blocks, it may improve clarity by highlighting what is different in the two branches.

src/runtime/experiment-flags.ts
fikus: If we do need a client experiment, see the note on L20 about adding the flag to test configurations.

mhkawano: We can filter out the action in the article endpoint, but I figured it would be better to maintain the logic in one place. Is there a reason we don't want the flag in the client?

fikus: Client flags are generally more difficult to work with. They aren't integrated as well with upstream systems and they require binary pushes to change. I also find it cleaner to implement the business logic in the server where possible, instead of sending data the client may ignore. In this case I think using a server-side flag would still result in maintaining the logic in a single location.

src/runtime/auto-prompt-manager.ts
mhkawano: Done.

src/runtime/auto-prompt-manager.ts
mhkawano: Done

mhkawano: Done.

src/runtime/experiment-flags.ts
mhkawano: Got it, I'll keep the flag out.

src/runtime/auto-prompt-manager.ts
```diff
@@ -770,6 +780,10 @@ export class AutoPromptManager {
         GoogleAnalyticsEventListener.isGtagEligible(this.deps_) ||
         GoogleAnalyticsEventListener.isGtmEligible(this.deps_);
       return isSurveyEligible && isAnalyticsEligible;
+    } else if (actionType == TYPE_REWARDED_ADS) {
```
fikus, line 783: nit: ===

fikus (approved):

PR #3183, opened 2023-09-15
Expand Impression local storage logging for Frequency Capping flow.

justinchou-google:
b/299846260

Add additional local storage handling for impressions of contributions (mini and large), newsletter, regwall, and survey prompts for frequency capping flow, guarded by experiment flag. Reuses some existing checks for local storage: 
1) Impressions on paygated content do not count towards frequency cap.
-  Manually triggered prompts (CTA) now count towards impressions.
2) A mini contribution prompt triggering a large contribution prompt should only count as one impression.

Enumerates the local storage keys as `imp_contribution`, `imp_newsletter`, `imp_regwall`, `imp_survey` since these are the only prompts applicable for frequency capping (in DD, originally planned to dynamically generate key based on action type).

https://docs.google.com/document/d/1H99mlR47aZZv_ZsCRkI6Iv_WUcpwaoeGgCExWSkdYCI/edit?resourcekey=0-9djhO0eRtzh5bMMrZ1dtBg#heading=h.mz8uos7t3l1t


src/runtime/auto-prompt-manager.ts
```diff
@@ -727,6 +752,37 @@ export class AutoPromptManager {
...
 
+  private async handleFrequencyCappingLocalStorage_(
+    analyticsEvent: AnalyticsEvent
+  ): Promise<void> {
+    // Impressions and dimissals of forced (for paygated) or manually triggered
+    // prompts do not count toward the frequency caps. Only monetization events
+    // can be triggered manually.
+    const promptTriggeredManually =
+      (monetizationImpressionEvents.includes(analyticsEvent) &&
+        !this.monetizationPromptWasDisplayedAsSoftPaywall_) ||
```
oyj9109, line 763: `shouldShowMonetizationPromptAsSoftPaywall` is determined based on the old frequency capping logic that will be replaced. I'm hoping we can greatly simplify the overall auto-prompt logic with this frequency improvement, while it would be pretty complicated if we continue to rely on the old path.

If I understand it correctly, the manually triggered path is for the button CTA scenario where it doesn't seem common to be mix-used with the regular auto prompting use case. Would it be reasonable or acceptable to count and use those impressions for the new frequency capping flow to significantly simpliy the logic?

src/utils/constants.ts
```diff
@@ -85,4 +85,15 @@ const StorageKeys = {
...
+/**
+ * Local storage keys for intervention impressions.
+ */
+const ImpressionStorageKeys = {
+  CONTRIBUTION: 'impressions_contribution',
+
+  NEWSLETTER: 'impressions_newsletter',
+
+  SURVEY: 'impressions_survey',
+};
```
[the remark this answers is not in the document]
justinchou-google: +1, will expand to include regwall.

src/runtime/auto-prompt-manager.ts
justinchou-google: Yes, the manual trigger is for CTA. I think it makes sense to include the CTA in the impressions, reducing this check to just the page lock.

src/runtime/auto-prompt-manager.ts
```diff
@@ -727,6 +752,37 @@ export class AutoPromptManager {
     }
   }
 
+  private async handleFrequencyCappingLocalStorage_(
+    analyticsEvent: AnalyticsEvent
+  ): Promise<void> {
+    // Impressions and dimissals of forced (for paygated) or manually triggered
+    // prompts do not count toward the frequency caps. Only monetization events
+    // can be triggered manually.
+    const promptTriggeredManually =
```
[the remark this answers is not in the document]
justinchou-google: see below.

src/runtime/auto-prompt-manager.ts
justinchou-google: Removed cta check, will need to revisit before phase 1 under new triggering flow. Added tracking todo bug: b/300963305.

src/utils/constants.ts
```diff
@@ -85,4 +85,17 @@ const StorageKeys = {
   TOAST: 'toast',
 };
 
-export {Constants, StorageKeys};
+/**
+ * Local storage keys for intervention impressions.
+ */
+const ImpressionStorageKeys = {
```
oyj9109, line 91: nit: can we use consistent type names as https://github.com/subscriptions-project/swg-js/blob/827e4418685a75ab0402a3ca15cfc0a010840890/src/runtime/audience-action-flow.ts#L67?
REGISTRATION_WALL, NEWSLETTER_SIGNUP, REWARDED_SURVEY, REWARDED_AD

src/utils/constants.ts
```diff
@@ -85,4 +85,17 @@ const StorageKeys = {
...
+  CONTRIBUTION: 'impressions_contribution',
```
oyj9109, line 92: nit: The key string does not need to be completely descriptive, but it would be a little bit beneficial to shorten it. What about using the shorter abbreviation "imp_" instead of "impressions_"?

src/runtime/auto-prompt-manager.ts
```diff
@@ -727,6 +753,33 @@ export class AutoPromptManager {
...
+    // frequency caps. TODO(b/300963305): manually triggered prompts should
+    // also be excluded.
+    if (
+      !INTERVENTION_TO_STORAGE_KEY_MAP.has(analyticsEvent) ||
+      this.pageConfig_.isLocked()
+    ) {
+      return;
+    }
+
+    const storageKey = INTERVENTION_TO_STORAGE_KEY_MAP.get(analyticsEvent)!;
```
oyj9109, line 769: nit: this can be moved below if statement or inlined in the return line.

src/runtime/auto-prompt-manager.ts
```diff
@@ -727,6 +753,33 @@ export class AutoPromptManager {
...
+    if (monetizationImpressionEvents.includes(analyticsEvent)) {
+      // Prompt impression should be stored if no previous one has been stored.
+      // This is to prevent the case that user clicks the mini prompt, and both
+      // impressions of the mini and large prompts would be counted towards the
+      // cap.
+      if (this.hasStoredImpression_) {
+        return;
+      }
+      this.hasStoredImpression_ = true;
```
oyj9109, line 778: should this be renamed to be specific to monetization impression?

src/runtime/auto-prompt-manager.ts
```diff
@@ -695,6 +712,15 @@ export class AutoPromptManager {
       );
     }
 
+    // ** Frequency Capping Impressions **
+    const enableFrequencyCappingLocalStorage = isExperimentOn(
+      this.doc_.getWin(),
```
oyj9109, line 717: Isn't the flag set up as mendel flag propagated via article response? I think this function is for checking client-only flag

src/runtime/auto-prompt-manager-test.js
```diff
@@ -443,6 +447,170 @@ describes.realWin('AutoPromptManager', (env) => {
...
+
+    await eventManagerCallback({
+      eventType: AnalyticsEvent.IMPRESSION_CONTRIBUTION_OFFERS,
+      eventOriginator: EventOriginator.UNKNOWN_CLIENT,
+      isFromUserAction: null,
+      additionalParameters: null,
+    });
+  });
+
+  it('should set frequency cap local storage if experiment is enabled and monetization prompts was triggered as a soft paywall', async () => {
```
oyj9109, line 478: Can we not use the notion of soft paywall in the new flow and just test it between locked and unlocked?

src/utils/constants.ts
justinchou-google: Sounds good, is `REWARDED_AD` something we want to add now?

src/utils/constants.ts
justinchou-google: Done, will update DD too.

src/runtime/auto-prompt-manager.ts
justinchou-google: Good catch.

src/runtime/auto-prompt-manager-test.js
justinchou-google: Yes, these needed to be updated along with the other change.

src/runtime/auto-prompt-manager.ts
justinchou-google: Added `hasStoredMiniPromptImpression_`.

src/runtime/auto-prompt-manager.ts
justinchou-google: Updated to check for article experiment.

src/runtime/auto-prompt-manager-test.js
```diff
@@ -443,6 +447,257 @@ describes.realWin('AutoPromptManager', (env) => {
     });
   });
 
+  it('should not set frequency cap local storage if experiment is enabled and content is locked', async () => {
```
oyj9109, line 450: Can you add one more case to cover when experiment is not enabled?

oyj9109 (approved):

src/utils/constants.ts
justinchou-google: Will add the enum now but not the event mapping as the Impression enum for rewarded ads has not yet been created.

src/runtime/auto-prompt-manager-test.js
justinchou-google: Added test where experiment is explicitly disabled.

PR #3193, opened 2023-09-20
Use completeaudienceaction endpoint for Rewarded Ads.

mhkawano:


src/runtime/audience-action-local-flow.ts
```diff
@@ -44,6 +47,8 @@ export interface AudienceActionLocalParams {
   monetizationFunction?: () => void;
 }
 
+const TYPE_REWARDED_AD = 'TYPE_REWARDED_AD';
```
oyj9109, line 50: nit: Can we define it in a centralized place? 
It would be awesome if you could update other parts, e.g., https://github.com/subscriptions-project/swg-js/blob/09df5f3068bcbff8d010eccfe7e679035fd09844/src/runtime/audience-action-flow.ts#L67. Thanks!

src/runtime/audience-action-local-flow.ts
```diff
@@ -324,6 +330,37 @@ export class AudienceActionLocalFlow implements AudienceActionFlow {
...
+      // TODO: mhkawano - configurationId should not be optional
+      ['configurationId', this.params_.configurationId!],
+      ['audienceActionType', this.params_.action],
+    ];
+    const url = queryParams.reduce(
+      (url, [param, value]) => addQueryParam(url, param, value),
+      serviceUrl(baseUrl)
+    );
+    this.deps_.win().fetch(url, init);
+    // TODO: mhkawano - log error
```
oyj9109, line 361: Maybe add a TODO to handle entitlement consumption logic on completion for paid content?

https://github.com/subscriptions-project/swg-js/blob/09df5f3068bcbff8d010eccfe7e679035fd09844/src/runtime/audience-action-flow.ts#L192

oyj9109 (approved):
LGTM. Just left a few nit comments.

src/runtime/audience-action-local-flow.ts
mhkawano: Done.

src/runtime/audience-action-local-flow.ts

mhkawano: Could use more followup cleanup, but will keep that for another cl.

PR #3195, opened 2023-09-21
Add click behavior to grey background

mborof:
Adds an option to suppress clicks or to close the dialog onclick.

See b/234490590 for more information.  This PR doesn't change any behavior, it instead provides the option for future PRs to adjust behavior when calling the dialogs.


fikus (commented):
- This includes bad formatting changes, please double check your editor settings.
- Please link to a bug that explains the functionality change being implemented.

src/components/dialog.ts
```diff
@@ -422,6 +433,14 @@ export class Dialog {
     this.hidden_ = false;
   }
 
+  /** Supresses click events, unless the window is closable. */
+  onGrayPaneClick_() {
```
fikus, line 437: method should be marked private.

src/components/dialog.ts
```diff
@@ -82,6 +82,7 @@ export interface DialogConfig {
   maxAllowedHeightRatio?: number;
   iframeCssClassOverride?: string;
   shouldDisableBodyScrolling?: boolean;
+  isClosable?: boolean;
```
fikus, line 85: add documentation for new property in above comment.

fikus, line 85: consider naming the property more specifically with the behavior it affects.

src/components/dialog-test.js
```diff
@@ -411,6 +411,86 @@ describes.realWin('Dialog', (env) => {
...
+        const el = dialog.graypane_.getElement();
+        const openedDialog = await dialog.open(NO_ANIMATE);
+        await openedDialog.openView(view);
+
+        el.click();
+
+        expect(wasClicked).to.be.true;
+
+        Dialog.prototype.onGrayPaneClick_ = old;
+      });
```
fikus, line 451: Whether a click event is registered is an implementation detail. The important behavior to verify is whether clicks are blocked/the dialog is closed.

src/components/dialog-test.js
```diff
@@ -411,6 +411,86 @@ describes.realWin('Dialog', (env) => {
...
+          {isClosable: false, shouldDisableBodyScrolling: true}
+        );
+        const el = dialog.graypane_.getElement();
+        const openedDialog = await dialog.open(NO_ANIMATE);
+        await openedDialog.openView(view);
+        expect(doc.body).to.have.class('swg-disable-scroll');
+
+        el.click();
+
+        expect(doc.body).to.have.class('swg-disable-scroll');
```
fikus, line 472: Checking the presence of the swg-disable-scroll class seems like a roundabout way to verify this behavior. Can you spy on close() instead?

src/components/dialog-test.js
```diff
@@ -411,6 +411,86 @@ describes.realWin('Dialog', (env) => {
...
+        el.click();
+
+        expect(wasClicked).to.be.true;
+
+        Dialog.prototype.onGrayPaneClick_ = old;
+      });
+
+      it('should have a click event that returns false', async () => {
+        dialog = new Dialog(globalDoc, {});
+        expect(dialog.onGrayPaneClick_()).to.be.false;
```
fikus, line 455: Similar here, we should be able to generate a click event and check the behavior, instead of calling a private method.

fikus (changes requested):

src/components/dialog-test.js
mborof: When I started adding spies it broke event registration and stopped calling the onclick function.  I'm not saying it can't be done, just that everything I tried was broken until I turned off the spies.

src/components/dialog-test.js
mborof: I'm not sure how to check the behavior of return false any other way.  Is there a way to get the return value of a click event?

mborof: This is also why I have it validating that function is the one being called.

src/components/dialog-test.js
mborof: Fixed

src/components/dialog-test.js
mborof: Fixed

src/components/dialog-test.js
```diff
@@ -411,6 +411,88 @@ describes.realWin('Dialog', (env) => {
       expect(styleDuringInit).to.equal('display: none !important;');
     });
 
+    describe('fade background', () => {
+      it('should not suppress click events by default', async () => {
+        let wasClicked = false;
+        const clickFun = function () {
+          wasClicked = true;
+        };
```
fikus, line 419: Can you use sandbox.spy() for these functions?

src/components/dialog-test.js
fikus: Can you add comments explaining that the swg-disable-scroll class is indicating whether the dialog is closed?

src/components/dialog.ts
```diff
@@ -422,6 +435,15 @@ export class Dialog {
     this.hidden_ = false;
   }
 
+  /** Supresses click events, unless the window is closable. */
```
fikus, line 438: This could be clarified, as the implementation always stops click events, but if the dialog is closeable it closes the dialog on click.

fikus, line 438: nit: Suppresses (spelling).

fikus (approved):

src/components/dialog-test.js
mborof: Any usage of spy seemed to break event registration so the click button no longer worked.

PR #3206, opened 2023-09-25
Bring back UiPredicate check for canDisplayAutoPrompt

oyj9109:
Context: b/288962599

This is the last step for fixing conditions for `canDisplayAutoPrompt`.
Step 1. Remove referece to the UiPredicate for monetization eligibilty check: https://github.com/subscriptions-project/swg-js/pull/3157 
Step 2. Remove offer related condition to correct its definition: cl/566416837
Step 3 (This PR). Bring it back as the top-level eligiblity check for auto prompting.


justinchou-google (approved):
Nice! Seems like we are finally redefining autoprompt to be any generic prompt that automatically triggers, not specific to a monetization prompt.

src/runtime/auto-prompt-manager.ts
```diff
@@ -249,6 +249,10 @@ export class AutoPromptManager {
       return;
     }
 
+    if (!clientConfig.uiPredicates?.canDisplayAutoPrompt) {
```
justinchou-google, line 252: nit: do we want to move this to before the entitlements check to save us an entitlement check in this scenario?

oyj9109: `entitlements.enablesThis()` is actually just checking the presence of entitlement, but doesn't perform any action on it. That being said, I think it makes sense to check UiPredicate first which is per-publication configuration rather than per-user data.

oyj9109:
> Nice! Seems like we are finally redefining autoprompt to be any generic prompt that automatically triggers, not specific to a monetization prompt.

Yay! thanks for review

PR #3211, opened 2023-09-26
Log SwgClientEvents for Frequency Capping flow.

justinchou-google:
b/295568021

SwgClientLog events added in #3210 .

https://docs.google.com/document/d/1H99mlR47aZZv_ZsCRkI6Iv_WUcpwaoeGgCExWSkdYCI/edit?resourcekey=0-9djhO0eRtzh5bMMrZ1dtBg#heading=h.gmv5a3u4fkk2


src/runtime/auto-prompt-manager.ts
```diff
@@ -997,7 +1023,16 @@ export class AutoPromptManager {
    */
   private async getActionImpressions_(actionType: string): Promise<number[]> {
     if (!ACTION_TO_IMPRESSION_STORAGE_KEY_MAP.has(actionType)) {
-      // TODO: handle missing storage key for action
+      this.eventManager_.logEvent({
```
oyj9109, line 1026: nit: can we use `logSwgEvent`?

src/runtime/auto-prompt-manager.ts
```diff
@@ -997,7 +1023,16 @@ export class AutoPromptManager {
...
+        eventType:
+          AnalyticsEvent.EVENT_ACTION_IMPRESSIONS_STORAGE_KEY_NOT_FOUND_ERROR,
+        eventOriginator: EventOriginator.SWG_CLIENT,
+        isFromUserAction: false,
+        additionalParameters: {
+          publicationid: this.pageConfig_.getPublicationId(),
+          actionType,
```
oyj9109, line 1033: I don't think this support arbitrary parameters like this. Do they need to be registered in the proto: http://google3/logs/proto/subscribewithgoogle/client/subscribewithgoogle_client_extension.proto;l=602;rcl=568642159?

src/runtime/auto-prompt-manager.ts
```diff
@@ -997,7 +1023,16 @@ export class AutoPromptManager {
...
   private async getActionImpressions_(actionType: string): Promise<number[]> {
     if (!ACTION_TO_IMPRESSION_STORAGE_KEY_MAP.has(actionType)) {
-      // TODO: handle missing storage key for action
+      this.eventManager_.logEvent({
+        eventType:
+          AnalyticsEvent.EVENT_ACTION_IMPRESSIONS_STORAGE_KEY_NOT_FOUND_ERROR,
+        eventOriginator: EventOriginator.SWG_CLIENT,
+        isFromUserAction: false,
+        additionalParameters: {
+          publicationid: this.pageConfig_.getPublicationId(),
```
oyj9109, line 1032: Isn't publication_id already logged by default?
http://google3/logs/proto/subscribewithgoogle/client/subscribewithgoogle_client_extension.proto;l=47;rcl=568642159

oyj9109 (approved):

src/runtime/auto-prompt-manager.ts
justinchou-google: I see, so I can exclude this?

src/runtime/auto-prompt-manager.ts
justinchou-google: This was my initial impression, but I saw some other usages of `additionalParameters` so I got optimistic... I will remove these parameters.

src/runtime/auto-prompt-manager.ts
oyj9109: Yes, it seems unnecessary. anyway it won't be logged.

PR #3216, opened 2023-09-27
Call getactionconfigurationui endpoint for Rewarded Ads

mhkawano:
Screenshot:
https://github.com/subscriptions-project/swg-js/assets/106177651/fd36afca-fc62-4d6e-9fb0-12e75756614f


src/runtime/audience-action-local-flow.ts
```diff
@@ -115,107 +119,115 @@ export class AudienceActionLocalFlow implements AudienceActionFlow {
...
-    // Replace with error view if init fails.
-    // TODO: mhkawano - Make closeable.
-    if (!initSuccess) {
-      this.makeErrorView(this.prompt_);
+  private async initPrompt_() {
+    if (this.params_.action === TYPE_REWARDED_AD) {
+      const config = await this.getConfig_();
+      // Setup callback for googletag init.
+      const googletag = this.deps_.win().googletag;
+      googletag.cmd.push(() => {
```
oyj9109, line 147: How would the prompt look like if this callback execution is delayed? Should we handle the case when it's never executed?

src/runtime/audience-action-local-flow.ts
```diff
@@ -115,107 +119,115 @@ export class AudienceActionLocalFlow implements AudienceActionFlow {
...
-
-    // Replace with error view if init fails.
-    // TODO: mhkawano - Make closeable.
-    if (!initSuccess) {
-      this.makeErrorView(this.prompt_);
+  private async initPrompt_() {
+    if (this.params_.action === TYPE_REWARDED_AD) {
+      const config = await this.getConfig_();
+      // Setup callback for googletag init.
+      const googletag = this.deps_.win().googletag;
```
oyj9109, line 146: Is it guaranteed that `googletag` object has been initialized?
You may want to add this line somewhere to conditionally initiated it 
`window.googletag = window.googletag || {cmd: []};`

src/runtime/audience-action-local-flow.ts
```diff
@@ -115,107 +119,115 @@ export class AudienceActionLocalFlow implements AudienceActionFlow {
...
+    } else {
+      this.renderErrorView_();
     }
   }
 
-  private initGpt_(resolve: (result: boolean) => void) {
-    // Save resolve so that it can be called in rewardedSlotReady_.
-    this.rewardedResolve_ = resolve;
+  private initRewardedAdWall_(config: AudienceActionConfig | null): boolean {
+    const validRewardedAdParams =
```
oyj9109, line 159: Can this validation be done before the config is passed so that it's caught earlier?

src/runtime/audience-action-local-flow.ts
```diff
@@ -352,9 +362,13 @@ export class AudienceActionLocalFlow implements AudienceActionFlow {
...
-  private async complete_() {
+  private async fetch_(
+    method: string,
+    endpoint: string,
+    queryParams: string[][]
+  ): Promise<Response> {
     const init: RequestInit = {
-      method: 'POST',
+      method,
       headers: {
```
oyj9109, line 372: Out of curiosity, do we need to manually set headers like this? I wonder if we can use (or extend) the existing method provided by fetcher.ts

src/runtime/audience-action-local-flow.ts
mhkawano: We check that the object has been initialized in AutoPromptManager.checkActionEligibility_, so this should be fine.

src/runtime/audience-action-local-flow.ts
mhkawano: I didn't wanna use fetcher for GET because it would hide the response, making it harder to catch errors for my logging. Didn't use it for POST for consistency, but I think we could use it.

src/runtime/audience-action-local-flow.ts
mhkawano: Done. Moved the validation to right after the fetch.

src/runtime/audience-action-local-flow.ts
mhkawano: Moved the timeout up a level so that it should catch the case where it's not called.

src/runtime/audience-action-local-flow.ts
oyj9109: Would you elaborate more on "GET would hide the response".
I wonder if we could either
- Provide a get method from the fetcher interface
- Or at least put header defined in the centralized place 

It will ensure all the request specs are consistently updated in the future.

oyj9109 (approved):
Have one question left about the fetcher method, but other than that looks great!

PR #3222, opened 2023-09-29
Update timestamp lifespan to two weeks for frequency capping events

justinchou-google:
b/302710324

Currently, timestamps older than 1 week are pruned upon setting and getting event timestamps. Frequency Capping supports up to 2 weeks, this change extends the lifespan for timestamps up to 2 weeks only for frequency capping events. When the old trigger flow is fully deprecated, these new functions will replace the legacy functionality.


src/runtime/storage.ts
```diff
@@ -101,12 +103,39 @@ export class Storage {
...
+   * Retrieves timestamps from local storage, under the storageKey provided
+   * for the frequency capping flow. To replace the legacy getEvent fn when the
+   * legacy triggering flow is deprecated. Prunes timestamps older than two
+   * weeks.
+   */
+  async getFrequencyCappingEvent(storageKey: string): Promise<number[]> {
+    const value = await this.get(storageKey, /* useLocalStorage */ true);
+    return this.pruneTimestamps_(
+      this.deserializeTimestamps_(value),
+      TWO_WEEKS_IN_MILLIS
```
oyj9109, line 136: Can you add a comment that 2 weeks is not a technical limitation, but based on the product requirement?

oyj9109 (approved):

PR #3224, opened 2023-10-02
Update entitlements for rewarded ads on complete.

mhkawano:
Follow the same steps as an iframe audience action.

  On a successful response from the dialog, we should:
  1) Store the updated user token
  2) Clear existing entitlements from the page
  3) Update READ_TIME in local storage to indicate that entitlements may have changed recently
  4) Re-fetch entitlements which may potentially provide access to the page


src/runtime/audience-action-local-flow.ts
```diff
@@ -439,10 +449,24 @@ export class AudienceActionLocalFlow implements AudienceActionFlow {
...
     };
-    this.fetcher_.sendPost(url, emptyMessage);
-    // TODO: mhkawano - log error
-    // TODO: mhkawano - handle entitlement consumption logic on completion
-    // ex: this.entitlementsManager_.getEntitlements();
+    const response = (await this.fetcher_.sendPost(
+      url,
+      emptyMessage
+    )) as unknown as CompleteAudienceActionResponse;
+    if (response.updated) {
```
fikus, line 456: Add test for the case that updated is false?

src/runtime/audience-action-local-flow.ts
```diff
@@ -439,10 +449,24 @@ export class AudienceActionLocalFlow implements AudienceActionFlow {
...
+      this.entitlementsManager_.clear();
+      if (response.swgUserToken) {
+        await this.deps_
+          .storage()
+          .set(Constants.USER_TOKEN, response.swgUserToken, true);
```
fikus, line 461: Add test?

src/runtime/audience-action-local-flow.ts
```diff
@@ -439,10 +449,24 @@ export class AudienceActionLocalFlow implements AudienceActionFlow {
...
+      }
+      const now = Date.now().toString();
+      await this.deps_
+        .storage()
+        .set(Constants.READ_TIME, now, /*useLocalStorage=*/ false);
+      await this.entitlementsManager_.getEntitlements();
```
fikus, line 467: Add test?

src/runtime/audience-action-local-flow.ts
mhkawano: Done.

src/runtime/audience-action-local-flow.ts
mhkawano: Done.

src/runtime/audience-action-local-flow.ts
mhkawano: Done.

src/runtime/audience-action-local-flows-test.js
```diff
@@ -391,6 +417,76 @@ describes.realWin('AudienceActionLocalFlow', (env) => {
...
+        expect(entitlementsManager.clear).to.not.be.called;
+        expect(entitlementsManager.getEntitlements).to.not.be.called;
+        expect(env.win.localStorage.setItem).to.not.be.calledWith(
+          'subscribe.google.com:USER_TOKEN',
+          'xyz'
+        );
+        expect(env.win.sessionStorage.setItem).to.not.be.calledWith(
+          'subscribe.google.com:READ_TIME'
+        );
+      });
```
fikus, line 460: I think these can also use `.to.not.be.called` so any calls with other arguments would also fail the tests

src/runtime/audience-action-local-flows-test.js
```diff
@@ -391,6 +417,76 @@ describes.realWin('AudienceActionLocalFlow', (env) => {
...
+
+      it('does not update token if non returned', async () => {
```
fikus, line 462: Typo "non"?

fikus (approved):

src/runtime/audience-action-local-flows-test.js
mhkawano: Done.

src/runtime/audience-action-local-flows-test.js
mhkawano: Fixed.

PR #3227, opened 2023-10-04
Render newsletter byop in swg.js

kristenwang:
Verified in qual that the code snippet could be correctly rendered: https://screenshot.googleplex.com/BH2FiGVUBZFJFeh.

Will follow up with prs to log events and put the prompt in the correct place.


src/runtime/audience-action-local-flow.ts
```diff
@@ -161,11 +173,64 @@ export class AudienceActionLocalFlow implements AudienceActionFlow {
...
     } else {
+      this.eventManager_.logSwgEvent(
+        AnalyticsEvent.EVENT_BYOP_NEWSLETTER_OPT_IN_CONFIG_ERROR
+      );
       this.renderErrorView_();
     }
   }
 
+  private renderOptinPrompt_(codeSnippet: string) {
+    if (!codeSnippet.includes('form')) {
```
fikus, line 206: Consider moving HTML validation to a separate method.

fikus, line 206: Rather than: 
1. Running a naive string check for 'form'
2. Setting innerHTML on the prompt
3. Checking for a <form> element

Would it make more sense to validate this once, eg by creating a new div, setting innerHTML, and checking its structure, before modifying prompt_?

kristenwang: Done! Checking for a <form> element is required for prompt_ since the result of querySelector() could be null, even if we check the <form> element in another div, so I had to replace prompt_ if we only want to check for 'form' once.

kristenwang: Resolving this comment since this is deleted!

fikus (approved):

PR #3228, opened 2023-10-04
Extend Frequency Capping flow to locked content (not openaccess) on non-subscription publications

justinchou-google:
b/303458595 Frequency capping triggering flow for non-openaccess posts on premonetization, contribution publications should behave the same as flow for openaccess posts.

TODO b/303489420 to cleanup autoPromptType variable.


justinchou-google:
See https://github.com/subscriptions-project/swg-js/pull/3229.

PR #3229, opened 2023-10-04
Extend frequency capping to all dismissible prompts

justinchou-google:
b/303458595, b/303459208 Frequency capping is applied to `:basic` premonetization and `:openaccess` subscription posts. This covers ALL dismissible prompts.


src/runtime/auto-prompt-manager.ts
```diff
@@ -690,12 +699,15 @@ export class AutoPromptManager {
...
+    }
+
     // FrequencyCapConfig may be undefined ONLY for subscription openacess
     // content. If so, display first valid action.
-    if (
-      this.isSubscription_(autoPromptType) ||
-      !this.isValidFrequencyCap_(frequencyCapConfig)
-    ) {
+    const isLockedContent =
+      this.isSubscription_(autoPromptType) && this.pageConfig_.isLocked();
```
oyj9109, line 709: Would `isClosable` be more accurate than subscription && locked?
What's the expected behavior for non-subscription && (isClosable == false)?

src/runtime/auto-prompt-manager.ts
```diff
@@ -704,8 +716,11 @@ export class AutoPromptManager {
       return actions[0];
     }
 
-    const globalFrequencyCapDuration =
-      frequencyCapConfig?.globalFrequencyCap?.frequencyCapDuration;
+    const isSubscriptionOpenaccess =
+      this.isSubscription_(autoPromptType) && !this.pageConfig_.isLocked();
+    const globalFrequencyCapDuration = isSubscriptionOpenaccess
+      ? SUBSCRIPTION_OPENACCESS_GLOBAL_FREQUENCY_CAP_DURATION
+      : frequencyCapConfig?.globalFrequencyCap?.frequencyCapDuration;
```
oyj9109, line 723: Can we set up default value for global frequency in the server? Then I guess subscription & openaccess would run through the rest of logic.

justinchou-google: Will move this to server side.

src/runtime/auto-prompt-manager.ts
justinchou-google: Discussed offline, extending to `isClosable` prompts.

oyj9109 (approved):

PR #3232, opened 2023-10-05
Update Rewarded Ads UX

mhkawano:
Focus on first button for a11y and add hover/focus css for the buttons.


fikus (approved):
Also consider adding e2e/screenshot tests (separately from this PR).

PR #3236, opened 2023-10-06
Fetch entitlements for showContributionOptions demo flow

fikus:
Fetch entitlements so the client config is populated before showing the contributions dialog in the local scenic demo flow.

Same change as
https://github.com/subscriptions-project/scenic-demo/pull/433 for local scenic.

b/279226492


examples/sample-pub/publisher.js
```diff
@@ -393,6 +396,17 @@ function startFlowAuto() {
     });
     return;
   }
+
+  if (flow === 'showContributionOptions') {
+    onEntitlements = (subscriptions) => {
```
justinchou-google, line 401: should we rename this `contributions`?

justinchou-google (approved):

examples/sample-pub/publisher.js
fikus: `subscriptions` is the standard name of the parameter (corresponds to the API in `subscriptions.ts`).

PR #3239, opened 2023-10-06
Add configuration for enabling new paywall background click behavior.

mborof:
The new behavior will only activate if the setting is explicitly set to true.  Eventually we can deactivate it if the setting is set to false.

Still TODO is to add tests


src/api/basic-subscriptions.ts
```diff
@@ -114,6 +114,10 @@ export interface ClientOptions {
   allowScroll?: boolean;
   /** Skip account creation screen if requested. */
   skipAccountCreationScreen?: boolean;
+  /** Prevents links from being clicked through the background.  Clicking also
```
fikus, line 117: Comment formatting is incorrect.

fikus, line 117: This is updating public API. Are integrators expected to control this flag value? I don't think this is covered by the current design.

mborof:
Going will server-side controlled experiment

PR #3243, opened 2023-10-10
Add activity specific support for dialog background click control

mborof:
Prevent clicks through the darkened background of offers, contributions, registration and surveys for SwgZero behind an experiment.  Also, close the dialog on click if the dialog is closable.


src/runtime/contributions-flow-test.js
```diff
@@ -348,6 +348,46 @@ describes.realWin('ContributionsFlow', (env) => {
...
+      .withExactArgs(
+        sandbox.match.any,
+        false,
+        sandbox.match({closeOnBackgroundClick: true})
+      )
+      .once();
+    await contributionsFlow.start();
+  });
+
+  it('opens dialog with closeOnBackgroundClick true when isClosable=false and useUpdatedOfferFlows=true', async () => {
```
fikus, line 369: should be false?

PR #3244, opened 2023-10-10
Make rewarded ads prompt text scale by rem, and buttons bigger

mhkawano:
Also made it so you can scroll the prompt so stuff doesn't get cut off.


fikus (approved):
Is there a bug ID you can link?

mhkawano:
> Is there a bug ID you can link?
b/304359663
b/304360857

PR #3245, opened 2023-10-10
Disable fade in for monetizaton prompts called from rewarded ads

mhkawano:
Disable the fade in animation for the graypane when using rewarded ads so that it does not flicker when switching between them.

b/301934795


src/api/subscriptions.ts
```diff
@@ -512,6 +512,11 @@ export interface OffersRequest {
    * upgrade or downgrade their current subscription.
    */
   oldSku?: string;
+
+  /**
+   * Optional. Sets if the fade in of is animated.
+   */
+  shouldNotAnimateFade?: boolean;
```
justinchou-google, line 519: can we make them all `shouldAnimateFade` for consistency?

mhkawano: Done.

mhkawano: I set them all to `shouldNotAnimateFade` so that if the value is unset, it will default to animating to preserve current behavior.

src/api/subscriptions.ts
```diff
@@ -512,6 +512,11 @@ export interface OffersRequest {
    * upgrade or downgrade their current subscription.
    */
   oldSku?: string;
+
+  /**
+   * Optional. Sets if the fade in of is animated.
```
fikus, line 517: nit: check the wording of this sentence?

src/runtime/auto-prompt-manager.ts
```diff
@@ -766,7 +767,8 @@ export class AutoPromptManager {
               isClosable,
               monetizationFunction: this.getLargeMonetizationPromptFn_(
                 autoPromptType,
-                !!isClosable
+                !!isClosable,
+                /* shouldNotAnimateFade */ true
```
fikus, line 771: Read from `options`?

src/api/subscriptions.ts
fikus: I agree, the logic is hard to follow with the double negations.

Also consider that this is updating the public API. Is this necessary? Can we pass this option in another way?

If we do need to update the API, it's important that we keep it readable. IMO it would be better to name this positively with a comment that it defaults to true, and implement accordingly.

src/runtime/auto-prompt-manager.ts
mhkawano: This is for rewarded ads, so it should always suppress the animation here.

src/api/subscriptions.ts
fikus: If the current implementation requires updating the API, can you add a TODO to revisit? This UI detail doesn't really belong in OffersRequest.

mhkawano: Done. Updated the comment and filed a bug to clean this up later.

src/api/subscriptions.ts
```diff
@@ -512,6 +512,13 @@ export interface OffersRequest {
    * upgrade or downgrade their current subscription.
    */
   oldSku?: string;
+
+  /**
+   * Optional. Disables the fade in animation if set to false. Defaults to true
+   * if unset.
+   * TODO: b/304803271 - remove this field form the api.
```
fikus, line 519: nit: typo "form"

fikus (approved):

PR #3252, opened 2023-10-13
Send analytics context to audience actions iframe

mborof:
Context: b/305097228

This sends components necessary for the new back link logging to occur.  This was also necessary for the redirect to work (based on the way the code is currently written).

### DIFF
--- a/src/ui/activity-iframe-view.ts
+++ b/src/ui/activity-iframe-view.ts
@@ -69,7 +69,8 @@ export class ActivityIframeView extends View {
     const port = await this.activityPorts_.openIframe(
       this.iframe_,
       this.src_,
-      this.args_
+      this.args_,
+      true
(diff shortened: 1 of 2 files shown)


src/ui/activity-iframe-view.ts
```diff
@@ -69,7 +69,8 @@ export class ActivityIframeView extends View {
     const port = await this.activityPorts_.openIframe(
       this.iframe_,
       this.src_,
-      this.args_
+      this.args_,
+      true
```
fikus, line 73: Since it is not clear from context, can you document the meaning of this argument with an inline comment?

fikus (commented):
Can you explain the change in some more detail? For example, it isn't clear how this change to activity iframe specifically affects audience actions.

PR #3254, opened 2023-10-13
Start buy flow if loading for rewarded ads fails

mhkawano:
In cases where we cannot init gpt.js we will mark rewarded ads as dismissed and show a buy flow.

b/305721069

### DIFF
--- a/src/runtime/audience-action-local-flow.ts
+++ b/src/runtime/audience-action-local-flow.ts
@@ -22,10 +22,11 @@ import {
 } from './audience-action-flow';
 import {AutoPromptType} from '../api/basic-subscriptions';
 import {
-  CLOSE_BUTTON_HTML,
   CONTRIBUTION_ICON,
   ERROR_HTML,
   LOADING_HTML,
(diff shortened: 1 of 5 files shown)


src/runtime/audience-action-local-flow.ts
```diff
@@ -335,18 +335,24 @@ export class AudienceActionLocalFlow implements AudienceActionFlow {
...
+        googletag.destroySlots([this.rewardedSlot_!]);
+      }
       this.eventManager_.logSwgEvent(
         AnalyticsEvent.EVENT_REWARDED_AD_GPT_ERROR
       );
-      resolve(true);
-      // TODO: mhkawano - Launch payflow if monetized, cancel if not.
+      if (this.params_.onCancel) {
+        this.params_.onCancel();
+      }
```
fikus, line 349: nit: replace with `this.params_.onCancel?.();`

fikus (approved):
Can you also add a bug ID for future reference?

mhkawano:
Applied the changed and updated the description with a bug.

PR #3256, opened 2023-10-16
Remove references to flag `populate-client-config-classic`

fikus:
b/280305573

### DIFF
--- a/src/runtime/basic-runtime-test.js
+++ b/src/runtime/basic-runtime-test.js
@@ -784,7 +784,7 @@ describes.realWin('BasicConfiguredRuntime', (env) => {
             engineId: '123',
           },
         })
-        .once();
+        .atLeast(1);
       const uiPredicates = new UiPredicates(/* canDisplayAutoPrompt */ true);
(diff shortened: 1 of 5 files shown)


justinchou-google (approved):
Seems like this a bit late for this question as this experiment is already launched, but do we intend the expected behavior to fetch the article response more than once?

fikus:
> Seems like this a bit late for this question as this experiment is already launched, but do we intend the expected behavior to fetch the article response more than once?

Thanks for the question Justin. The multiple calls to `getArticle` are expected, they do not actually result in multiple fetches as subsequent calls will re-use the same response promise.

An earlier attempt at this clean-up PR removed the apparent redundant call to getEntitlements in basic-runtime. This caused another issue because getEntitlements has the side effect of initializing analytics. There are details in the linked bug.

For the multiple `fetch` calls in `runtime-test`, these were caused by separate fetches to /entitlements and /clientconfig - the test code had not been updated to use /article instead. I've updated the tests to correct this. We also have b/279927284 to clean up this unneeded option and always use /article.

PR #3275, opened 2023-11-01
Add focus trap for rewarded ads

mhkawano:
Prevents readers from moving focus out of the prompt.

Bug with demo: b/308786966


oyj9109 (approved):
Attaching a screencast would be helpful for future refenrece if you have a demo up. Thanks!

mhkawano:
Bug with demo: b/308786966

src/runtime/audience-action-local-flow.ts
```diff
@@ -573,12 +576,47 @@ export class AudienceActionLocalFlow implements AudienceActionFlow {
...
+
+  private updateFocusable_() {
+    this.focusableElements_ = this.prompt_.querySelectorAll(
+      'a[href], button, input, textarea, select, details, [tabindex]:not([tabindex="-1"])'
+    );
+  }
+
+  private focusTrap_(e: KeyboardEvent) {
+    if (
+      e.key !== 'Tab' ||
```
fikus, line 598: I think we should detect focus events rather than assuming that Tab keydown is the only way focus would be changing. Tab key would not be applicable on mobile for example.

src/runtime/audience-action-local-flow.ts
```diff
@@ -573,12 +576,47 @@ export class AudienceActionLocalFlow implements AudienceActionFlow {
...
+      this.focusableElements_ == null ||
```
fikus, line 599: nit: `===`

fikus (changes requested):
- Can the focus trap functionality be extracted to a separate class, so it can be reused?
- Please include tests.

src/runtime/audience-action-local-flow.ts

src/runtime/audience-action-local-flow.ts
mhkawano: Removed due to rework.

mhkawano:
Test is giving me some trouble, will add next.

mhkawano:
Tests are added.

Implemented the trap by adding an event listener to every element outside of our dialog. Kept the implementation simple by having it focus on the last item by default. I'm thinking that I eventually want to replace this with <dialog>, but it doesn't seem to work well with shadow doms, and I don't want to get in the way of the BYOP work, so doing this for now.

src/runtime/audience-action-local-flow.ts
```diff
@@ -573,12 +574,42 @@ export class AudienceActionLocalFlow implements AudienceActionFlow {
...
     setStyle(this.doc_.body, 'overflow', '');
+    this.removeFocusTrap_();
+  }
+
+  private focusRewardedAds_() {
+    (this.prompt_.querySelector('.rewarded-ad-prompt')! as HTMLElement).focus();
+  }
+
+  private addFocusTrap_() {
+    this.focusLastBound_ = this.focusLast_.bind(this);
```
fikus, line 585: Can focusLastBound be initialized where it is declared?

src/runtime/audience-action-local-flow.ts
```diff
@@ -573,12 +574,42 @@ export class AudienceActionLocalFlow implements AudienceActionFlow {
...
+    this.docFocusableElement = this.doc_.body.querySelectorAll(
```
fikus, line 586: nit: plural name since this is a collection?

src/runtime/audience-action-local-flow.ts
```diff
@@ -573,12 +574,42 @@ export class AudienceActionLocalFlow implements AudienceActionFlow {
...
+      'a[href], button, input, textarea, select, details, [tabindex]:not([tabindex="-1"])'
+    );
+    this.docFocusableElement.forEach((element) => {
+      element.addEventListener('focus', this.focusLastBound_!);
+    });
+  }
+  private removeFocusTrap_() {
```
fikus, line 593: nit: blank space before function

src/runtime/audience-action-local-flow.ts
```diff
@@ -573,12 +574,42 @@ export class AudienceActionLocalFlow implements AudienceActionFlow {
...
+
+  private focusRewardedAds_() {
+    (this.prompt_.querySelector('.rewarded-ad-prompt')! as HTMLElement).focus();
+  }
+
+  private addFocusTrap_() {
+    this.focusLastBound_ = this.focusLast_.bind(this);
+    this.docFocusableElement = this.doc_.body.querySelectorAll(
+      'a[href], button, input, textarea, select, details, [tabindex]:not([tabindex="-1"])'
+    );
```
fikus, line 588: Extract to separate function since this query is repeated?

fikus (commented):
- Did you consider my earlier comment about moving this to a separate class so the logic can be reused?
- In terms of correctness, it seems safer not to interact with the hosting page document since this can have arbitrary behavior outside of our control. For example, new focusable elements could be created dynamically while the dialog is visible, without the trapping focus event listener.

mhkawano:
> Did you consider my earlier comment about moving this to a separate class so the logic can be reused?

Yes, I wrote above that I don't want to make this a permanent solution and would prefer to use the dialog element(seems that using the angle brackets made it disappear), but that would require a more complicated rework. If we don't want to use the dialog element I can rework this into being its own class that other prompts could use.

> In terms of correctness, it seems safer not to interact with the hosting page document since this can have arbitrary behavior outside of our control. For example, new focusable elements could be created dynamically while the dialog is visible, without the trapping focus event listener.

Ack, added the sentinels.

fikus (approved):

PR #3303, opened 2023-11-10
Rewarded Ads auto dismiss

mhkawano:
We want to automatically dismiss the thanks prompt after 3 seconds. Also generally cleaned up timeouts and make a small css adjustment.

b/310192870


src/runtime/auto-prompt-manager.ts
```diff
@@ -1117,8 +1117,14 @@ export class AutoPromptManager {
...
     } else if (actionType === TYPE_REWARDED_ADS) {
-      // Because we have fetched the article endpoint googletag.cmd should already exist.
-      const googletagExists = !!this.deps_.win().googletag?.cmd;
+      /*
+       * Because we have fetched the article endpoint gpt.js should
+       * already be ready, but this is a race condition. Failing the race
+       * condition results in not showing the prompt.
+       */
+      const googletagExists =
+        !!this.deps_.win().googletag?.apiReady &&
```
oyj9109, line 1126: I guess this change will go away once you sync to head?

oyj9109 (approved):

src/runtime/auto-prompt-manager.ts
mhkawano: Yes, will reconfirm everything looks correct after syncing.

PR #3310, opened 2023-11-15
Add rewarded ads experiment arms and better error handling

mhkawano:
This PR contains two changes, but they were bundled together to keep testing simpler.

# 1. Adding new experiments
This contains an experiment to disable the close button on the prompt, forcing the user to click one of the buttons first, at which point the prompt becomes closable. It also contains an experiment to swap the order and styling of the buttons.

# 2. More robust error handling
Based on our experiments, it seems that in the real world we cannot depend on gpt.js being loaded ahead of the action filtering. This adds the check to the rewarded ads action, has it poll for 5 seconds, then launches a buyflow if it can't detect gpt.js.

b/310653573
b/309441494
b/309440326


src/runtime/audience-action-local-flow.ts
```diff
@@ -647,15 +701,22 @@ export class AudienceActionLocalFlow implements AudienceActionFlow {
   }
 
   private getCloseButtonHtml_(html: string) {
```
oyj9109, line 703: nit: the function name seems misleading that it can return an empty string. I think ideally the caller should check the condition, but I guess we can simply correct it by renaming the function like `getCloseButtonOrEmptyHtml_`?

src/runtime/experiment-flags.ts
```diff
@@ -54,4 +54,14 @@ export enum ArticleExperimentFlags {
    * closable popups.
    */
   BACKGROUND_CLICK_BEHAVIOR_EXPERIMENT = 'background_click_behavior_experiment',
+
+  /**
+   * When enabled the rewarded ad wall isn't closable until the ad is viewed.
+   */
+  REWARDED_ADS_CLOSABLE_ENABLED = 'rewarded_ads_closable_enabled',
```
oyj9109, line 61: The flag name seems quite misleading. It actually disables "closability" when "closable_enabled" is true. Can we rename this something like "always blcoking"?

src/runtime/audience-action-local-flow.ts
```diff
@@ -647,15 +701,22 @@ export class AudienceActionLocalFlow implements AudienceActionFlow {
...
+    if (
+      !this.params_.isClosable ||
```
oyj9109, line 705: Would it be possible to override this value at the top (probably in the constructor?) so that it's the single source of the truth throughout the action flow? There are other code referecning this fields and the outcome can be unexpected if `this.params_.isClosable` = true but close button is actually not rendered.

src/runtime/audience-action-local-flow.ts
```diff
@@ -304,35 +326,61 @@ export class AudienceActionLocalFlow implements AudienceActionFlow {
...
+  }
 
-    const config = await this.getConfig_();
+  private async checkGoogletagAvailable_(): Promise<boolean> {
+    const googletagReady = await this.googletagReady_();
+    return googletagReady && !!this.deps_.win().googletag?.getVersion();
+  }
 
+  private async initRewardedAdWall_() {
+    this.eventManager_.logSwgEvent(AnalyticsEvent.IMPRESSION_REWARDED_AD);
```
oyj9109, line 345: Should this be logged on successful impression?

src/runtime/audience-action-local-flow.ts
```diff
@@ -304,35 +326,61 @@ export class AudienceActionLocalFlow implements AudienceActionFlow {
...
   }
 
-  private async initRewardedAdWall_() {
-    this.eventManager_.logSwgEvent(AnalyticsEvent.IMPRESSION_REWARDED_AD);
+  private async googletagReady_(): Promise<boolean> {
+    for (let i = 0; i < this.detectGptRetries_; i++) {
+      if (this.deps_.win().googletag?.apiReady !== undefined) {
+        return this.deps_.win().googletag.apiReady;
+      }
+      await new Promise((r) => setTimeout(r, this.detectGptRetriesMs_));
```
oyj9109, line 334: User perceivable delay up to 5s seems pretty long. Have you considered delaying initiation of the flow by respecting the client config, `clientDisplayTrigger`? You just need to register the prompt type in the list: https://github.com/subscriptions-project/swg-js/blob/7618e056fab16fb0b465d090a157ca2fe2fe8f97/src/runtime/auto-prompt-manager.ts#L888

src/runtime/audience-action-local-flow.ts
```diff
@@ -442,22 +484,31 @@ export class AudienceActionLocalFlow implements AudienceActionFlow {
       .replace('$REWARDED_AD_CLOSE_BUTTON_HTML$', closeHtml)
       .replace('$ICON$', icon)
       .replace('$MESSAGE$', message)
-      .replace('$VIEW_AN_AD$', viewad)
+      .replace('$VIEW_AN_AD$', prioritySwaped ? support : viewad)
```
oyj9109, line 487: It looks like it's basically swapping the two button elements. Would it be possible to make 2 placeholders for where the buttons are displayed and dynamically placing the entire button elements w/o having to cross-bind callback to each other?
If it requires too much work for refactoring, leaving a TODO can be okay for a temporary experiment.

src/runtime/audience-action-local-flow.ts
mhkawano: Done

src/runtime/audience-action-local-flow.ts
mhkawano: I don't think we want to do that. We only want the initial prompt to be unclosable, but for it to otherwise act like a closable prompt. For example, if an ad view is canceled causing the `rewardedSlotClosed` callback to be issued, we want to allow the page to be unlocked. If we override `isClosable` in the constructor it would keep it locked.

src/runtime/audience-action-local-flow.ts
mhkawano: This should be logged for everything. In retrospect I realize that's different from how other prompts do it, but I prefer this way because it allows us to easily see everything as a portion of overall impressions.

src/runtime/audience-action-local-flow.ts
mhkawano: Added a todo. But because this is a temporary experiment, I think it's more likely we end up removing it before refactoring.

src/runtime/audience-action-local-flow.ts
mhkawano: Rewarded ads should already be added to isActionPromptWithDelay_. There are a number of cases I considered regarding delay:

Contribution:
Normal client - gpt.js should already be loaded because of the `clientDisplayTrigger` delay
Has fake gpt.js - should instantly be detected(no 5 second delay)
Has gpt.js completely blocked - will have delay, but it should be a small portion of readers

Subscription:
There is potentially delay for all clients, but I also think we want to lock the page immediately.

src/runtime/experiment-flags.ts
mhkawano: Done. Updated backend CL too.

src/runtime/audience-action-local-flow.ts
oyj9109: Ah, good to know rewarded ads is already added. Just out of curiosity, I believe the delay is 10s for the alpha publisher. Wasn't it sufficient to wait for gpt.js to fully load? Do you think waiting for additional 5s would resolve the issue a lot?

src/runtime/audience-action-local-flow.ts
oyj9109: I'm afraid that could lead to misunderstanding in analytics. Impression by definition implies successful impressions and making an exception would create confusions for other people. I would recommend creating a separate event type to count the overall number of flow initiation, rather than using the impression event for this purpose. Also, refer to [go/swg-logged-event-types](https://goto.google.com/swg-logged-event-types)

src/runtime/audience-action-local-flow.ts
oyj9109: Ah, got it. Thanks for clarification. It's really hard to understand from code that closability between initial and other prompts can differ. Can we make it easy to understand by declaring variables with clear name?

src/runtime/audience-action-local-flow.ts
mhkawano: For prompts that already have a delay, probably not. I can skip the re-try if we detect that this is a contribution prompt.

src/runtime/audience-action-local-flow.ts
mhkawano: Ack. Moved impressions to when the prompt is ready to be shown. Don't think we need the new event yet as we can infer it from the total of the impressions and errors, but added a todo because it will be nice to have.

src/runtime/audience-action-local-flow.ts
mhkawano: Moved the condition into a named variable.

src/runtime/audience-action-local-flow.ts
oyj9109: Would it be more accurate to check if it's closable? (covers all premon/contribution/subscription openaccess)

oyj9109 (approved):
LGTM. Thanks for addressing all the comments.

PR #3321, opened 2023-11-22
Update translation strings

mhkawano:
Also the Swedish country code seems to be incorrect for some of them.


src/i18n/swg-strings.ts
```diff
@@ -512,45 +512,44 @@ export const SWG_I18N_STRINGS = {
...
-    'zh-tw': 'View an ad',
+    'es': 'Ver un anuncio',
+    'es-419': 'Ver un anuncio',
+    'fr': 'Visionner une annonce',
+    'fr-ca': 'Afficher une annonce',
+    'hi': 'कोई विज्ञापन देखें',
+    'id': 'Lihat iklan',
+    'it': 'Visualizza un annuncio',
+    'ja': '広告を表示',
+    'ko': 'V광고 보기',
```
oyj9109, line 539: Looks like 'V' is a typo here? (As a korean I can tell this is wrong to be there)

oyj9109 (approved):

oyj9109:
Looks like auto-merged. Can you follow up on my comment to correct the typo?

mhkawano:
Ack. I will include it in my experiment PR.

PR #3325, opened 2023-11-27
record completion immediately after form submission

kristenwang:
For Convertkit (another ESP tool), submitting the form will refresh the current page the reader is on. 
Therefore, we need to record the completion event immediately after form submission, otherwise, it will not be recorded after the page has been redirected.


src/runtime/audience-action-local-flow.ts
```diff
@@ -302,13 +302,14 @@ export class AudienceActionLocalFlow implements AudienceActionFlow {
     // Hide prompt before closing the prompt.
     setImportantStyles(this.wrapper_, {'opacity': '0'});
     // Wait for form submit request to send before closing the prompt.
-    await this.delay_(1000);
     this.eventManager_.logSwgEvent(
       AnalyticsEvent.ACTION_BYOP_NEWSLETTER_OPT_IN_SUBMIT
     );
+    await this.complete_();
+
+    await this.delay_(1000);
```
justinchou-google, line 310: is this delay still needed?

kristenwang: It is possible. However, I don't think we can intercept a third party post call & it's response, especially we are expanding to multiple thrid parties, I don't think there is an unanimous signal from the different ESP tools.

We could potentially wait for a longer period of time if we think 1 second if not enough?

PR #3374, opened 2024-01-03
Update z-index for local audience action to match other prompts

mhkawano:
This decreases the z-index by 1 so that it will appear below other items with the max z-index, like the anchor banner in the screenshot below. This changes the behavior to match our other prompts, and was decided by PM.

Bug: b/318516613

Screenshot:
https://github.com/subscriptions-project/swg-js/assets/106177651/bb25933f-b89a-4a46-9a32-ed6b4597805e


oyj9109:
Can you provide more specific information that it puts our prompt behind the anchored ads in the PR description?

oyj9109 (approved):

mhkawano:
Updated description. Re-running jobs.

PR #3378, opened 2024-01-04
Update local audience action error view

mhkawano:
Make the error view prompt look slightly better for local audience actions.  Does not currently render a close button  for rewarded ads, as it will only render for locked pages(see `bailoutPrompt_`). We may need to make the prompt closable for BYOP @kristenwang .

Before:
https://screenshot.googleplex.com/9bLjmyJcs56xxhx

After:
https://screenshot.googleplex.com/3uW8zYknQYFys7f

b/318537252


justinchou-google:
Does this mean that if the rewarded ad fails, the reader is entirely blocked from reading the publication given there is no option to dismiss?

mhkawano:
> Does this mean that if the rewarded ad fails, the reader is entirely blocked from reading the publication given there is no option to dismiss?

Yes, but this shouldn't be an issue. For rewarded ads the error view is only rendered if:
1. There was an error, otherwise render normally
2. There is no monetization function, otherwise use the monetization function
3. The content is locked, otherwise just dismiss

See the `bailoutPrompt_` method. Basically this case will only happen for locked premonetization pages, which should be pretty rare. There may be some issues with BYOP though, but those are existing issues.

justinchou-google (approved):
Thanks for the clarification!

PR #3422, opened 2024-02-14
Enable localhost reverse proxy serving

everyplace:
This PR moves some changes from the stand-alone `scenic-demo` repo to the embedded version within this repo. These changes allow for running the local app behind a reverse proxy, which can be of benefit for some development environments. The changes migrate only some of the changes from [PR #217](https://github.com/subscriptions-project/scenic-demo/pull/217/files), as that PR handled two things:

1. Enable serving behind a reverse proxy
2. Using env vars to update some varables

This smaller PR only does the first of those two tasks.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -22,6 +22,7 @@ if (!isCiBuild()) {
   execOrDie('npm i');
 }
 
+const dotenv = require('dotenv').config();
 const $$ = require('gulp-load-plugins')();
 const gulp = $$.help(require('gulp'));
 const {assets} = require('./build-system/tasks/assets');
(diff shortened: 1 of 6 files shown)


gulpfile.js
```diff
@@ -23,6 +23,7 @@ if (!isCiBuild()) {
 }
 
 const $$ = require('gulp-load-plugins')();
+const dotenv = require('dotenv').config(); // eslint-disable-line
```
everyplace, line 26: Because dotenv runs immediately to read `.env` files if available, it does not need to be used after creation. As this violates the `eslint/no-unused-vars` rule, I've disabled it for this line.

everyplace: @ChrisAntaki I tested this, and it still worked fine, so I've removed it in https://github.com/subscriptions-project/swg-js/pull/3422/commits/78a6d31648539ac5a8deccb30e141a8bd6e31323

PR #3436, opened 2024-02-21
Escape publisher provided strings

mhkawano:
This escapes the title, and message, both provided by the publisher. We use the safevalues library, which is an open source Google library recommended by the ISE's secure coding site.

b/326280984


src/runtime/audience-action-local-flow-test.js
```diff
@@ -413,6 +413,41 @@ describes.realWin('AudienceActionLocalFlow', (env) => {
         expect(signinButton).to.be.null;
       });
 
+      it('escapes bad input', async () => {
+        const BAD_CONFIG = `
+          {
+            "publication": {
+              "name": "<script>PUBLICATOIN_NAME</script>"
```
justinchou-google, line 420: nit: PUBLICATION_NAME (my understanding is the html is what makes it a bad input and not the typo, correct me if i'm wrong)

package.json
```diff
@@ -27,6 +27,7 @@
     "build-storybook": "build-storybook"
   },
   "dependencies": {
+    "safevalues": "0.5.2",
```
justinchou-google, line 30: not sure what our policy is on using external libraries... @oyj9109 @ChrisAntaki do you know?

mhkawano:
Tested Firefox and Safari for polyfill. Merging.

PR #3442, opened 2024-02-23
Evaluate PromptFrequencyCap before GlobalFrequencyCap

justinchou-google:
b/325512849

Reorders test cases such that globalfrequencycap is evaluated first, as well.


src/runtime/auto-prompt-manager.ts
```diff
@@ -746,9 +733,28 @@ export class AutoPromptManager {
...
       }
-      return action;
+      potentialAction = action;
+      break;
+    }
+
+    if (!potentialAction) {
+      return;
+    }
+
```
oyj9109, line 743: Can you add some comments to make it clear that the order between prompt/global frequency cap evaluation matters and this is the intended order?

oyj9109 (approved):

src/runtime/auto-prompt-manager.ts
justinchou-google: Done.

PR #3446, opened 2024-02-26
Computes frequency cap based on dismissal timestamps instead of impressions.

justinchou-google:
b/325512188.

Guarded by new flag. If enabled, this flow will take precedence over the impression based flow.

### DIFF
--- a/src/runtime/auto-prompt-manager.ts
+++ b/src/runtime/auto-prompt-manager.ts
@@ -1041,7 +1041,6 @@ export class AutoPromptManager {
       ) ||
       !this.isClosable_
     ) {
-      console.log('here');
       return;
     }
 
--- a/src/runtime/auto-prompt-manager.ts
+++ b/src/runtime/auto-prompt-manager.ts
@@ -149,6 +149,7 @@ const ACTION_TO_DISMISSAL_STORAGE_KEY_MAP = new Map([
   [TYPE_REGISTRATION_WALL, DismissalStorageKeys.REGISTRATION_WALL],
   [TYPE_REWARDED_SURVEY, DismissalStorageKeys.REWARDED_SURVEY],
   [TYPE_REWARDED_ADS, DismissalStorageKeys.REWARDED_AD],
+  [TYPE_SUBSCRIPTION, DismissalStorageKeys.SUBSCRIPTION],
 ]);
 
 const ACTION_TO_IMPRESSION_STORAGE_KEY_MAP = new Map([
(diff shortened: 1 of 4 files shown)


src/runtime/auto-prompt-manager.ts
```diff
@@ -106,6 +143,14 @@ const INTERVENTION_TO_STORAGE_KEY_MAP = new Map([
   [AnalyticsEvent.IMPRESSION_OFFERS, ImpressionStorageKeys.SUBSCRIPTION],
 ]);
 
+const ACTION_TO_DISMISSAL_STORAGE_KEY_MAP = new Map([
+  [TYPE_CONTRIBUTION, DismissalStorageKeys.CONTRIBUTION],
```
oyj9109, line 147: What about subscription?

src/runtime/auto-prompt-manager.ts
```diff
@@ -930,7 +986,10 @@ export class AutoPromptManager {
     }
 
     // ** Frequency Capping Impressions **
-    if (this.frequencyCappingLocalStorageEnabled_) {
+    if (
+      this.frequencyCappingByDismissalsEnabled_ ||
```
oyj9109, line 990: What's the ramp-up strategy for the new flag? 
- Are we going to turn down ongoing experiment and set up a new experiment with dismissal that goes up to 50%?
- Or are we going to keep the impression experiment and run a new experiment with dismissals that sits on top of it and goes up to 100%?

If the latter, wouldn't it be accurate to only refer to `frequencyCappingLocalStorageEnabled_` to keep it to 50%?

src/runtime/auto-prompt-manager.ts
```diff
@@ -930,7 +986,10 @@ export class AutoPromptManager {
     }
 
     // ** Frequency Capping Impressions **
```
oyj9109, line 988: Should we update the comment?

src/runtime/auto-prompt-manager.ts
justinchou-google: Added.

src/runtime/auto-prompt-manager.ts
justinchou-google: Good point, keeping this check to only `frequencyCappingLocalStorageEnabled_` so that the new flag will be able to ramp to 100%.

src/runtime/auto-prompt-manager.ts
justinchou-google: Updated!

oyj9109 (approved):

PR #3453, opened 2024-03-12
Prompt frequency by completions and dismissals, Global frequency by impressions.

justinchou-google:
b/325512188
https://docs.google.com/document/d/1H99mlR47aZZv_ZsCRkI6Iv_WUcpwaoeGgCExWSkdYCI/edit?resourcekey=0-9djhO0eRtzh5bMMrZ1dtBg&tab=t.0#heading=h.8ztm2we0184l

### DIFF
--- a/src/runtime/auto-prompt-manager.ts
+++ b/src/runtime/auto-prompt-manager.ts
@@ -204,7 +204,7 @@ export class AutoPromptManager {
 
     this.clientConfigManager_ = deps_.clientConfigManager();
     assert(
-      this.clientConfigManger_,
+      this.clientConfigManager_,
       'AutoPromptManager requires an instance of ClientConfigManager.'
     );
 
--- a/src/runtime/auto-prompt-manager.ts
+++ b/src/runtime/auto-prompt-manager.ts
@@ -205,6 +205,7 @@ export class AutoPromptManager {
     this.clientConfigManager_ = deps_.clientConfigManager();
     assert(
       this.clientConfigManager_,
+
       'AutoPromptManager requires an instance of ClientConfigManager.'
     );
 
--- a/src/runtime/auto-prompt-manager.ts
+++ b/src/runtime/auto-prompt-manager.ts
@@ -38,18 +38,14 @@ import {ClientEvent} from '../api/client-event-manager-api';
 import {ClientEventManager} from './client-event-manager';
 import {ConfiguredRuntime} from './runtime';
 import {Deps} from './deps';
-import {
-  DismissalStorageKeys,
-  ImpressionStorageKeys,
-  StorageKeys,
-} from '../utils/constants';
 import {Doc} from '../model/doc';
 import {Entitlements} from '../api/entitlements';
 import {GoogleAnalyticsEventListener} from './google-analytics-event-listener';
+import {ImpressionStorageKeys, StorageKeys} from '../utils/constants';
 import {MiniPromptApi} from './mini-prompt-api';
 import {OffersRequest} from '../api/subscriptions';
 import {PageConfig} from '../model/page-config';
-import {Storage} from './storage';
+import {Storage, pruneTimestamps} from './storage';
 import {assert} from '../utils/log';
 import {isExperimentOn} from './experiments';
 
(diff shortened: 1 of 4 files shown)


src/runtime/auto-prompt-manager.ts
```diff
@@ -1134,51 +1172,131 @@ export class AutoPromptManager {
...
-  private async getActionImpressions_(actionType: string): Promise<number[]> {
-    if (!ACTION_TO_IMPRESSION_STORAGE_KEY_MAP.has(actionType)) {
-      this.eventManager_.logSwgEvent(
-        AnalyticsEvent.EVENT_ACTION_IMPRESSIONS_STORAGE_KEY_NOT_FOUND_ERROR
+  async getTimestamps(): Promise<LocalStorageTimestamps[]> {
+    const stringified = await this.storage_.get(
+      StorageKeys.TIMESTAMPS,
+      /* useLocalStorage */ true
+    );
+    const array = stringified ? JSON.parse(stringified) : [];
```
oyj9109, line 1195: Can you leave a TODO to handle parsing error so that it doesn't break the entire flow?

src/runtime/auto-prompt-manager.ts
```diff
@@ -167,6 +161,13 @@ export interface ShowAutoPromptParams {
   isClosable?: boolean;
 }
 
+interface LocalStorageTimestamps {
```
oyj9109, line 164: nit: what do you think about renaming this to `ActionTimestamps` to make it clear that it's per action type?

src/runtime/auto-prompt-manager.ts
```diff
@@ -1134,51 +1172,131 @@ export class AutoPromptManager {
...
+  async getTimestamps(): Promise<LocalStorageTimestamps[]> {
+    const stringified = await this.storage_.get(
+      StorageKeys.TIMESTAMPS,
+      /* useLocalStorage */ true
+    );
+    const array = stringified ? JSON.parse(stringified) : [];
+
+    return array
+      .map((t: LocalStorageTimestamps) => {
+        const impressions = t.impressions
```
oyj9109, line 1199: nit: what about
const impressions = pruneTimestamps(t.impressions || [], TWO_WEEKS_IN_MILLIS)

src/runtime/auto-prompt-manager.ts
```diff
@@ -1134,51 +1172,131 @@ export class AutoPromptManager {
...
+          : [];
+        const dismissals = t.dismissals
+          ? pruneTimestamps(t.dismissals, TWO_WEEKS_IN_MILLIS)
+          : [];
+        const completions = t.completions
+          ? pruneTimestamps(t.completions, TWO_WEEKS_IN_MILLIS)
+          : [];
+        return {
+          action: t.action,
+          ...(impressions?.length && {impressions}),
```
oyj9109, line 1210: Can we just set empty array instead of leaving it undefined? I think making timestamp array fields in `LocalStorageTimestamps` interface required would be more strongly typed and easier to process since we don't need to check the presence. WDYT?

With the suggested change above, it can be

```
return {
  action: t.action,
  impressions: pruneTimestamps(t.impressions || [], TWO_WEEKS_IN_MILLIS),
  ...
}
```

src/runtime/auto-prompt-manager.ts
```diff
@@ -1239,6 +1357,29 @@ export class AutoPromptManager {
...
+        GoogleAnalyticsEventListener.isGaEligible(this.deps_) ||
+        GoogleAnalyticsEventListener.isGtagEligible(this.deps_) ||
+        GoogleAnalyticsEventListener.isGtmEligible(this.deps_);
+      if (!isAnalyticsEligible) {
+        return false;
+      }
+      const index = timestamps.findIndex(
+        (t) => t.action === TYPE_REWARDED_SURVEY
+      );
+      return index === -1 ? true : !timestamps[index].completions?.length;
```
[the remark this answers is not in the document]
justinchou-google: Agreed.

src/runtime/auto-prompt-manager.ts
justinchou-google: The tradeoff here is optimizing for timestamps lookups / fetching vs. minimizing local storage bloating. 

I think changing `LocalStorageTimestamps` to a map by action type makes sense, with that in mind we may have the following local storage timestamps:
`
{
    'TYPE_CONTRIBUTION': {impressions: [], dismissals: [], completions: []},
    'TYPE_REWARDED_SURVEY: {impressions: [], dismissals: [], completions: []},
    'TYPE_REGISTRATION_WALL: {impressions: [], dismissals: [], completions: []},
}
`

One possible benefit that I can think of is that this type of local storage can be used to infer that a reader at least had a prompt impression > 2 weeks ago. Otherwise, things that are written to local storage will always persist (other than pruned timestamps). WDYT?

src/runtime/auto-prompt-manager.ts
justinchou-google: Didn't commit this change, added to followup PR.

src/runtime/auto-prompt-manager.ts
justinchou-google: Created https://github.com/subscriptions-project/swg-js/pull/3456.

PR #3457, opened 2024-03-13
Default open on rewarded ads when an ad cannot be shown

mhkawano:
Defaulting to open the page when it's locked, but we cannot display an ad. Will explore alternatives such as making the behavior configurable post-GA.

b/328639440


src/runtime/audience-action-local-flow-test.js
```diff
@@ -138,7 +133,6 @@ describes.realWin('AudienceActionLocalFlow', (env) => {
         '.audience-action-local-wrapper'
       );
       expect(wrapper).to.be.null;
-      expect(params.monetizationFunction).to.be.calledOnce.calledWithExactly();
```
justinchou-google, line 141: Maybe change this to explicitly be `to.not.be.called`? (and add to other failure tests)

mhkawano: Done.

PR #3463, opened 2024-03-20
Adjust darkened background click behavior

mborof:
b/287687727

This adjusts the background CSS so it doesn't suppress click events and modifies the way the iframe is closed so it handles logging.


justinchou-google:
Can you clarify what exactly the desired new "click friendly" behavior is? Is there a bug?

mborof:
I've updated the description

src/components/graypane.ts
```diff
@@ -29,10 +29,11 @@ export class Graypane {
       .getWin()
       .document.createElement('swg-popup-background');
 
+    this.fadeBackground_.setAttribute('role', 'button');
```
justinchou-google, line 32: can we add a comment that actually setting the role button attribute is to block clicks through the pane?

justinchou-google (approved):

PR #3479, opened 2024-03-28
Make background unclickable for client side interventions

mhkawano:
Tested locally

b/328637765


justinchou-google:
Matt just worked on something to make the background clickable so as to suppress clickthroughs to the page (not sure what the scope of that work was). https://github.com/subscriptions-project/swg-js/pull/3463

Can you check with him how these two features are supposed to coexist? Perhaps this is a conversation for Seth and Sagiv.

mhkawano:
These are client side prompt changes to keep the behavior consistent with @mborof 's work.

mborof:
It looks correct to me.  But to validate, ensure it does these two things when the background is clickable/closable:
  1) You can reopen the popup (if the UI allows it).  This was a problem with the original method of closing (re-opening failed)
  2) It logs that the user clicked the close button.

justinchou-google (approved):

mhkawano:
Thanks for the review. It does not log closes, which I have updated it to do. We don't provide a way to reopen the prompt, so I don't think that's a concern.

PR #3494, opened 2024-04-23
Does not record impression timestamp for prompts triggered by CTA button

justinchou-google:
b/333536312
adds attribute promptIsFromCtaButton when cta button to show prompt is clicked, used to not record impression timestamp


oyj9109 (approved):
Thanks for following up on this!

PR #3500, opened 2024-05-01
Add enterprise newsletter API

mhkawano:
Expose funnel prompts to enterprise publishers through the API.

This change is only meant to support adding newsletters, but includes small changes to eventually expand it to support all prompt types.

Screencast in the first bug.

b/270686569
b/336516781
b/270686569


src/runtime/audience-action-flow.ts
```diff
@@ -333,7 +354,10 @@ export class AudienceActionIframeFlow implements AudienceActionFlow {
     const {onResult} = this.params_;
     if (onResult) {
       try {
-        return await onResult(request);
+        return await onResult({
+          configurationId: this.params_.configurationId,
+          data: request,
```
oyj9109, line 359: What does it mean to return the request?

src/runtime/audience-action-flow.ts
```diff
@@ -179,12 +181,25 @@ export class AudienceActionIframeFlow implements AudienceActionFlow {
...
     response: CompleteAudienceActionResponse
   ): void {
+    const {onResult, configurationId} = this.params_;
     this.dialogManager_.completeView(this.activityIframeView_);
     this.entitlementsManager_.clear();
     const userToken = response.getSwgUserToken();
     if (userToken) {
       this.deps_.storage().set(Constants.USER_TOKEN, userToken, true);
     }
+    if (this.isOptIn(this.params_.action) && onResult) {
```
oyj9109, line 191: Out of curiosity, what's the plan to handle non-opt-in cases?

src/api/interventions.ts
```diff
@@ -0,0 +1,109 @@
...
+ */
+export interface Intervention {
+  // Indicates what type of intervention this is.
+  readonly type: InterventionType;
+  // ID used to fetch the configuration for the intervention. IDs are found in
+  // the Publisher Center.
+  readonly configurationId?: string;
+  // Indicates if the intervention should be Google provided, or publisher
+  // provided.
+  readonly preference?: string;
```
oyj9109, line 83: Do we want to expose this now? Not sure if we would want to support BYOP for RRM:E

oyj9109 (approved):
Overall LGTM. Thanks for adding the demo site! I left a few questions.

src/runtime/audience-action-flow.ts
mhkawano: For surveys it is handled in the data transfer request. For rewarded ads it will not be called.

src/runtime/audience-action-flow.ts
mhkawano: The result is used in handleSurveyDataTransferRequest_ as:
```
const dataTransferSuccess = await this.attemptSurveyDataTransfer(request);
```
The result of the callback is used to determine if the data has been stored correctly.

src/api/interventions.ts
mhkawano: Is it a big deal if it's exposed? I can def try to rework it to hide `Intervention` and keep the property out of `AvailableInterventions`.

mhkawano: Chris brought it up before too, so I moved `Intervention` back into the entitlements manager.

PR #3501, opened 2024-05-01
Clean up AutoPromptManager vars to be class attributes

justinchou-google:
b/303489420


oyj9109 (approved):
(No action needed, just leaving a note)
I think ideally, isClosable should be treated as a prompt-level attribute. This change seems fine as we're not going to show multiple prompts within a page.

It might make more sense to rename the attribute to `isLocked` to indicate the content type of the page. We can think about making this change in the upcoming launch where we support per-prompt dismissibility, to distinguish `isClosable` attribute at prompt level and `isLocked` attribute at the page level (which is identical to the class instance).

PR #3514, opened 2024-05-10
Add logging for config id

mhkawano:
Because we can now have multiple prompts of the same type we now need to provide the configuration id for publishers so that they can aggregate their logs by specific prompts.

b/336857804

### DIFF
--- a/src/runtime/analytics-service.ts
+++ b/src/runtime/analytics-service.ts
@@ -286,6 +286,9 @@ export class AnalyticsService {
     const meta = new AnalyticsEventMeta();
     meta.setEventOriginator(event.eventOriginator);
     meta.setIsFromUserAction(!!event.isFromUserAction);
+    if (event.configurationId) {
+      meta.setConfigurationId(event.configurationId);
+    }
     // Update the request's timestamp.
     this.context_.setClientTimestamp(toTimestamp(event.timestamp!));
     const loadEventStartDelay = this.getLoadEventStartDelay_();
--- a/src/runtime/analytics-service.ts
+++ b/src/runtime/analytics-service.ts
@@ -286,9 +286,6 @@ export class AnalyticsService {
     const meta = new AnalyticsEventMeta();
     meta.setEventOriginator(event.eventOriginator);
     meta.setIsFromUserAction(!!event.isFromUserAction);
-    if (event.configurationId) {
-      meta.setConfigurationId(event.configurationId);
-    }
     // Update the request's timestamp.
     this.context_.setClientTimestamp(toTimestamp(event.timestamp!));
     const loadEventStartDelay = this.getLoadEventStartDelay_();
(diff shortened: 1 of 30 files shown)


src/runtime/google-analytics-event-listener.ts
```diff
@@ -111,6 +110,7 @@ export class GoogleAnalyticsEventListener {
         'event_category': gaEvent.eventCategory,
         'event_label': gaEvent.eventLabel,
         'non_interaction': gaEvent.nonInteraction,
+        'configurationId': event.configurationId || '',
```
oyj9109, line 113: I think it's redundant information fo GA logging. The key needs to be manually configured as a custom dimension to be processed.

src/runtime/analytics-service.ts
```diff
@@ -286,6 +286,9 @@ export class AnalyticsService {
     const meta = new AnalyticsEventMeta();
     meta.setEventOriginator(event.eventOriginator);
     meta.setIsFromUserAction(!!event.isFromUserAction);
+    if (event.configurationId) {
+      meta.setConfigurationId(event.configurationId);
```
oyj9109, line 290: It seems redundant unless we have plans to log it in clearcut events. For it to be properly propagated, a conversion logic needs to be added in analytics service here:
http://google3/java/com/google/subscribewithgoogle/client/ui/common/analytics/analytics_service.ts;l=656;rcl=625751917

oyj9109 (commented):
Have you verified your change in local demo? You can register event listeners by calling `registerEventListener` API.

https://github.com/subscriptions-project/swg-js/blob/e2448dd32ee50cdb8f3ba351d9786c1c93314b78/src/api/client-event-manager-api.ts#L48

The function can simply print out propagated event objects and you can check the presence of configId field there.

src/runtime/google-analytics-event-listener.ts
mhkawano: Makes sense. Removed.

src/runtime/analytics-service.ts
mhkawano: I don't think we plan on gathering this information. Will remove.

mhkawano:
> Have you verified your change in local demo? You can register event listeners by calling `registerEventListener` API.
> 
> https://github.com/subscriptions-project/swg-js/blob/e2448dd32ee50cdb8f3ba351d9786c1c93314b78/src/api/client-event-manager-api.ts#L48
> 
> The function can simply print out propagated event objects and you can check the presence of configId field there.

Thanks for the review! I don't have the config ids passed from the newsletter yet, but I used a hard coded value in `activities.ts` and see it propagate to the event handler. I updated the demo site to print out all logs using `registerEventListener`. Fairly sure this works properly, so I will clean up the PR and send it out for review.

oyj9109 (approved):

PR #3525, opened 2024-05-24
Bypass UiPredicates, entitlements and frequency capping when onsite preview should be rendered

kristenwang:
Preview flas was added in cl/636919307.

previewEnabled param was added in Article in cl/634864138.

[go/rrm-onsite-preview-dd](http://goto.google.com/rrm-onsite-preview-dd)

### DIFF
--- a/src/runtime/auto-prompt-manager.ts
+++ b/src/runtime/auto-prompt-manager.ts
@@ -16,6 +16,7 @@
 
 import {AnalyticsEvent, EventOriginator} from '../proto/api_messages';
 import {Article, EntitlementsManager} from './entitlements-manager';
+import {ArticleExperimentFlags} from './experiment-flags';
 import {
   AudienceActionFlow,
   AudienceActionIframeFlow,
@@ -129,6 +130,7 @@ export class AutoPromptManager {
   private lastAudienceActionFlow_: AudienceActionFlow | null = null;
   private isClosable_: boolean | undefined;
   private autoPromptType_: AutoPromptType | undefined;
+  private onsitePreviewEnabled_: boolean = false;
 
   private readonly doc_: Doc;
   private readonly pageConfig_: PageConfig;
@@ -214,6 +216,11 @@ export class AutoPromptManager {
       return;
     }
     // Set experiment flags here.
+    const articleExpFlags =
+      this.entitlementsManager_.parseArticleExperimentConfigFlags(article);
+    this.onsitePreviewEnabled_ = articleExpFlags.includes(
+      ArticleExperimentFlags.ONSITE_PREVIEW_ENABLED
+    );
   }
 
   /**
@@ -230,12 +237,20 @@ export class AutoPromptManager {
       return;
     }
 
-    if (!clientConfig.uiPredicates?.canDisplayAutoPrompt) {
+    const shouldRenderOnsitePreview =
+      article.previewEnabled && this.onsitePreviewEnabled_;
+
+    // If onsite preview should be rendered, canDisplayAutoPrompt will be ignored.
+    if (
+      !shouldRenderOnsitePreview &&
+      !clientConfig.uiPredicates?.canDisplayAutoPrompt
+    ) {
       return;
     }
 
+    // If onsite preview should be rendered, entitlements will be ignored.
     const hasValidEntitlements = entitlements.enablesThis();
-    if (hasValidEntitlements) {
+    if (!shouldRenderOnsitePreview && hasValidEntitlements) {
       return;
     }
 
@@ -259,6 +274,7 @@ export class AutoPromptManager {
     const potentialAction = await this.getPotentialAction_({
       article,
       frequencyCapConfig,
+      shouldRenderOnsitePreview,
     });
     const promptFn = this.isMonetizationAction_(potentialAction?.type)
       ? this.getMonetizationPromptFn_()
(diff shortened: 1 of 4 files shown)


src/runtime/auto-prompt-manager.ts
```diff
@@ -230,12 +237,20 @@ export class AutoPromptManager {
...
+    // If onsite preview should be rendered, canDisplayAutoPrompt will be ignored.
+    if (
+      !shouldRenderOnsitePreview &&
+      !clientConfig.uiPredicates?.canDisplayAutoPrompt
+    ) {
       return;
     }
 
+    // If onsite preview should be rendered, entitlements will be ignored.
     const hasValidEntitlements = entitlements.enablesThis();
```
justinchou-google, line 252: Would it be more readable if we wrapped these checks together so that onsitePreview can skip them all?
```
    if (!shouldRenderOnsitePreview) {
       ...  // Current checks for uiPredicates, entitlements
    }
```

justinchou-google: though I am worried that this flow is starting to get messy like the former flow we deprecated. Perhaps it would be beneficial to break this function up into smaller functions?

kristenwang: I think breaking this function up just for the uiPredicates & entitlements might be too much, I merged the two value into one and simplified the code a little, let me know if this makes sense, thanks!

src/runtime/auto-prompt-manager.ts
```diff
@@ -259,6 +269,7 @@ export class AutoPromptManager {
     const potentialAction = await this.getPotentialAction_({
       article,
       frequencyCapConfig,
+      shouldRenderOnsitePreview,
```
oyj9109, line 272: How much of the remaining flow do we expect to be shared between preview and regular flows moving forward? 
Given that the preview flow is much simpler than the regular flow, would it be better to create a separate flow from the begining which would help maintain individual flow cleaner?

kristenwang: Created a separate flow for preview prompt.

src/runtime/auto-prompt-manager.ts
```diff
@@ -214,6 +223,50 @@ export class AutoPromptManager {
...
+    article: Article,
+    params: ShowAutoPromptParams
+  ): Promise<void> {
+    const actions = article.audienceActions?.actions;
+    if (!actions || actions.length === 0) {
+      return;
+    }
+    // Article response is honored over code snippet in case of conflict, such
+    // as when publisher changes revenue model but does not update snippet.
+    this.autoPromptType_ = this.getAutoPromptType_(
```
oyj9109, line 246: I'm not sure if this part is necessary. If it was added for the closabilty fallback logic, see my comment below.

src/runtime/auto-prompt-manager.ts
```diff
@@ -214,6 +223,50 @@ export class AutoPromptManager {
...
+      article.audienceActions?.actions,
+      params.autoPromptType
+    )!;
+
+    // Default isClosable to what is set in the page config.
+    // Otherwise, the prompt is blocking for publications with a
+    // subscription revenue model, while all others can be dismissed.
+    this.isClosable_ = params.isClosable ?? !this.isSubscription_();
```
oyj9109, line 254: Hi @justinchou-google, I'm not sure if this fallback is still necessary (for both preview and regular flow) after we added a logic to drive this value based off product id in the first place. Do you think we can clean this up?

src/runtime/auto-prompt-manager.ts
```diff
@@ -214,6 +223,50 @@ export class AutoPromptManager {
...
+
+    const potentialAction = actions[0];
```
oyj9109, line 256: nit: previewAction

At the point, it's clear it should be displayed.

src/runtime/auto-prompt-manager.ts
```diff
@@ -214,6 +223,50 @@ export class AutoPromptManager {
...
+      ? this.getMonetizationPromptFn_()
+      : this.getAudienceActionPromptFn_({
+          actionType: potentialAction.type,
+          configurationId: potentialAction.configurationId,
+          preference: potentialAction.preference,
+        });
+
+    // Set delay for preview prompt to be 0 at first, and change if we need to.
+    const displayDelayMs = 0;
+    this.deps_.win().setTimeout(promptFn, displayDelayMs);
```
oyj9109, line 268: nit: can we directly invoke `promptFn`?

src/runtime/auto-prompt-manager.ts
```diff
@@ -214,6 +223,50 @@ export class AutoPromptManager {
...
+    )!;
+
+    // Default isClosable to what is set in the page config.
+    // Otherwise, the prompt is blocking for publications with a
+    // subscription revenue model, while all others can be dismissed.
+    this.isClosable_ = params.isClosable ?? !this.isSubscription_();
+
+    const potentialAction = actions[0];
+
+    const promptFn = this.isMonetizationAction_(potentialAction?.type)
```
oyj9109, line 258: Can we make it as a function if it's a shared logic between both flows?

oyj9109 (approved):
Overall looks great! I like how simple the preview flow is. Just left a few comments for nit.

src/runtime/auto-prompt-manager.ts
kristenwang: Done

src/runtime/auto-prompt-manager.ts
kristenwang: Done.

src/runtime/auto-prompt-manager.ts
kristenwang: Done.

src/runtime/auto-prompt-manager.ts
kristenwang: Both AudienceActionLocalFlow & AudienceActionIframeFlow are using autoPromptType_ & isClosable_ values, and the two values are used in other places as well. Not gonna change them all in this pr, we can probably clean them up later.

src/runtime/auto-prompt-manager.ts
kristenwang: Ditto.

PR #3527, opened 2024-06-06
Set prompt language from element lang

mhkawano:
Tested locally with `<html lang="fr-FR">` and `<html lang="fr">`.

b/345212480

### DIFF
--- a/src/runtime/runtime-test.js
+++ b/src/runtime/runtime-test.js
@@ -392,6 +392,16 @@ describes.realWin('Runtime', (env) => {
         .getClientConfig();
       expect(clientConfig.paySwgVersion).to.eq('123');
     });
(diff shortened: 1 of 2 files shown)


src/runtime/runtime-test.js
```diff
@@ -392,6 +392,16 @@ describes.realWin('Runtime', (env) => {
         .getClientConfig();
       expect(clientConfig.paySwgVersion).to.eq('123');
     });
+
+    it('sould set ClientOptions based on root element language', async () => {
```
justinchou-google, line 396:
```suggestion
    it('should set ClientOptions based on root element language', async () => {
```

oyj9109 (approved):

PR #3533, opened 2024-07-16
Update rewarded ad narrow screen width

mhkawano:
Update the width on small screens to take up 100% even if the content is small.

b/353509279

Before: https://github.com/user-attachments/assets/f7a27e1c-b6e7-4e51-ab48-14a6603d077b
After: https://github.com/user-attachments/assets/33f2d019-0581-4389-960e-2c5b42f10228

### DIFF
--- a/demos/public/prod/enterprise/enterprise-funnel.html
+++ b/demos/public/prod/enterprise/enterprise-funnel.html
@@ -61,8 +61,6 @@
         `;
         row.innerHTML = html;
         table.appendChild(row);
-
-        console.log(result);
       }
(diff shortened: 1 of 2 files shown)


justinchou-google:
The "after" definitely looks better than the "before" in my opinion, but have we discussed with Clara/Milind what the mobile view should look like?

mhkawano:
I believe that was the intended behavior, but it just doesn't work in this case. Will send them a message confirming.

justinchou-google (approved):

PR #3540, opened 2024-08-09
FPA M0.5 Adds contentType to article query param.

justinchou-google:
Infers contentType from page config isLocked.

b/359271461


src/runtime/entitlements-manager.ts
```diff
@@ -823,6 +824,18 @@ export class EntitlementsManager {
     if (this.useArticleEndpoint_) {
       url = addQueryParam(url, 'locked', String(this.pageConfig_.isLocked()));
     }
+    const experimentFlags = await this.getExperimentConfigFlags();
+    if (
+      experimentFlags.includes(
+        ArticleExperimentFlags.CONTENT_TYPE_ARTICLE_PARAM_ENABLED
```
justinchou-google, line 830: hi @oyj9109 , I realized that these experiment flags are returned from the article response itself, and so aren't available prior to the /article call. 

Do you think these changes are minimally invasive enough to not require a flag? If we do want to ramp up, I can look into adding this flag at client startup.

oyj9109: We have another mechanism to ramp up experiment flag that's available from the beginning of runtime. See cl/476477078 and https://github.com/subscriptions-project/swg-js/commit/c62001405c01b33350f0a58ef0e7d3cef604339f for references.

Experiment flag returned from article response is meant to be used to co-ordinate rolling out changes between swg.js and swg-internal server, mostly with the diversion based on user token.

That being said, I think this change adding an additional query param here doesn't break anything itself and is simple enough to submit without flag guarding. And putting it in place early is benefitial to have more version coverage, given that old swg.js versions can be live due to caching. So I agree ramp-up is not required.

PR #3542, opened 2024-08-15
Rewarded ads rrme

mhkawano:
Enable rewarded ads for RRME.

Demo: b/358588957


src/runtime/audience-action-local-flow.ts
```diff
@@ -599,6 +611,7 @@ export class AudienceActionLocalFlow implements AudienceActionFlow {
       AnalyticsEvent.ACTION_REWARDED_AD_CLOSE,
       /* isFromUserAction */ true
     );
+    this.rewardedResult(true, false);
```
oyj9109, line 614: nit: add parameter name comments so that it's understanable without having to check the function definition.

See go/tsjs-style#comments-when-calling-a-function
“Parameter name” comments should be used whenever the method name and parameter value do not sufficiently convey the meaning of the parameter.

src/runtime/audience-action-local-flow.ts
```diff
@@ -558,9 +561,12 @@ export class AudienceActionLocalFlow implements AudienceActionFlow {
       AnalyticsEvent.ACTION_REWARDED_AD_CLOSE_AD,
       /* isFromUserAction */ true
     );
+    this.rewardedResult(true, false);
```
oyj9109, line 564: same as below

src/runtime/audience-action-local-flow.ts
```diff
@@ -787,10 +800,29 @@ export class AudienceActionLocalFlow implements AudienceActionFlow {
...
       this.closeRewardedAdWall_();
-    } else if (this.params_.action === TYPE_NEWSLETTER_SIGNUP) {
+    } else if (
+      this.params_.action === InterventionType.TYPE_NEWSLETTER_SIGNUP
+    ) {
       this.closeOptInPrompt_();
     }
   }
+
+  private rewardedResult(
```
oyj9109, line 812: The function name doesn't seem to convey what exactly it does. Can you rename it to be more specific on what it does?

oyj9109 (approved):

PR #3543, opened 2024-08-19
FPA M0.5

justinchou-google:
Adds a sequence flow to AutoPromptManager to process actionOrchestration from the article endpoint. 

- Defines many more interfaces from article response.
- Continues to store and fetch frequency capping events by ACTION TYPE.
- Maintains current fallback logic with default frequency cap configs.
- Computes `contentType` (to determine which funnel to fetch) equivalent to what is written to the `pageConfig`. NOTE: this differs slightly from the current prod version's computation of `isClosable`, which also falls back to revenue model.
  - **Question**: how do we want to update this definition for OnsitePreview? It currently uses this definition of `isClosable`.
  - Follow up item to update Onsite Preview (b/362347951).
- `CLOSED` contentType explicitly skips over FrequencyCapping logic, even if FrequencyCapping values are provided. I believe this is the only contentType specific logic in the flow.
- Defines new `secondsDuration` proto on the `frequencyCapDuration`: this will be validated and converted as part of `getTargetedInterventionFunnel` Action (TODO b/362344744).
- Fixes `nano` implementation from Duration.

### DIFF
--- a/src/api/basic-subscriptions.ts
+++ b/src/api/basic-subscriptions.ts
@@ -99,6 +99,15 @@ export enum AutoPromptType {
 }
 
 /**
+ * The types of supported publication content type. Should be maintained with:
+ * google3/java/com/google/subscribewithgoogle/audienceactions/proto/audience_member_flow_service.proto.
+ */
+export enum ContentType {
+  OPEN = 'OPEN',
+  CLOSED = 'CLOSED',
+}
+
+/**
  * Options for configuring all client UI.
  */
 export interface ClientOptions {
--- a/src/runtime/auto-prompt-manager.ts
+++ b/src/runtime/auto-prompt-manager.ts
@@ -199,7 +199,10 @@ export class AutoPromptManager {
       this.autoPromptType_ = this.getPromptTypeToDisplay_(
         params.autoPromptType
       );
-      this.isClosable_ = params.isClosable ?? !this.isSubscription_();
+      this.isClosable_ = this.actionOrchestrationExperiment_
+        ? params.contentType === ContentType.OPEN
+        : params.isClosable ?? !this.isSubscription_();
+
       const promptFn = this.getMonetizationPromptFn_();
       promptFn();
       return;
--- a/src/runtime/auto-prompt-manager.ts
+++ b/src/runtime/auto-prompt-manager.ts
@@ -517,12 +517,7 @@ export class AutoPromptManager {
     const eligibleActions = article.audienceActions?.actions;
     let targetedInterventions =
       article.actionOrchestration?.interventionFunnel?.prompts;
-    if (
-      !eligibleActions ||
-      eligibleActions.length === 0 ||
-      !targetedInterventions ||
-      targetedInterventions.length === 0
-    ) {
+    if (!eligibleActions?.length || !targetedInterventions?.length) {
       return;
     }
 
--- a/src/runtime/auto-prompt-manager.ts
+++ b/src/runtime/auto-prompt-manager.ts
@@ -133,6 +133,7 @@ export class AutoPromptManager {
   private autoPromptType_: AutoPromptType | undefined;
   private onsitePreviewEnabled_: boolean = false;
   private shouldRenderOnsitePreview_: boolean = false;
+  private actionOrchestrationExperiment_: boolean = false;
 
   private readonly doc_: Doc;
   private readonly pageConfig_: PageConfig;
@@ -225,10 +226,13 @@ export class AutoPromptManager {
       return;
     }
     // Set experiment flags here.
-    const articleExpFlags =
-      this.entitlementsManager_.parseArticleExperimentConfigFlags(article);
-    this.onsitePreviewEnabled_ = articleExpFlags.includes(
+    this.onsitePreviewEnabled_ = this.isArticleExperimentEnabled_(
+      article,
       ArticleExperimentFlags.ONSITE_PREVIEW_ENABLED
+    );
+    this.actionOrchestrationExperiment_ = this.isArticleExperimentEnabled_(
+      article,
+      ArticleExperimentFlags.ACTION_ORCHESTRATION_EXPERIMENT
     );
   }
 
@@ -865,5 +869,18 @@ export class AutoPromptManager {
           configurationId: action.configurationId,
           preference: action.preference,
         });
+  }
+
+  /**
+   * Checks if provided ExperimentFlag is enabled within article experiment
+   * config.
+   */
+  private isArticleExperimentEnabled_(
+    article: Article,
+    experimentFlag: string
+  ): boolean {
+    const articleExpFlags =
+      this.entitlementsManager_.parseArticleExperimentConfigFlags(article);
+    return articleExpFlags.includes(experimentFlag);
   }
 }
--- a/src/runtime/auto-prompt-manager.ts
+++ b/src/runtime/auto-prompt-manager.ts
@@ -31,8 +31,6 @@ import {
   Closability,
   InterventionFunnel,
   InterventionOrchestration,
-  SwgDuration,
-  SwgDurationUnit,
 } from '../api/action-orchestration';
 import {ConfiguredRuntime} from './runtime';
 import {Deps} from './deps';
--- a/src/runtime/auto-prompt-manager.ts
+++ b/src/runtime/auto-prompt-manager.ts
@@ -328,6 +328,9 @@ export class AutoPromptManager {
             this.isClosable_ =
               params.contentType === ContentType.CLOSED ? false : true;
         }
+        console.log(
+          `setting isclosable: ${nextIntervention.closability} ${this.isClosable_}`
+        );
         potentialAction = article.audienceActions?.actions?.find(
           (action) => action.configurationId === nextIntervention.configId
         );
@@ -561,6 +564,7 @@ export class AutoPromptManager {
         clientConfig.autoPromptConfig?.frequencyCapConfig
       )
     ) {
+      console.log('freq cap is invalid');
       this.eventManager_.logSwgEvent(
         AnalyticsEvent.EVENT_FREQUENCY_CAP_CONFIG_NOT_FOUND_ERROR
       );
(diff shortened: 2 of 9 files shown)


src/runtime/basic-runtime.ts
```diff
@@ -243,6 +244,7 @@ export class BasicRuntime implements BasicSubscriptions {
       autoPromptType,
       alwaysShow,
       isClosable,
+      contentType: this.getContentType_(isAccessibleForFree ?? isOpenAccess),
```
justinchou-google, line 247: Want to call this out as a point we should reconfirm with Product — specifically, what defines ContentType?

Previously, AutoPromptManager synonymously used "openaccess content" to mean "content with dismissible prompts", and computed dismissibility based on: isAccessibleForfree (if true), else fallback to `openaccess` productId (if true), else fallback to contribution/subscription model. 

In the above implementation, I have removed the last check for revenue model, and just computed contentType on isAccessibleForFree (if true), else fallback to `openaccess` productId.

src/runtime/auto-prompt-manager.ts
```diff
@@ -813,28 +973,56 @@ export class AutoPromptManager {
...
+    frequencyCapConfig: FrequencyCapConfig,
+    interventionFunnel: InterventionFunnel
+  ): Duration | undefined {
+    const swgDuration = interventionFunnel.globalFrequencyCap?.duration;
+    return swgDuration
+      ? this.convertSwgDurationToSeconds_(swgDuration)
+      : frequencyCapConfig.globalFrequencyCap!.frequencyCapDuration;
+  }
+
+  private convertSwgDurationToSeconds_(
```
justinchou-google, line 1001: This conversion used to occur on the server side, but since we defined the article response proto to match the storage proto, I am now bringing this conversion to the client side.

src/runtime/auto-prompt-manager.ts
```diff
@@ -813,28 +973,56 @@ export class AutoPromptManager {
...
+    swgDuration: SwgDuration
+  ): Duration | undefined {
+    const invalidUnits = [
+      SwgDurationUnit.SECOND,
+      SwgDurationUnit.MONTH,
+      SwgDurationUnit.YEAR,
+    ];
+    if (invalidUnits.includes(swgDuration.unit)) {
+      // TODO: Log error
```
justinchou-google, line 1010: Throughout this PR, a few new errors should be logged. I wonder if we can reuse some existing events [(event for prompt frequency not found)](https://source.corp.google.com/piper///depot/google3/subscribewithgoogle/api/messages/api_messages.proto;l=781) for this error (unsupported unit). Previously, when this was calculated on the server, an exception was thrown.

src/runtime/auto-prompt-manager.ts
```diff
@@ -813,28 +973,56 @@ export class AutoPromptManager {
...
+      case SwgDurationUnit.MINUTE:
+        return new Duration(60 * swgDuration.count, 0);
+      case SwgDurationUnit.HOUR:
+        return new Duration(3600 * swgDuration.count, 0);
+      case SwgDurationUnit.DAY:
+        return new Duration(86400 * swgDuration.count, 0);
+      case SwgDurationUnit.WEEK:
+        return new Duration(604800 * swgDuration.count, 0);
+      default:
+        // TODO: log error
```
justinchou-google, line 1023: Same as above message.

src/runtime/auto-prompt-manager.ts
```diff
@@ -813,28 +973,56 @@ export class AutoPromptManager {
...
+  ): Duration | undefined {
+    const swgDuration = interventionOrchestration.promptFrequencyCap?.duration;
+
+    if (!swgDuration) {
+      this.eventManager_.logSwgEvent(
+        AnalyticsEvent.EVENT_PROMPT_FREQUENCY_CONFIG_NOT_FOUND
+      );
+      return frequencyCapConfig.anyPromptFrequencyCap?.frequencyCapDuration;
+    }
+    return this.convertSwgDurationToSeconds_(swgDuration);
```
justinchou-google, line 988: This follows the current flow: if the prompt frequency is PRESENT but INVALID (unsupported unit), frequency capping will not be calculated at all, and the first intervention will be displayed.

When revisiting this logic, I thought maybe the better experience would be to _still_ fallback to the default AnyPromptFrequencyConfig.

src/runtime/auto-prompt-manager.ts
```diff
@@ -813,28 +973,56 @@ export class AutoPromptManager {
...
+  }
+
+  private getGlobalFrequencyCapDuration_(
+    frequencyCapConfig: FrequencyCapConfig,
+    interventionFunnel: InterventionFunnel
+  ): Duration | undefined {
+    const swgDuration = interventionFunnel.globalFrequencyCap?.duration;
+    return swgDuration
+      ? this.convertSwgDurationToSeconds_(swgDuration)
```
justinchou-google, line 997: Same as above, if the provided globalFrequencyCap is present but invalid, then the current logic is to NOT frequency cap the prompt. Should we revisit this logic to fallback to the default?

src/runtime/basic-runtime.ts
```diff
@@ -243,6 +244,7 @@ export class BasicRuntime implements BasicSubscriptions {
       autoPromptType,
       alwaysShow,
       isClosable,
```
justinchou-google, line 246: @kristenwang Are we ok updating this value for OnsitePreview to also use `contentType`? I believe in today's flow we use `isClosable`, but this will be deprecated.

src/runtime/auto-prompt-manager.ts
```diff
@@ -467,6 +501,110 @@ export class AutoPromptManager {
...
+      );
+      return targetedInterventions[0];
+    }
+
+    for (const intervention of targetedInterventions) {
+      const frequencyCapDuration = this.getPromptFrequencyCapDuration_(
+        clientConfig.autoPromptConfig?.frequencyCapConfig!,
+        intervention
+      );
+      if (this.isValidFrequencyCapDuration_(frequencyCapDuration)) {
```
justinchou-google, line 563: Note: much of the computation for whether a prompt is FrequencyCapped (or if the flow is globally capped) is identical to the old flow.

src/runtime/basic-runtime.ts
justinchou-google: Spoke offline, the fallbacks we had were specific eng decisions we put in place. Moving forward, `contentType` will solely be calculated from `isAccessibleForFree` and `productId`.

src/runtime/basic-runtime.ts
justinchou-google: Note: we can no longer assume the contribution prompt will be present in the audience actions for all contribution revenue models.

src/runtime/auto-prompt-manager.ts
```diff
@@ -190,7 +199,10 @@ export class AutoPromptManager {
       this.autoPromptType_ = this.getPromptTypeToDisplay_(
         params.autoPromptType
       );
-      this.isClosable_ = params.isClosable ?? !this.isSubscription_();
+      this.isClosable_ = this.actionOrchestrationExperiment_
```
oyj9109, line 202: Should we move up `setArticleExperimentFlags_` invocation so that the flag value is set at this point?

src/runtime/auto-prompt-manager.ts
```diff
@@ -294,20 +309,42 @@ export class AutoPromptManager {
...
+    if (this.actionOrchestrationExperiment_ && !!article.actionOrchestration) {
+      // FPA M0.5 Flow: get next Intervention of the Targeted Funnel.
+      const nextIntervention = await this.getTargetedInterventionOrchestration(
+        clientConfig,
+        article,
+        params.contentType
+      );
+
+      if (!!nextIntervention) {
+        // Fallback to content type for unspecified?
```
oyj9109, line 322: Sounds like a good idea. What do you think about organizing the logic using switch-case?

src/runtime/auto-prompt-manager.ts
```diff
@@ -467,6 +504,110 @@ export class AutoPromptManager {
...
+    eligibleActions = eligibleActions.filter((action) =>
+      this.checkActionEligibility_(action.type, actionsTimestamps!)
+    );
+    if (eligibleActions.length === 0) {
+      return;
+    }
+
+    // Filter the funnel of interventions by eligibility.
+    targetedInterventions = targetedInterventions.filter((intervention) =>
+      eligibleActions
```
oyj9109, line 535: nit: looks like it's dealing with array conversion every iteration within outer filter loop. What about creating a set of eligible config ids instead?

src/runtime/auto-prompt-manager.ts
```diff
@@ -813,28 +979,56 @@ export class AutoPromptManager {
...
+    const swgDuration = interventionFunnel.globalFrequencyCap?.duration;
+    return swgDuration
+      ? this.convertSwgDurationToSeconds_(swgDuration)
+      : frequencyCapConfig.globalFrequencyCap!.frequencyCapDuration;
+  }
+
+  private convertSwgDurationToSeconds_(
+    swgDuration: SwgDuration
+  ): Duration | undefined {
+    const invalidUnits = [
```
oyj9109, line 1010: I'd recommend validating configuration on the server side if possible, so that the client side is more generally implemented and agnostic to product details and we rahter have more control on the server side.

src/runtime/auto-prompt-manager.ts
oyj9109: Assuming we can handle validaty checks on the server side (see my comment below), skipping frequency capping for invalid/absent frequency value makes sense to me.

src/runtime/auto-prompt-manager.ts
```diff
@@ -294,20 +309,42 @@ export class AutoPromptManager {
...
+
+      if (!!nextIntervention) {
+        // Fallback to content type for unspecified?
+        this.isClosable_ =
+          nextIntervention?.closability === Closability.BLOCKING;
+        potentialAction = article.audienceActions?.actions?.find(
+          (action) => action.configurationId === nextIntervention.configId
+        );
+      }
+    } else {
```
oyj9109, line 329: Just to clarify, are there cases where flag is enabled but `article.actionOrchestration` is absent?

oyj9109:
Thanks for thoroughly documented description. The overall logic looks good and I left some minor comments.

Regarding some questions in the description

> Question: how do we want to update this definition for OnsitePreview? It currently uses this definition of isClosable.

I feel like having a fallback value based on content type would generally make sense in case prompt level dimissibility is absent (e.g., no action orchestration available for onsite preview).

src/api/basic-subscriptions.ts
```diff
@@ -98,6 +98,15 @@ export enum AutoPromptType {
   SUBSCRIPTION_LARGE = 'subscription_large',
 }
 
+/**
+ * The types of supported publication content type. Should be maintained with:
+ * google3/java/com/google/subscribewithgoogle/audienceactions/proto/audience_member_flow_service.proto.
+ */
+export enum ContentType {
```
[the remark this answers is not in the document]
oyj9109: Makes sense. Thanks for clarifying!

src/runtime/basic-runtime-test.js
```diff
@@ -318,54 +318,63 @@ describes.realWin('BasicRuntime', (env) => {
...
         isClosable: undefined,
+        contentType: 'CLOSED',
       },
-    ].forEach(({isAccessibleForFree, isPartOfProductId, isClosable}) => {
-      it(`shows autoPrompt with isClosable=${isClosable} when isAccessibleForFree=${isAccessibleForFree} and isPartOfProductId=${isPartOfProductId}`, async () => {
-        const setupAndShowAutoPromptStub = sandbox.stub(
-          basicRuntime,
-          'setupAndShowAutoPrompt'
-        );
+    ].forEach(
```
oyj9109, line 353: I really love to see how extensible the original test implementation was

oyj9109 (approved):
Looks great!

PR #3553, opened 2024-08-30
Add interventionOrchestration eligibility checks based on Repeatability

justinchou-google:
b/358646155

Implements swg.js client side processing of number of completions for repeatable actions, using Repeatability defined in the funnel-level prompt config.

Updates some variable names to be more consistent.


src/runtime/auto-prompt-manager.ts
```diff
@@ -961,6 +971,37 @@ export class AutoPromptManager {
...
+  private checkOrchestrationEligibility_(
+    orchestration: InterventionOrchestration,
+    eligibleActionIds: Set<string | undefined>,
+    article: Article
+  ): boolean {
+    if (!eligibleActionIds.has(orchestration.configId)) {
+      return false;
+    }
+    if (
+      this.isRepeatableIntervention_(orchestration.type) &&
```
oyj9109, line 987: Do we need to keep maintaining this list in the client side or can we defer it to the server-side such that repeatability values are set up properly for repeatable actions types?

justinchou-google: Agreed, removing this check and deferring to the server to handle repeatabiltiy type checks.

PR #3555, opened 2024-09-03
FPA M0.5: Onsite Preview updates

justinchou-google:
b/362347951

Updates onsite preview to use contentType to determine closability for FPA M0.5.

Also cleans up onsite preview experiment rampup flags, fixes some dead references, and adds a TODO (b/364344782) to revisit this logic with new PM for M1+.


kristenwang (approved):

src/runtime/auto-prompt-manager.ts
```diff
@@ -262,10 +256,13 @@ export class AutoPromptManager {
...
 
-    // Default isClosable to what is set in the page config.
-    // Otherwise, the prompt is blocking for publications with a
-    // subscription revenue model, while all others can be dismissed.
-    this.isClosable_ = params.isClosable ?? !this.isSubscription_();
+    // Default isClosable to what is set in the page config and fallback to
+    // if the prompt to preview is a subscription prompt.
+    // For FPA M0.5 - default to the contentType.
+    // TODO(b/364344782): Determine closability for FPA M1+.
+    this.isClosable_ = this.actionOrchestrationExperiment_
```
oyj9109, line 263: Can we move it up to a higher level so that the same logic is shared between showPreviewAutoPrompt_ and showAutoPrompt_?

justinchou-google: I guess this logic is shared with `showAutoPrompt_` for the default case when Closability is not specified, but `showAutoPrompt_` should first read from the `InterventionOrchestration`'s Closability.

I can refactor this so that the default case is shared, but wanted to call out that the logic is not 100% the same. Another downside is that `isClosable_` computation will be fragmented into two different locations. WDYT?

oyj9109: Thanks for clarifying! Since it's a temporary state during the migration, having isClosable_ computation individually and clean up the one in showAutoPrompt_ sounds good to me.

oyj9109 (approved):

src/runtime/auto-prompt-manager.ts
oyj9109: Disregard my previous comment regarding migration. I was confused with something else.

Is params.contentType being set regardless of the experiment flag value? If so, can we simply determine isClosable_ based on that?

In addition, I'm not sure why fallback to revenue model is necessary.

justinchou-google: > Is params.contentType being set regardless of the experiment flag value? 
Yes.
> If so, can we simply determine isClosable_ based on that?
Do you mean remove the experiment check and just replace the entire expression right now? I agree this would simplify everything. One thing to note, though, is that the `parms.isClosable` calculation is slightly more complicated than `productId = openaccess`. It allows for some null values (and hence falls-back to revenue model).

> In addition, I'm not sure why fallback to revenue model is necessary.
Yes, I noticed this in my discussion with Kristen, too. It seems like this is actually only applicable to the case where the Subscription prompt is being previewed. 

Overall, I am in favor of moving to a world where, in the absence of `Action Orchestration`, ContentType determines the dismissibility. One question I have is, are we ok making those changes now before we fully launch FPA?

oyj9109: Yes, simplifying this for onsite-preview first would be great. 

I think fallback to revenue model was copied over from the original logic. I'm not sure if it's still necessary for the regular flow. I think it should be fine to clean up the legacy at least for onsite preview.

justinchou-google: I created b/364344782 to explore how we want onsitePreview to work for M1+, do you think this is needed / if htis simplification is good to keep?

kristenwang (approved):

src/runtime/auto-prompt-manager.ts
justinchou-google: Followed up offline, for simplicity, we can remove any revenue model fallback and just use contentType.

PR #3572, opened 2024-09-20
Seperate the canonicalTag fetch from `getCanonicalUrl` to seperately log the canonicalTag from markup.

caroljli:
For b/360141766

Want to separate out the full URL to log the path generated from the `canonicalRef` of markup (if available). The current result of `getCanonicalUrl(...)` is already being logged as `AnalyticsContext.url`. This will be used for analyzing mismatching domains logged for publications under `AnalyticsContext.url` (ie. logging domains that don't fall under or start with any of the publication's `verified_domains`).


mhkawano:
Is there a reason you don't want to use the origin + path as a default? Seems like it would be a solid fallback as I imagine it's not uncommon to be missing a canonical link. Some context and an associated bug would be nice.

caroljli:
> Is there a reason you don't want to use the origin + path as a default? Seems like it would be a solid fallback as I imagine it's not uncommon to be missing a canonical link. Some context and an associated bug would be nice.

good point, here's the bug: b/360141766 

Basically I want to log what is appearing for the `canonicalRef` tag to start since path should be the *true* URL, since somehow we are logging mismatching URLs for the publication (ie. not starting with any of the verified domains). So I separated them out in order to correctly log just the canonicalRef (the full URL from `getCanonicalDomain` is already being logged with the context `url`).

caroljli:
Updated the description also, PTAL @mhkawano

mhkawano (approved):
LGTM. Thanks for the explanation.

PR #3575, opened 2024-09-25
Add a publicationId suffix to storage key when enabled by flag

oyj9109:
In order to partition browser storage access across different publications (if any under the same domain), we've decied to add a publicationId suffix to storage keys.

This pull request implements the proposed change to existing storage keys by default, but it also allows opting out by keeping a separate list of keys in `StorageKeysWithoutPublicationIdSuffix`. An example is the storage key for PPS signals, as legacy code snippets with fixed storage key values are shared externally.

It also include refactoring to consolidate constants for storage keys into `StorageKeys`.

Design doc: go/sut-pub-id-validation-1-pager

Demo: https://screencast.googleplex.com/cast/NTMyOTkzMTA0NjIyMzg3Mnw4OTFmOWE1Yy1kMA
Verifying two scenarios:
- Senario 1: Control -> Experiment transition: Existing data for old keys is carried over to new keys.
- Senario 2: Experiment -> Control transition: Once the user has been exposed to experiment and thus have a value present for the new key, it continues to use the new key even if the user is in the control group.


src/runtime/storage.ts
```diff
@@ -26,23 +29,42 @@ const TWO_WEEKS_IN_MILLIS = 2 * 604800000;
...
+
+    const valueWithNewKey = await this.getInternal_(newKey, useLocalStorage);
+    if (valueWithNewKey !== null) {
+      return valueWithNewKey;
+    } else {
+      return this.getInternal_(oldKey, useLocalStorage);
+    }
+  }
+
+  private getInternal_(
```
mhkawano, line 56: Is there a reason this needs to be a promise? Seems like you could just `return storage.getItem(finalKey)` or null.

mhkawano: ditto on the set and delete methods

src/runtime/storage.ts
```diff
@@ -26,23 +29,42 @@ const TWO_WEEKS_IN_MILLIS = 2 * 604800000;
...
+    this.pageConfig_ = pageConfig;
     this.values_ = {};
   }
 
-  get(key: string, useLocalStorage = false): Promise<string | null> {
-    if (!this.values_[key]) {
-      this.values_[key] = new Promise((resolve) => {
+  async get(baseKey: string, useLocalStorage = false): Promise<string | null> {
+    // The old version of storage key without publication identifier.
+    // To be deprecaed in favor of the new version of key.
```
justinchou-google, line 43:
```suggestion
    // To be deprecated in favor of the new version of key.
```

src/runtime/storage.ts
```diff
@@ -52,18 +74,49 @@ export class Storage {
...
+    return this.values_[finalKey];
+  }
+
+  async set(
+    baseKey: string,
+    value: string,
+    useLocalStorage = false
+  ): Promise<void> {
+    // The old version of storage key without publication identifier.
+    // To be deprecaed in favor of the new version of key.
```
justinchou-google, line 86:
```suggestion
    // To be deprecated in favor of the new version of key.
```

src/runtime/storage.ts
```diff
@@ -52,18 +74,49 @@ export class Storage {
...
+    const oldKey = this.getStorageKeyWithoutPublicationId_(baseKey);
+    // The new version of storage key with publication identifier.
+    const newKey = this.getStorageKeyMaybeWithPublicationId_(baseKey);
+    const valueWithNewKey = await this.getInternal_(newKey, useLocalStorage);
+
+    if (
+      valueWithNewKey !== null ||
```
justinchou-google, line 93: Can we add a comment that we explicitly also want to honor the new key when the experiment is off? and same for `remove` function below?

justinchou-google (approved):

src/runtime/storage.ts
oyj9109: It was copied over to encapsulate the original implementation to deal with storage interaction and caching. It looks like it caches the promises for retrieveing values in the 'values_' variable, not directly the values themselves.
I don't have context on why it's originally implemented to cache promises, but I think maintaining consistent return types for these internal methods (returning a promise) aligns with the cached values and simplifies the logic as it can just return the cached value if present.
I'm open to suggestions for refactoring this code. If you have any ideas for improvement, I'd be happy to follow up.

PR #3583, opened 2024-10-17
Use publication revenue model as backup to determine UX

mhkawano:
AutoPromptType is determined by looking at the enabled prompts. With on site preview it only returns the single action, so this uses the publication's revenue model as a backup. More details in the bug.

b/362767324


src/runtime/audience-action-local-flow.ts
```diff
@@ -356,17 +361,21 @@ export class AudienceActionLocalFlow implements AudienceActionFlow {
...
   }
 
-  private isSubscription() {
+  private isSubscription(): boolean {
     return (
-      this.params_.autoPromptType == AutoPromptType.SUBSCRIPTION ||
-      this.params_.autoPromptType == AutoPromptType.SUBSCRIPTION_LARGE
+      this.params_.autoPromptType === AutoPromptType.SUBSCRIPTION ||
+      this.params_.autoPromptType === AutoPromptType.SUBSCRIPTION_LARGE ||
+      (!this.params_.autoPromptType &&
```
justinchou-google, line 368: Can we add a comment that this expression is for onsite preview post FPA launch?

Just thinking aloud, after FPA is fully ramped up, does our current definition of `autoPromptType` need to be revisited? Are we able to solely rely on revenue model, at least for `AudienceActionLocalFlow`? If that's the case, maybe the comment should also have a TODO to clean these functions up?

justinchou-google (approved):

src/runtime/audience-action-local-flow.ts
mhkawano: The problem exists now for onsite preview, and this change should address the issue until FPA, at which point we may have contribution pubs without contributions prompts(or subscription). But as discussed in the bug, I think this is a minor edge case. i will make a todo in our backlog.

PR #3592, opened 2024-11-01
[FPA M0.5] Handle number_of_completions count when CTA Repeatability is not specified.

justinchou-google:
b/376451935

After enabling a CTA as repeatable, it will continue to be returned in `audienceActions` of the article response even after completions. As such, even if a CTA config is missing any repeatabiltiy config, we still need to honor any number of completions recorded.

Added some fallback logic for `FINITE` repeatability count.

### DIFF
--- a/src/runtime/auto-prompt-manager-test.js
+++ b/src/runtime/auto-prompt-manager-test.js
@@ -4753,6 +4753,19 @@ describes.realWin('AutoPromptManager', (env) => {
       expect(isEligible).to.be.true;
     });
 
+    it('Orchestration with finite eligibility defaults to 1 eligible completion', async () => {
+      const isEligible = autoPromptManager.checkOrchestrationEligibility_(
+        {
+          configId: 'action_id',
+          type: 'TYPE_REWARDED_AD',
+          repeatability: {type: 'FINITE'},
+        },
+        new Set(['action_id']),
+        new Map([['action_id', 1]])
+      );
+      expect(isEligible).to.be.false;
+    });
+
     it('Repeatable Orchestration with infinite repeatability is eligible with completions', async () => {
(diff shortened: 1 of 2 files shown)


src/runtime/auto-prompt-manager.ts
```diff
@@ -1004,18 +1004,15 @@ export class AutoPromptManager {
     if (!eligibleActionIds.has(orchestration.configId)) {
       return false;
     }
-    if (
-      !!orchestration.repeatability &&
-      orchestration.repeatability.type != RepeatabilityType.INFINITE
-    ) {
+
+    if (orchestration.repeatability?.type != RepeatabilityType.INFINITE) {
```
mhkawano, line 1008: nit: can we use strict inequality?

src/runtime/auto-prompt-manager-test.js
```diff
@@ -4753,6 +4753,19 @@ describes.realWin('AutoPromptManager', (env) => {
       expect(isEligible).to.be.true;
     });
 
+    it('Orchestration with finite eligibility defaults to 1 eligible completion', async () => {
```
mhkawano, line 4756: Not requesting a change since all the other tests are like this too, but I think it would be better to keep act / assert blocks separated.
go/unit-testing-practices#structure

src/runtime/auto-prompt-manager.ts
justinchou-google: Done.

src/runtime/auto-prompt-manager-test.js
justinchou-google: Updated the tests of in this suite for helper functions.

PR #3602, opened 2024-12-17
Add swgVersion URL param to Pay URLs

kathuan:
Also updates chrome driver version + assertions to match updated iframes.

### DIFF
--- a/third_party/gpay/src/payments_web_activity_delegate.js
+++ b/third_party/gpay/src/payments_web_activity_delegate.js
@@ -371,7 +371,7 @@ class PaymentsWebActivityDelegate {
     );
     const opener = this.activities.open(
       GPAY_ACTIVITY_REQUEST,
-      this.getHostingPageUrl_(),
+      this.getHostingPageUrl_(this.getSwgVersion_(paymentDataRequest)),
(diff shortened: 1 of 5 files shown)


third_party/gpay/src/payments_web_activity_delegate.js
justinchou-google, line 1: Just curious, is there a test file for this?

kathuan: Since this file lives in third_party (and is supposedly imported), it doesn't seem to have a test file nor does it seem (? not sure) to be included in the test patterns:

https://github.com/subscriptions-project/swg-js/blob/2584e46f4063b808985325c274b96ca3ef8acb32/build-system/config.js#L50

and these methods I'm changing ultimately get stubbed out in pay-client-test.js. Is the right approach here to add a separate test file + update the test patterns?

justinchou-google: Ah, I thought these changes were manually added, but seems like I misunderstood.

justinchou-google: relevant bug for tracking sync and adding tests: b/178416993

Issue #3605, opened 2024-12-20
Correct docs and error message for `getProductIDFromPageConfig_()`


xurizaemon:
Currently, if a product ID is not obtained from `getProductIDFromJsonLdPageConfig_`, the returned error is:

> Showcase articles must define a **_publisher ID_** with either JSON-LD or Microdata.

The current error appears to misdirect integrators.

I believe this (and the function docs) should warn the user about the _product_ ID, not the _publisher_ ID.

https://github.com/subscriptions-project/swg-js/pull/3606

xurizaemon:
Fixed via https://github.com/subscriptions-project/swg-js/pull/3606

PR #3606, opened 2024-12-20
Correct docs and error from getProductIDFromJsonLdPageConfig_()

xurizaemon:
Ref subscriptions-project/swg-js#3605


xurizaemon:
Done @ChrisAntaki

PR #3615, opened 2025-02-26
Always prune timestamps older than 2 weeks

ChrisAntaki:
Improves branch coverage and updates coding style


justinchou-google (approved):

justinchou-google (changes requested):
On second thought, I think we should remove the parameter from https://github.com/subscriptions-project/swg-js/blob/main/src/runtime/auto-prompt-manager.ts#L861 and update this default to be 2 weeks. Currently, we do not allow anything to persist in local storage for over 2 weeks, and this would keep that as the status quo, WDYT?

justinchou-google (approved):
Thanks Chris!!

PR #3637, opened 2025-04-11
Add new pay/play specific payflow events

mborof:


src/runtime/pay-flow.ts
```diff
@@ -123,13 +123,13 @@ export class PayStartFlow {
...
     const clientConfig = await this.clientConfigManager_.getClientConfig();
-    this.start_(clientConfig.paySwgVersion);
+    this.start_(!clientConfig.useUpdatedOfferFlows, clientConfig.paySwgVersion);
   }
 
   /**
    * Starts the payments flow for the given version.
    */
-  private start_(paySwgVersion?: string): void {
+  private start_(useClassicFlow: boolean, paySwgVersion?: string): void {
```
justinchou-google, line 132: Would it be clearer if we didn't have to negate the input parameter, and instead this was `usePayFlow`?

mborof: ok

justinchou-google (approved):

PR #3656, opened 2025-04-24
Update Frequency Capping eligibility to handle localstorage timestamps keyed by configId.

justinchou-google:
In https://github.com/subscriptions-project/swg-js/pull/3651, dual writes for frequency capping events was implemented.

Global frequency cap eligibility is determined by looking at all timestamps other than the current action. This change updates that eligibility check to ignore configId timestamps.


src/runtime/auto-prompt-manager.ts
```diff
@@ -516,11 +516,14 @@ export class AutoPromptManager {
...
       const globalTimestamps = Array.prototype.concat.apply(
         [],
-        Object.entries(actionsTimestamps!).map(([type, timestamps]) =>
-          type === nextOrchestration!.type
-            ? timestamps.completions // Completed repeatable actions count towards global frequency
-            : timestamps.impressions
-        )
+        Object.entries(actionsTimestamps!)
+          // Ignore events keyed by configId before FCA Phase 1 rampup
+          .filter(([key, _]) => key in InterventionType)
```
mhkawano, line 521: I believe the `in` operator will check against the keys of InterventionType. That's not necessarily a problem since the keys and values are the same, but I think we use the values as authoritative and I worry that this could cause a mismatch in the future. May not be a concern if we're going to remove this after the rampup.

mhkawano (approved):

PR #3679, opened 2025-04-28
Add partial support for linking multiple subscriptions

mborof:
Adds underlying code to support linking multiple subscriptions simultaneously.  This avoids adding the new methods to the API so we can test released code internally first.


src/api/subscriptions.ts
```diff
@@ -601,3 +601,24 @@ export interface LinkSubscriptionResult {
   publisherProvidedId?: string | null;
   success: boolean;
 }
+
+export interface SubscriptionLinkRequest {
```
mhkawano, line 605: We should make sure to comment stuff in the API.

src/runtime/subscription-linking-flow-test.js
```diff
@@ -31,6 +34,12 @@ describes.realWin('SubscriptionLinkingFlow', (env) => {
 
   const PUBLICATION_ID = 'pub1';
   const REQUEST = {publisherProvidedId: 'ppid'};
+  const MULTI_REQUEST = {
```
mhkawano, line 37: nit: this looks like it's meant to be immutable, so we should either reset it in before each, or freeze it somehow.

src/runtime/subscription-linking-flow-test.js
```diff
@@ -80,21 +91,71 @@ describes.realWin('SubscriptionLinkingFlow', (env) => {
...
+      let hidden;
+      let dialogConfig;
+      dialogManagerMock
+        .expects('openView')
+        .once()
+        .callsFake((viewArg, hiddenArg, dialogConfigArg) => {
+          activityIframeView = viewArg;
+          hidden = hiddenArg;
+          dialogConfig = dialogConfigArg;
+          return Promise.resolve();
```
mhkawano, line 115: nit: can this just be an async function?

src/runtime/subscription-linking-flow-test.js
```diff
@@ -80,21 +91,71 @@ describes.realWin('SubscriptionLinkingFlow', (env) => {
...
+        });
+
+      subscriptionLinkingFlow.startMultipleLinks(MULTI_REQUEST);
+
+      const url = new URL(activityIframeView.src_);
```
mhkawano, line 120: don't love that we depend on the implementation details of activityIframeView. Can we mock it and use the public interface like other tests?

mborof: What do you think about this instead?

mborof: I was able to implement that, but I had to inject new promises so it could wait for things to finish rendering before running the checks.

mhkawano: LGTM. Thanks!

mhkawano (approved):

PR #3682, opened 2025-04-29
Close iframe on gray pane click

mborof:
b/394900839

### DIFF
--- a/src/components/dialog-test.js
+++ b/src/components/dialog-test.js
@@ -19,6 +19,7 @@ import {Dialog} from './dialog';
 import {GlobalDoc} from '../model/doc';
 import {getStyle} from '../utils/style';
 import {setExperimentsStringForTesting} from '../runtime/experiments';
+import {CloseWindowRequest} from '../proto/api_messages';
 
 const NO_ANIMATE = false;
 const ANIMATE = true;
@@ -42,13 +43,13 @@ describes.realWin('Dialog', (env) => {
     globalDoc = new GlobalDoc(win);
 
     element = doc.createElement('div');
-    element.contentWindow = {postMessage: (message) => (lastMessage = message)};
     view = {
       getElement: () => element,
       init: (dialog) => Promise.resolve(dialog),
       resized: () => {},
       shouldFadeBody: () => true,
       shouldAnimateFade: () => true,
+      execute: (msg) => (lastMessage = msg),
     };
 
     sandbox.stub(self, 'requestAnimationFrame').callsFake((callback) => {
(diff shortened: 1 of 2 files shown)


justinchou-google:
I recall after we first implemented, there was some some concern raised about wanting to allow readers to click through the scrim so that they can visit links like the homepage title link (I suppose in the nondismissible case). Can you confirm this original issue is resolved by "Back to homepage" (I am unable to find references for this).

And if we confirm that we no longer need to click through the scrim, can you add a test that the CTA is not dismissed on scrim click for the blocking CTA case?

mborof:
1) We explicitly disabled scrim click-through.  Always.  This happened awhile ago.
2) We tried to have it close the window on scrim click, there was a problem and it took me awhile to stumble across the fix.
3) There are already tests validating no close message on scrim click if it is not closable.

mborof:
Also, note that this won't work until cl/752857869 is deployed.  There shouldn't be any issues with releasing this first.

justinchou-google:
> 1. We explicitly disabled scrim click-through.  Always.  This happened awhile ago.
My comment was in reference to https://b.corp.google.com/issues/234490590#comment3. I don't have the full context on this worksteam, but it seems like "Back to homepage" was intended to replace this click-through behavior. Is my understanding correct?

mborof:
Yes, but we also didn't add the back to home page button to all the CTAs - just a few of them.

Even if we decide to allow click-through for non-closable screens again we should still have code in place to close the screen when you click the scrim.

justinchou-google:
How many is just a few? To my knowledge we added these to nearly all of the audience action CTAs: [newsletter](https://source.corp.google.com/piper///depot/google3/java/com/google/subscribewithgoogle/client/ui/newsletteriframe/jswire/scuba_goldens/chrome-linux/newsletteriframe_notClosable_renderBackLink_golden_1000x1000_chrome-linux.png;bpv=1;bpt=0), [survey](google3/java/com/google/subscribewithgoogle/client/ui/surveyiframe/jswire/scuba_goldens/chrome-linux/surveyiframe_backlink_golden_1000x1000_chrome-linux.png), [byocta](google3/java/com/google/subscribewithgoogle/client/ui/byoctaiframe/jswire/scuba_goldens/chrome-linux/byoctaiframe_standardized_withBacklink_golden_1000x1000_chrome-linux.png)

> Even if we decide to allow click-through for non-closable screens again we should still have code in place to close the screen when you click the scrim for closable ones.

Makes sense to me, can we just confirm this is working as intended in the dismissible and nondismissible case for these other CTAs (newsletter, survey, regwall, byocta, rewarded ad)?

justinchou-google (approved):

justinchou-google:
Fixes to other CTAs will need to be added to the controller.

PR #3683, opened 2025-04-30
Add inline cta api to swgjs

kristenwang:
Inline cta api will support the rendering of inline cta.

screencast: https://screencast.googleplex.com/cast/NjEzNDE0MjM0ODU1ODMzNnxiMWNiNzVkZC03MQ

Will add to basic-runtime later.

### DIFF
--- a/src/runtime/inline-cta-api.ts
+++ b/src/runtime/inline-cta-api.ts
@@ -1,10 +1,13 @@
 import {ActionToIframeMapping, parseUrl} from '../utils/url';
 import {ActivityIframeView} from '../ui/activity-iframe-view';
 import {ActivityPorts} from '../components/activities';
+import {ClientConfigManager} from './client-config-manager';
 import {Deps} from './deps';
 import {Doc} from '../model/doc';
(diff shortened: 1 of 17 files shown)


src/runtime/inline-cta-api.ts
```diff
@@ -0,0 +1,154 @@
...
+      this.clientConfigManager_,
+      'InlineCta requires an instance of ClientConfigManager.'
+    );
+  }
+
+  private actionToUrlPrefix_(
+    configId: string,
+    actions: Intervention[]
+  ): string {
+    for (const action of actions) {
```
justinchou-google, line 56: nit: maybe somethign like 

```
const action = actions.find((action) => action.configurationId === configId));
return ActionToIframeMapping[action?.type] ?? '';
```

justinchou-google (approved):

src/runtime/inline-cta-api.ts
kristenwang: Will do it in follow up cl! My git it acting weird and cannot push my changes.

PR #3688, opened 2025-05-01
Add rewarded ad responses

mhkawano:
b/412645499


src/proto/api_messages-test.js
justinchou-google, line 1: Just curious, is the formatting being done by your IDE? are you able to save this without formatting changes? (cmd + shift + p > file: save without formatting)? Seems like this file is constantly being uploaded as raw -> reformatted -> raw ...

I guess the documentation does say to apply lint, but is this file the output of that linter?

justinchou-google (approved):

src/proto/api_messages-test.js
mhkawano: I ran it through the linter in g3, then ran it through the presubmit linter for this project. Do we want to skip that step in the future?

mhkawano: I think you need to run the project linter or you'll get caught by presubmit checks.

justinchou-google: To close the loop, we discovered prettier lint was being applied on top of g3 formatting (the file in this PR actually reverted the formatting to without prettier, which should be the desired state) see forward fix to ignore prettier lint: https://github.com/subscriptions-project/swg-js/pull/3690

PR #3692, opened 2025-05-01
Handle completion for inline CTA

kristenwang:
We will remove the inline CTA and show a toast if newsletter or regwall has been completed before.

Otherwise, the inline CTA will remain on the page, to show either the error UI or the completion UI.

Will follow up with completion event recording.


src/runtime/inline-cta-api.ts
```diff
@@ -74,15 +70,76 @@ export class InlincCtaApi {
...
+    if (response.getAlreadyCompleted()) {
+      this.clearInlineCta_(div);
+      this.showAlreadyOptedInToast_(actionType);
+    }
+    const now = Date.now().toString();
+    this.storage_.set(StorageKeys.READ_TIME, now, /*useLocalStorage=*/ false);
+    this.entitlementsManager_.getEntitlements();
+  }
+
+  private showAlreadyOptedInToast_(actionType: string): void {
```
justinchou-google, line 92: Does it make sense to refactor the original function to make it reusable here?

kristenwang: Created a shared util for the function.

justinchou-google (approved):

PR #3701, opened 2025-05-09
Add standard rewarded ad flag

mhkawano:
b/416542509

### DIFF
--- a/src/runtime/auto-prompt-manager.ts
+++ b/src/runtime/auto-prompt-manager.ts
@@ -146,6 +146,7 @@ export class AutoPromptManager {
   private contentType_: ContentType | undefined;
   private shouldRenderOnsitePreview_: boolean = false;
   private dismissibilityCtaFilterExperiment_: boolean = false;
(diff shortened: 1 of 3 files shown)


src/runtime/auto-prompt-manager.ts
```diff
@@ -146,6 +146,7 @@ export class AutoPromptManager {
   private contentType_: ContentType | undefined;
   private shouldRenderOnsitePreview_: boolean = false;
   private dismissibilityCtaFilterExperiment_: boolean = false;
+  private standardRewardedAdExperiment = false;
```
justinchou-google, line 149: nit: The `_` suffix was used for private class properties, but seems like this was a JS practice and may have changed for TS  :go/tsjs-style#identifiers-underscore-prefix-suffix. I'll leave this as optional to you

mhkawano: ack. imo we should move towards removing the suffix

PR #3702, opened 2025-05-09
Allow audience actions to be completed from AudienceActionFlow

mhkawano:
b/412644428

### DIFF
--- a/src/runtime/audience-action-flow.ts
+++ b/src/runtime/audience-action-flow.ts
@@ -550,9 +550,6 @@ export class AudienceActionIframeFlow implements AudienceActionFlow {
   }
 
   private async completeAudienceAction() {
-    if (!!this.params_.shouldRenderPreview) {
-      return;
-    }
     const swgUserToken = await this.deps_
       .storage()
       .get(StorageKeys.USER_TOKEN, true);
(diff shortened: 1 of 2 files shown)


src/runtime/audience-action-flow.ts
```diff
@@ -535,6 +541,53 @@ export class AudienceActionIframeFlow implements AudienceActionFlow {
     this.params_.monetizationFunction?.();
   }
 
+  private async completeAudienceAction() {
```
justinchou-google, line 544: Just curious, are we unable to handle the completion from the controller like the other actions?

mhkawano: For rewarded ads completion happens after an ad finishes. We could handle it through the controller, but we would need to keep the iframe open but hide it during the ad. This is a much simpler solution.

src/runtime/audience-action-flow.ts
```diff
@@ -535,6 +541,53 @@ export class AudienceActionIframeFlow implements AudienceActionFlow {
...
+      url,
+      emptyMessage
+    )) as unknown as DirectCompleteAudienceActionResponse;
+    if (response.updated) {
+      await this.updateEntitlements(response.swgUserToken);
+    }
+    // TODO: mhkawano - else log error
+  }
+
+  private async updateEntitlements(swgUserToken: string | undefined | null) {
```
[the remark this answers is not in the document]
mhkawano: I tried that initially, but it wants all three :(

src/runtime/audience-action-flow.ts
justinchou-google: Makes sense, thanks!

PR #3710, opened 2025-05-12
Update audience action types

mhkawano:
Clean up some of the constants related to audience actions


src/runtime/audience-action-type.ts
```diff
@@ -0,0 +1,19 @@
...
+  InterventionType.TYPE_NEWSLETTER_SIGNUP,
+  InterventionType.TYPE_REWARDED_AD,
+  InterventionType.TYPE_REWARDED_SURVEY,
+] as const;
+
+export type AudienceActionType = (typeof AUDIENCE_ACTION_TYPES_VALUES)[number];
+
+const values = AUDIENCE_ACTION_TYPES_VALUES as ReadonlyArray<InterventionType>;
+
+export function isAudienceActionType(
```
justinchou-google, line 15: Can we add some tests for this?

src/runtime/auto-prompt-manager.ts
```diff
@@ -906,10 +958,10 @@ export class AutoPromptManager {
       this.storeCompletion(action!);
     } else if (GENERIC_COMPLETION_EVENTS.includes(event)) {
       if (this.isContribution_()) {
-        this.storeCompletion(TYPE_CONTRIBUTION);
+        this.storeCompletion(InterventionType.TYPE_CONTRIBUTION);
```
justinchou-google, line 961: Is this going to impact frequency capping, particularly how the keys are written? The current tests setup frequency capping using raw strings: https://github.com/subscriptions-project/swg-js/blob/main/src/runtime/auto-prompt-manager-test.js#L213

As the current implementation, I am not 100% sure if the tests cover this conversion of `InterventionType` to the string used as the localstorage key.

src/runtime/audience-action-type.ts
mhkawano: IMO it would pretty much just be a test that restates the type definition, which I don't think would be very helpful. Ideally these would get tested in the users of the function.

src/runtime/auto-prompt-manager.ts
mhkawano: These are enum strings, so I don't think it should impact how its written unless someone changes the enum value. Would it be better if disentangled the article names from the storage names?

src/runtime/audience-action-type.ts
justinchou-google: I think it would be helpful to show via tests that we are explicitly defining this subset of InterventionTypes to exclude Contribution and Subscription, but I'll defer this decision to you.

src/runtime/auto-prompt-manager.ts
justinchou-google: Actually, it seems like the enum -> string conversion _is_ being tested via the set expectation: https://github.com/subscriptions-project/swg-js/blob/main/src/runtime/auto-prompt-manager-test.js#L4438

src/runtime/audience-action-type.ts
mhkawano: Ack added test

PR #3715, opened 2025-05-12
Render old UI for Subscription Linking

mborof:
Use the old query string parameter (ppid) for the currently released subscription linking workflow.  This causes the system to render the old UI instead of the new UI (which is not yet approved).

### DIFF
--- a/src/runtime/subscription-linking-flow.ts
+++ b/src/runtime/subscription-linking-flow.ts
@@ -169,13 +169,78 @@ export class SubscriptionLinkingFlow {
       throw new Error('Missing required field: publisherProvidedId');
     }
     const publicationId = this.pageConfig_.getPublicationId();
-    return this.startMultipleLinks({
-      linkTo: [{publicationId, publisherProvidedId}],
-    }).then((result) => {
-      return {
-        success: !result.anyFailure,
-        publisherProvidedId: result.links[0].publisherProvidedId,
-      };
+    const args = feArgs({publicationId});
+    const activityIframeView = new ActivityIframeView(
+      this.win_,
+      this.activityPorts_,
+      feUrl('/linksaveiframe', {
+        subscriptionLinking: 'true',
+        ppid: publisherProvidedId,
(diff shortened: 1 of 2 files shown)


src/runtime/subscription-linking-flow.ts
```diff
@@ -169,13 +169,78 @@ export class SubscriptionLinkingFlow {
...
+      this.activityPorts_,
+      feUrl('/linksaveiframe', {
+        subscriptionLinking: 'true',
+        ppid: publisherProvidedId,
+      }),
+      args,
+      /* shouldFadeBody= */ false
+    );
+
+    activityIframeView.onCancel(() => {
```
mhkawano, line 184: Optional: what do you think of splitting these handlers into their own methods?

src/runtime/subscription-linking-flow.ts
```diff
@@ -169,13 +169,78 @@ export class SubscriptionLinkingFlow {
...
+
+    activityIframeView.on(
+      SubscriptionLinkingCompleteResponse,
+      (response: SubscriptionLinkingCompleteResponse) => {
+        const completionStatus = response.getSuccess()
+          ? AnalyticsEvent.EVENT_SUBSCRIPTION_LINKING_SUCCESS
+          : AnalyticsEvent.EVENT_SUBSCRIPTION_LINKING_FAILED;
+
+        this.deps_.eventManager().logSwgEvent(completionStatus);
+        const success = response.getSuccess() ?? false;
```
mhkawano, line 204: nit: `!!response.getSuccess()`

mhkawano (approved):

src/runtime/subscription-linking-flow.ts
mborof: I think I am going to delete these in fairly short order when we finish migrating to the new UI and that it's not really necessary for any reason.  Perhaps the ones in startMultipleLinks should get migrated to their own functions when we clean that up.

PR #3727, opened 2025-05-19
share handleSurveyDataTransferRequest between inline cta and audienceactionflow

kristenwang:
b/417699932


src/utils/cta-utils.ts
```diff
@@ -49,3 +60,165 @@ export function showAlreadyOptedInToast(
   }
   new Toast(deps, feUrl('/toastiframe', urlParams)).open();
 }
+
+/* Supports survey data tansfer, log survey events and store PPS values.*/
+export async function handleSurveyDataTransferRequest(
+  request: SurveyDataTransferRequest,
+  deps: Deps,
+  activityIframeView: ActivityIframeView,
+  configurationId: string | undefined,
```
mhkawano, line 69: Seems like config id should always be available? Maybe this is an artifact from the past. Also nit `configurationId?: string`. Same with `attemptSurveyDataTransfer`.

src/runtime/audience-action-flow-test.js
```diff
@@ -867,528 +781,12 @@ describes.realWin('AudienceActionIframeFlow', (env) => {
...
-      .once();
-
-    const messageCallback = messageMap[TEST_SURVEYDATATRANSFERREQUEST.label()];
-    messageCallback(TEST_SURVEYDATATRANSFERREQUEST);
-
-    await tick(10);
-
-    expect(self.console.warn).to.have.been.calledWithExactly(
-      '[swg.js] Exception in publisher provided logging callback: Error: Test Callback Exception'
+  it('handleSurveyDataTransferRequest called on SurveyDataTransferRequest', async () => {
```
mhkawano, line 784: If you're going to mock the handler can we remove `storageMock`?

src/utils/cta-utils-test.js
```diff
@@ -13,28 +13,124 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
+import {ActivityIframeView} from '../ui/activity-iframe-view';
```
mhkawano, line 16: This is a lot of code so what do you think of moving it to something like `survey-utils.ts`?

mhkawano (approved):

src/runtime/audience-action-flow-test.js
kristenwang: Cannot remove storageMock in the tests here since storage is still being accessed in audience-action-flow through `this.deps_.storage()`

src/utils/cta-utils.ts
kristenwang: Done.

src/utils/cta-utils-test.js
kristenwang: Sounds good!

PR #3730, opened 2025-05-20
Add details for SL logging

mborof:


mhkawano (changes requested):
Seems like one of the tests needs to be updated based on the proto change. I'm also not sure why `api_messages-test.js` diffs the whole file. Seems the same though.

mborof:
I updated the test that broke.  I'm not clear on why the diff is so large - I copied it out of the auto generated code.  I could try creating a whole new branch if you want and seeing if that takes care of it.

mhkawano (approved):

PR #3733, opened 2025-05-22
Remove rewarded ads code

mhkawano:
b/412642510


justinchou-google:
Is the boq implementation fully ramped up?

mhkawano:
> Is the boq implementation fully ramped up?

Not yet, but I plan to ramp it up today.

mhkawano:
> > Is the boq implementation fully ramped up?
> 
> Not yet, but I plan to ramp it up today.

That did remind me that I should remove the branch into this code. Will do that now.

justinchou-google:
> > Is the boq implementation fully ramped up?
> 
> Not yet, but I plan to ramp it up today.

Is the plan to merge this code immediately? Would it be safer to let the boq changes bake for a few releases in case we need to fallback to the swg rendering?

mhkawano:
> > > Is the boq implementation fully ramped up?
> > 
> > 
> > Not yet, but I plan to ramp it up today.
> 
> Is the plan to merge this code immediately? Would it be safer to let the boq changes bake for a few releases in case we need to fallback to the swg rendering?

Yeah I don't plan to merge immediately.

src/runtime/audience-action-local-flow.ts
```diff
@@ -267,8 +211,8 @@ export class AudienceActionLocalFlow implements AudienceActionFlow {
 
     const validNewsletterSignupParams =
       codeSnippet &&
-      config?.optInParameters?.promptPreference ===
-        PREFERENCE_PUBLISHER_PROVIDED_PROMPT;
+      config?.optInParameters?.promptPreference ==
```
justinchou-google, line 214: Should we keep triple equality `===` here?

src/runtime/auto-prompt-manager-test.js
```diff
@@ -4348,30 +4344,6 @@ describes.realWin('AutoPromptManager', (env) => {
         .resolves(clientConfig)
         .once();
       getArticleExpectation = entitlementsManagerMock.expects('getArticle');
-      getArticleExpectation
```
justinchou-google, line 4351: perhaps we should add a rewarded ad case for the non-local flow test?

src/runtime/audience-action-local-flow.ts

src/runtime/auto-prompt-manager-test.js
mhkawano: Added a test to verify that it calls into the non-local flow.

PR #3754, opened 2025-06-10
Share getContributionsUrl between inline and pop up CTA

kristenwang:
This will render contribution CTA inline: https://screenshot.googleplex.com/BMZuK8nTGEUj9Mv


src/runtime/inline-cta-api.ts
```diff
@@ -130,11 +131,30 @@ export class InlineCtaApi {
...
-      'supportsEventManager': true,
-      'productType': DEFAULT_PRODUCT_TYPE,
-    });
+    const fetchUrl =
+      action.type === InterventionType.TYPE_CONTRIBUTION
+        ? getContributionsUrl(
+            clientConfig,
+            this.clientConfigManager_,
+            this.deps_.pageConfig(),
+            /* ctaMode */ true
```
justinchou-google, line 140:
```suggestion
            /* isInlineCta */ true
```

src/utils/cta-utils.ts
```diff
@@ -50,3 +53,39 @@ export function showAlreadyOptedInToast(
...
+  clientConfig: ClientConfig,
+  clientConfigManager: ClientConfigManager,
+  pageConfig: PageConfig,
+  isInlineCta: boolean = false
+): string {
+  if (!clientConfig.useUpdatedOfferFlows) {
+    return feUrl('/contributionsiframe');
+  }
+
+  if (clientConfigManager.shouldForceLangInIframes()) {
```
justinchou-google, line 70: Can we simplify this section with something like:
```suggestion
  params = {'publicationId': pageConfig.getPublicationId()};
  if (clientConfigManager.shouldForceLangInIframes()) {
    params['hl'] = clientConfigManager.getLanguage();
  }
  if (isInlineCta) {
    params['ctaMode'] = 'CTA_MODE_INLINE';
  }
  return feUrl(/contributionoffersiframe', params);
```

src/runtime/inline-cta-api.ts
kristenwang: Done.

PR #3762, opened 2025-06-16
Only transition animate dialog on center position case

justinchou-google:
b/419286123

Removing this transition in the case where the dialog is not required to be in the center seems to improve CLS when tested locally. The original transition should only have a visual effect when the dialog is to be displayed in the center.

Demo that there is no visible changes to non-center position case: https://screencast.googleplex.com/cast/NTM2MDQ1MTUxNTg0MjU2MHw3MDc1OWEwNS0wMA


src/components/dialog.ts
```diff
@@ -428,16 +428,18 @@ export class Dialog {
...
-        this.getElement(),
-        {
-          'transform': this.getDefaultTranslateY_(),
-          'opacity': '1',
-          'visibility': 'visible',
-        },
-        300,
-        'ease-out'
-      );
+      if (this.shouldPositionCenter_()) {
```
mhkawano, line 431: Should this be an else with the above animation? Seems to both are doing similar things.

mhkawano: I'm wondering if it should be something like:
```
   if (this.shouldPositionCenter_()) {
     await transition(
       this.getElement(),
        {
          'transform': this.getDefaultTranslateY_(),
          'opacity': '1',
          'visibility': 'visible',
        },
        300,
        'ease-out'
      );
} else {
    await transition(setImportantStyles(this.getElement(), {
      'transform': 'translateY(100%)',
      'opactiy': '1',
      'visibility': 'visible',
    }));
}
```
We have two animations being applied, that are nearly identical, so maybe we should only do one at a time? We could probably even do something like ` 'transform':  this.shouldPositionCenter_() ? this.getDefaultTranslateY_() 'translateY(100%)'`. Maybe not a big deal either way since it will overwrite the styles.

justinchou-google: Seems like the transition is dependent on the previously set style: https://github.com/subscriptions-project/swg-js/blob/4139736ad77c6693a0c1c599525a09dc18dc8df7/src/utils/animation.ts#L53.

For this, I am inclined to leave the initial `setImportantStyles`. Does that sound ok to you?

justinchou-google: Followed up offline, further improvements will be reserved for future fixes.

PR #3763, opened 2025-06-16
Render subscription iframe for inline cta

kristenwang:
Tested locally: https://screenshot.googleplex.com/4DCvo6N74PQp9Cw


src/runtime/inline-cta-api-test.js
```diff
@@ -441,6 +445,51 @@ describes.realWin('InlineCtaApi', (env) => {
 
       expect(handleSurveyDataTransferRequestSpy).to.be.called;
     });
+
+    it('opens iframe for subscription', async () => {
+      callbacksMock.verify();
+      win.document.body.removeChild(newsletterSnippet);
```
mhkawano, line 451: Does this line do anything? I see it in other tests too. If the tests are meant to have different setups should we split it into different describe blocks?

mhkawano (approved):

src/runtime/inline-cta-api-test.js
kristenwang: The default for the tests are to render newsletterSnippet. And for some tests, we need to remove the newsletterSnippet. I think this is simpler, compared to creating a set of different tests, because lots of other set ups are the same, and creating a different set of tests means we need to repeat those. Or we can add on the code snippet at the start for each test, which would also add more code. Therefore from what I'm seeing, the current set up is the simplest. I would keep this in mind and change the set up if we are adding more tests that doesn't need the newsletterSnippet.

src/utils/cta-utils.ts
```diff
@@ -101,3 +102,40 @@ export function startPayFlow(deps: Deps, response: SkuSelectedResponse): void {
...
+
+  if (clientConfig.uiPredicates?.purchaseUnavailableRegion) {
+    params['purchaseUnavailableRegion'] = 'true';
+  }
+
+  if (isInlineCta) {
+    params['ctaMode'] = 'CTA_MODE_INLINE';
+  }
+
+  return feUrl('/subscriptionoffersiframe', params);
```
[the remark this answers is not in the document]
kristenwang: We can use the values in ActionToIframeMapping. But since this string value only appears once here, I'm not sure we need to use the enum. I think it would make since if it's used in multiple places.

src/runtime/inline-cta-api-test.js
mhkawano: ack

src/utils/cta-utils.ts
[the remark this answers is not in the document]
mhkawano: I don't think using `ActionToIframeMapping` is necessary since we already know the url, but having something like this would be nice:
```
enum EndpointUrls {
  SUBSCRIPTION_OFFERS = "/offersiframe",
```
That way we're not repeating the strings in different places. We could use the enum for `ActionToIframeMapping` too.

PR #3768, opened 2025-06-18
Add isMonetizationClosable_

mhkawano:
When the monetization cta is used as an alternate action, it uses the dismissal settings of the CTA that it was diverted from. We decided that it should continue to use its own.

Tested locally, but don't have a good way to write a test for this right now.

b/425609829

### DIFF
--- a/src/runtime/auto-prompt-manager.ts
+++ b/src/runtime/auto-prompt-manager.ts
@@ -188,6 +188,7 @@ export class AutoPromptManager {
   private lastAudienceActionFlow_: AudienceActionFlow | null = null;
   private configId_?: string;
   private isClosable_?: boolean;
+  private isMonetizationClosable_?: boolean;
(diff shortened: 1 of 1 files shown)


mhkawano:
> One question I have: dismissibility is used to determine whether frequency capping timestamp should be logged. If we are recording separate dismissibility for monetization CTAs, do we also need to update how the frequency capping timestamp is logged for these events?

Had not considered that, but it looks like this is based on `contentType_`, which is independent of the individual CTA setting. Seems like it's fine to keep it as it is?

justinchou-google (approved):
Yes, you are completely right. I do think it would be helpful to add tests to cover the execution of alternate actions, but I don't want that to block this. Can we create a bug to track this?

mhkawano:
Thanks. Added this to track making a test: b/426566448

I'm thinking I want to revisit after refactoring the auto-prompt-manager.

PR #3777, opened 2025-06-26
render pay confirmation iframe inline where the code snippet is

kristenwang:
Right now the pay confirmation iframe would render in an dialog, we need to render it where the pay iframe is.

b/418814334 and b/424789518


justinchou-google (approved):
Cool, can we add a screenshot of a local demo?

kristenwang:
> Cool, can we add a screenshot of a local demo?

The payment flow doesn't work in local demo unfortunately, will need to test in qual. I'll be onduty next week, and I'll test when releasing the new version of swgjs.

PR #3781, opened 2025-07-02
bake in dismissibility experiment

kathuan:
feature has been launched for > 2 months

### DIFF
--- a/src/runtime/auto-prompt-manager-test.js
+++ b/src/runtime/auto-prompt-manager-test.js
@@ -3697,10 +3697,7 @@ describes.realWin('AutoPromptManager', (env) => {
     });
 
     describe('when reader cannot purchase', () => {
-      const createArticleWithDismissibilityFilterExperiment = (
-        intervention,
-        closability
-      ) => ({
+      const createArticle = (intervention, closability) => ({
         audienceActions: {actions: [intervention], engineId: '123'},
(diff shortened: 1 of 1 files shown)


src/runtime/auto-prompt-manager-test.js
```diff
@@ -3696,7 +3696,7 @@ describes.realWin('AutoPromptManager', (env) => {
       expect(autoPromptManager.isInDevMode_).to.be.false;
     });
 
-    describe('when dismissibility filter experiment enabled', () => {
+    describe('when reader cannot purchase', () => {
       const createArticleWithDismissibilityFilterExperiment = (
```
mhkawano, line 3700: We should rename this. Is there maybe an existing def we can use?

src/runtime/auto-prompt-manager-test.js
```diff
@@ -3696,7 +3696,7 @@ describes.realWin('AutoPromptManager', (env) => {
       expect(autoPromptManager.isInDevMode_).to.be.false;
     });
 
-    describe('when dismissibility filter experiment enabled', () => {
```
mhkawano, line 3699: Seems like the describe should be removed and the tests unindented if it's no longer applicable.

mhkawano (approved):

PR #3785, opened 2025-07-07
Set pay confirmation iframe width to be 100%

kristenwang:
Before: https://screenshot.googleplex.com/4RqZ9rV5j8zhkmR
After: https://screenshot.googleplex.com/BgxResKJsVpGmqb

Similar code in inline-cta-api: https://github.com/subscriptions-project/swg-js/blob/main/src/runtime/inline-cta-api.ts#L192-L194


mhkawano (approved):
LGTM, but can we add a test for this? Seems like setting the iframe's style is an important part of the spec / output.

PR #3790, opened 2025-07-10
Change swg-disable-scroll to apply to the root instead of body

mhkawano:
Setting `overflow-x: hidden` seems to implicitly set `overflow-y:auto`. This causes the scroll lock to break. You can recreate by adding this to a test page:

```
<style>
  html {
    overflow-x: hidden;
  }
</style>
```

Reported here:
https://wordpress.org/support/topic/articles-behind-reader-revenue-allow-scrolling-behind-membership-popup/
Live example:
https://thebulrushes.com/2025/07/06/kzn-police-boss-mkhwanazi-accuses-police-minister-of-meddling-in-crime-cases/

b/430945678


justinchou-google (approved):
LGTM, just confirming this is WAI when testing locally?

PR #3827, opened 2025-08-08
Add survey logging for pop up cta

kristenwang:
Now we are only logging cta mode for inline cta, adding the logging for pop up as well.

### DIFF
--- a/src/utils/survey-utils.ts
+++ b/src/utils/survey-utils.ts
@@ -197,7 +197,8 @@ async function storePpsValuesFromSurveyAnswers(
   // TODO(caroljli): clearcut event logging
 }
 
-function createCtaModeEventParams(ctaMode: CtaMode): EventParams {
+// Create a EventParams with ctaMode param
+export function createCtaModeEventParams(ctaMode: CtaMode): EventParams {
   const eventParams: EventParams = new EventParams();
   eventParams.setCtaMode(ctaMode);
   return eventParams;


src/utils/survey-utils-test.js
```diff
@@ -217,7 +217,20 @@ describes.realWin('Survey utils', (env) => {
           eventType: AnalyticsEvent.EVENT_SURVEY_DATA_TRANSFER_COMPLETE,
           eventOriginator: EventOriginator.SWG_CLIENT,
           isFromUserAction: true,
-          additionalParameters: null,
+          additionalParameters: new EventParams([
```
mborof, line 220: Could we do this once at the top in a const?  const EVENT_PARAMS = ...

kristenwang: Done

src/utils/survey-utils-test.js
```diff
@@ -170,7 +170,7 @@ describes.realWin('Survey utils', (env) => {
             eventType: AnalyticsEvent.ACTION_SURVEY_DATA_TRANSFER,
             eventOriginator: EventOriginator.SWG_CLIENT,
             isFromUserAction: true,
-            additionalParameters: null,
+            additionalParameters: {ctaMode: CtaMode.CTA_MODE_POPUP},
```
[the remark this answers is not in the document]
kristenwang: This is the same as others. It's just for others the logEvent was triggered through logSwgEvent, where some null values were filled in, and here the logEvent was triggered directly, so no other null values were passed.

src/utils/survey-utils-test.js
```diff
@@ -95,6 +95,16 @@ const TEST_SURVEYDATATRANSFERREQUEST_WITHPPS_NOVALUES =
   new SurveyDataTransferRequest();
 TEST_SURVEYDATATRANSFERREQUEST_WITHPPS_NOVALUES.setStorePpsInLocalStorage(true);
 
+/**
+ * @param {CtaMode} ctaMode
+ * @return {!EventParams}
+ */
+function getEventParams(ctaMode = CtaMode.CTA_MODE_POPUP) {
```
mhkawano, line 102: nit: `get` seems like the wrong prefix. Maybe something like `make`?

src/utils/survey-utils.ts
```diff
@@ -204,8 +197,8 @@ async function storePpsValuesFromSurveyAnswers(
   // TODO(caroljli): clearcut event logging
 }
 
-function getInlineCtaEventParams(): EventParams {
+function getCtaModeEventParams(ctaMode: CtaMode): EventParams {
```
mhkawano, line 200: ditto

src/utils/survey-utils-test.js
mhkawano: can this be imported from utils?

mhkawano (approved):

src/utils/survey-utils-test.js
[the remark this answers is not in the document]
mborof: This sounds super nuanced, like someone might miss it in the future.  Would a static variable with an explanatory name be a good idea?  Or semi-explanatory with a comment to hint future devs that there are two kinds of event parameters to worry about and why?

mborof (approved):
My comment is optional

src/utils/survey-utils.ts
kristenwang: Done

PR #4070, opened 2026-04-11
Update Monetization Prompt Flows for Multi Instance Monetary CTA

minyuj:
b/473060228

**Please only review the second commit in this PR since the first one is in a previous PR**

This PR implements the core logic for the `multiInstanceMonetaryCtaExperiment` in the monetization prompt flows. It ensures that the `configurationId` associated with an inline CTA is correctly propagated to the subscription and contribution offer flows when the experiment is enabled.

This is the second PR in the series for the Multi Instance Monetary CTA. It depends on the foundation laid in PR 1.

### DIFF
--- a/src/runtime/auto-prompt-manager-test.js
+++ b/src/runtime/auto-prompt-manager-test.js
@@ -830,6 +830,30 @@ describes.realWin('AutoPromptManager', (env) => {
     });
   });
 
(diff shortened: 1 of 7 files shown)


mhkawano (approved):
LGTM, but you will need to get codecov passing

PR #4074, opened 2026-04-15
Make gisMode more robust

mhkawano:
There are edge cases where we lose a race condition, leading to gisMode being inaccurate. This helps to mitigate that.

b/502620080


src/runtime/gis/gis-utils.ts
```diff
@@ -19,15 +16,15 @@ export enum GisMode {
...
+  gisInterop?: boolean
 ): GisMode {
   const isGisAllowed = action === InterventionType.TYPE_REGISTRATION_WALL;
   const isSafari =
     /Safari/i.test(win.navigator.userAgent) &&
     !/Chrome|Chromium|Edg/i.test(win.navigator.userAgent);
-  const state = gisInteropManager?.getState();
   const gisConnecting =
-    !!state && state !== GisInteropManagerStates.WAITING_FOR_PING;
+    !!gisInterop || !!gisInteropManager?.isConnectionExpected();
```
kristenwang, line 27: I'm curious about how gisInterop will be used? In which case are we going to set this value in deps_.config to true and in which case to false? Because seems like once we it's set to true, the other value doesn't matter anymore.

mhkawano: It's an optional flag once the experiment flag is turned on(see runtime.ts). When the flag is on we will pretty much always be expecting a connection, so this clarifies that.

kristenwang (approved):

PR #4075, opened 2026-04-16
feat: pass origin parameter to offer URLs when multi-instance monetary CTA experiment is enabled

minyuj:
b/473060228

Passing the `origin` to the iframe for calling the /SwgClientService.GetActionConfigurationUi to get the monetary CTA configuration. It has been used in cl/900993729.

### DIFF
--- a/src/runtime/contributions-flow-test.js
+++ b/src/runtime/contributions-flow-test.js
@@ -266,6 +266,38 @@ describes.realWin('ContributionsFlow', (env) => {
     await contributionsFlow.start();
   });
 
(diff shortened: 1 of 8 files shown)


kristenwang (commented):
Would to good to add the reason why we need to pass the origin in the desciprion, and the corresponding cl that will handle the origin passed from here.

minyuj:
> Would to good to add the reason why we need to pass the origin in the desciprion, and the corresponding cl that will handle the origin passed from here.

Done, thanks!

kristenwang (approved):

PR #4079, opened 2026-04-21
feat: add isFromButton parameter to subscription and contribution flow configurations

minyuj:
cl/904012934 and cl/904014344 handle the new `isFromButton` param

### DIFF
--- a/src/api/subscriptions.ts
+++ b/src/api/subscriptions.ts
@@ -555,6 +555,11 @@ export interface OffersRequest {
    * can be shown to the readers.
    */
   configurationId?: string;
+
+  /**
+   * Optional. Whether the request was triggered from a button click.
+   */
+  isFromButton?: boolean;
 }
 
 export interface LoginRequest {
--- a/src/runtime/basic-runtime-test.js
+++ b/src/runtime/basic-runtime-test.js
@@ -732,6 +732,7 @@ describes.realWin('BasicRuntime', (env) => {
         .expects('showOffers')
         .withExactArgs({
           isClosable: true,
+          isFromButton: true,
         })
         .once();
       await subscriptionButton.click();
@@ -740,6 +741,7 @@ describes.realWin('BasicRuntime', (env) => {
         .expects('showContributionOptions')
         .withExactArgs({
           isClosable: true,
+          isFromButton: true,
         })
         .once();
       await contributionButton.click();
@@ -802,6 +804,7 @@ describes.realWin('BasicRuntime', (env) => {
         .expects('showOffers')
         .withExactArgs({
           isClosable: true,
+          isFromButton: true,
         })
         .once();
       await subscriptionButton.click();
@@ -810,6 +813,7 @@ describes.realWin('BasicRuntime', (env) => {
         .expects('showContributionOptions')
         .withExactArgs({
           isClosable: true,
+          isFromButton: true,
         })
         .once();
       await contributionButton.click();
--- a/src/runtime/basic-runtime.ts
+++ b/src/runtime/basic-runtime.ts
@@ -691,11 +691,13 @@ export class ConfiguredBasicRuntime implements Deps, BasicSubscriptions {
         [ButtonAttributeValues.SUBSCRIPTION]: () => {
           this.configuredClassicRuntime_.showOffers({
             isClosable: true,
+            isFromButton: true,
           });
         },
         [ButtonAttributeValues.CONTRIBUTION]: () => {
           this.configuredClassicRuntime_.showContributionOptions({
             isClosable: true,
+            isFromButton: true,
           });
         },
       }
--- a/src/runtime/contributions-flow-test.js
+++ b/src/runtime/contributions-flow-test.js
@@ -298,6 +298,40 @@ describes.realWin('ContributionsFlow', (env) => {
     await contributionsFlow.start();
   });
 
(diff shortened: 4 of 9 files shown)


kristenwang:
Curious about why do we need to add the "isFromButton" parameter?

minyuj:
> Curious about why do we need to add the "isFromButton" parameter?

It's used to tell the offer iframe that the request is triggered by a SwG button. So even if there's no configuration ID given to the iframe, it will find the corresponding AAC for the offers to be displayed in the carousel. 

More details in the [design](https://docs.google.com/document/d/1-Wui-H70n2cIBJgZrBJQtyLitcB7MO142Aimk57Mjs0/edit?tab=t.0#heading=h.l2vgdu3s6djt).

kristenwang:
> > Curious about why do we need to add the "isFromButton" parameter?
> 
> It's used to tell the offer iframe that the request is triggered by a SwG button. So even if there's no configuration ID given to the iframe, it will find the corresponding AAC for the offers to be displayed in the carousel.
> 
> More details in the [design](https://docs.google.com/document/d/1-Wui-H70n2cIBJgZrBJQtyLitcB7MO142Aimk57Mjs0/edit?tab=t.0#heading=h.l2vgdu3s6djt).

A few more questions, I wonder if we can just get the configuration id from the [interventionFunnel](https://github.com/minyuj/swg-js/blob/f1abaff0afff8cf86f68da3d5dec7e3d1d2753ad/src/api/action-orchestration.ts#L24) value we get from the article end point? That way, we don't need to call the GetTargetedInterventionFunnelAction from the iframe. Also for multi-instance monetary cta, we are omitting the config id here: [inline-cta-api](https://github.com/minyuj/swg-js/blob/592c2f76afb0ec9cd7ae44fdfbee86a9a6e76d46/src/runtime/inline-cta-api.ts#L174-L176), I'm not sure why?

minyuj:
> > > Curious about why do we need to add the "isFromButton" parameter?
> > 
> > 
> > It's used to tell the offer iframe that the request is triggered by a SwG button. So even if there's no configuration ID given to the iframe, it will find the corresponding AAC for the offers to be displayed in the carousel.
> > More details in the [design](https://docs.google.com/document/d/1-Wui-H70n2cIBJgZrBJQtyLitcB7MO142Aimk57Mjs0/edit?tab=t.0#heading=h.l2vgdu3s6djt).
> 
> A few more questions, I wonder if we can just get the configuration id from the [interventionFunnel](https://github.com/minyuj/swg-js/blob/f1abaff0afff8cf86f68da3d5dec7e3d1d2753ad/src/api/action-orchestration.ts#L24) value we get from the article end point? That way, we don't need to call the GetTargetedInterventionFunnelAction from the iframe. Also for multi-instance monetary cta, we are omitting the config id here: [inline-cta-api](https://github.com/minyuj/swg-js/blob/592c2f76afb0ec9cd7ae44fdfbee86a9a6e76d46/src/runtime/inline-cta-api.ts#L174-L176), I'm not sure why?

Thanks for the thorough review!

Regarding question #1: the idea could work for subscription pub in closed content funnel and contribution pub in open content funnel. But it wouldn't work well for subscription pub in open content funnel because it's possible that the returned interventionFunnel does not include a subscription CTA.

For question #2: I wasn't sure if I understand it correctly, but the change was to add configuration id instead of omitting it when the multi-instance mon cta is enabled. It was returning empty string prior to my [PR](https://github.com/minyuj/swg-js/commit/68ac62fbf3b00095d653cb5223728bae9e4f7a4f#diff-5bb18a90d0cb02c9050cc2ee7ee6a35200f42ea780e4e99530963de6d9c51c1aL164) IIRC.

kristenwang:
Sorry misunderstood for question 2!

> Regarding question #1: the idea could work for subscription pub in closed content funnel and contribution pub in open content funnel. But it wouldn't work well for subscription pub in open content funnel because it's possible that the returned interventionFunnel does not include a subscription CTA.

Following up on this, IIUC, if publisher removed contribution CTA in the open funnel, then no CTA would show up after clicking the button, correct? As for subscription pub, I wonder when we have multiple funnels under the "closed" content type, how will we decide which CTA to show up? 

I'm asking because for contribution CTA, it seems that using the [interventionFunnel](https://github.com/minyuj/swg-js/blob/f1abaff0afff8cf86f68da3d5dec7e3d1d2753ad/src/api/action-orchestration.ts#L24) value would yield the same result, then we don't need to make the call twice. And for Subscription CTA, we can use the only active Subscription CTA returned in `audienceActions` if it's an open content funnel when there's no subscription CTA in that funnel. And we can think of coming back to this when we want to launch multi funnel.

minyuj:
>IIUC, if publisher removed contribution CTA in the open funnel, then no CTA would show up after clicking the button, correct?

That is correct, yes!

>I'm asking because for contribution CTA, it seems that using the [interventionFunnel](https://github.com/minyuj/swg-js/blob/f1abaff0afff8cf86f68da3d5dec7e3d1d2753ad/src/api/action-orchestration.ts#L24) value would yield the same result, then we don't need to make the call twice. And for Subscription CTA, we can use the only active Subscription CTA returned in audienceActions if it's an open content funnel when there's no subscription CTA in that funnel. And we can think of coming back to this when we want to launch multi funnel.

That makes sense, I've created a different [PR](https://github.com/subscriptions-project/swg-js/pull/4084) to implement this idea, thanks for the suggestion!

PR #4084, opened 2026-04-24
feat: extract and pass configurationId to subscription and contribution flows in basic runtime

minyuj:


src/runtime/basic-runtime.ts
```diff
@@ -679,6 +680,45 @@ export class ConfiguredBasicRuntime implements Deps, BasicSubscriptions {
...
+            intervention.type === InterventionType.TYPE_SUBSCRIPTION
+        )?.configId;
+      } else {
+        // Open Content - Get contribution from funnel, subscription from audienceActions
+        contributionConfigId = interventions?.find(
+          (intervention) =>
+            intervention.type === InterventionType.TYPE_CONTRIBUTION
+        )?.configId;
+
+        const funnelSubscriptionIds = interventions
```
kristenwang: I'm kind of confused about the logic here, if a subscription cta exists in the open funnel, we should just show that subscription CTA, right? However it looks like we are filtering it out. Also if the same subscription CTA exists in both open and closed funnel, seems like we are also filtering it out?

minyuj: the decision we had with PM is we want to always use the subscription within the closed content funnel. I've updated the logic, thanks for pointing the issue out!

src/runtime/basic-runtime.ts
```diff
@@ -679,6 +681,58 @@ export class ConfiguredBasicRuntime implements Deps, BasicSubscriptions {
...
+
+      if (isLocked) {
+        // Closed Content - Get subscription from funnel
+        subscriptionConfigId = interventions?.find(
+          (intervention) =>
+            intervention.type === InterventionType.TYPE_SUBSCRIPTION
+        )?.configId;
+      } else {
+        // Open Content - Get contribution from funnel, subscription from audienceActions
+        contributionConfigId = interventions?.find(
```
kristenwang: Can move this logic out of this if()... else().. since contributionconfigid is not affect by it.

src/runtime/basic-runtime.ts
```diff
@@ -679,6 +681,58 @@ export class ConfiguredBasicRuntime implements Deps, BasicSubscriptions {
...
+        } else {
+          const funnelSubscriptionIds = interventions
+            ? interventions
+                .filter((i) => i.type === InterventionType.TYPE_SUBSCRIPTION)
+                .map((i) => i.configId)
+            : [];
+
+          subscriptionConfigId = audienceActions?.find(
+            (action) =>
+              action.type === InterventionType.TYPE_SUBSCRIPTION &&
```
kristenwang: can change to audienceSubscriptionActions.find((action) => action.configurationId &&
            !funnelSubscriptionIds.includes(action.configurationId)
          )?.configurationId;

Also add a comment to explain the logic here,

minyuj: done, thanks

kristenwang (approved):

PR #4102, opened 2026-06-04
add referrer domain param to article endpoint

kristenwang:
Design doc: http://goto.google.com/rrm-segmentation-and-targeting-dd

b/520046276

Also updaing chromedriver version so that e2e tests can pass

### DIFF
--- a/src/runtime/entitlements-manager.ts
+++ b/src/runtime/entitlements-manager.ts
@@ -52,7 +52,12 @@ import {PageConfig} from '../model/page-config';
 import {Storage} from './storage';
 import {StorageKeys} from '../utils/constants';
 import {Toast} from '../ui/toast';
-import {addQueryParam, getCanonicalUrl, parseQueryString} from '../utils/url';
+import {
+  addQueryParam,
+  getCanonicalUrl,
+  parseQueryString,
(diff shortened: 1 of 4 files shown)


src/runtime/entitlements-manager.ts
```diff
@@ -1029,6 +1036,21 @@ function addPreviewKeyToUrl(location: Location, url: string): string {
...
+ */
+function addReferrerDomainToUrl(document: Document, url: string): string {
+  if (!document.referrer) {
+    return url;
+  }
+  const hostname = parseUrl(document.referrer).hostname.replace(/^www\./, '');
+  if (!hostname || hostname === 'localhost') {
+    return url;
+  }
+  return addQueryParam(url, 'referrerDomain', hostname);
```
Wayne-X: nit - is worth naming this in a way that differentiates more strongly between "referrer passed to the swg.js load page" vs "referrer in the HTTP request to OnePlatform server"?

Wayne-X (approved):

src/runtime/entitlements-manager.ts
kristenwang: Done

PR #4119, opened 2026-07-13
Remove mhkawano from CODEOWNERS

mhkawano:


ChrisAntaki:
Looks like this was done in #4141 :pray:

PR #4166, opened 2026-08-05
feat(publisher): implement ESM for publisher.js

everyplace:
This PR adds a new export to the build script for esm bundles, enabling modern import syntax as a module. 
- 100% of existing tests pass, plus some new ones
- Type safety additions
- Export preferredSource instance directly from publisher-main.ts entry point.
- Configure Vite multi-format bundling (IIFE and ES) exclusively for the publisher target without impacting legacy bundles, and opens the door for future splits
- Add unit tests for API return types, .ready() resolution, and autoStart suppression.

### DIFF
--- a/build_binaries.sh
+++ b/build_binaries.sh
@@ -54,15 +54,21 @@ function create_binaries_for_environment() {
     shift 4
 
     for basename in "swg" "swg-basic" "swg-gaa" "publisher"; do
-        # Copy files.
-        cp dist/$basename.template.js dist/$basename$target.js
-        cp dist/$basename.template.js.map dist/$basename$target.js.map
+        for ext in "js" "mjs"; do
+            if [[ ! -f dist/$basename.template.$ext ]]; then
+                continue
+            fi
 
-        # Replace values.
-        sed -i "s|https://FRONTEND.com|$frontend|g"                dist/$basename$target*
-        sed -i "s|___PAY_ENVIRONMENT___|$pay_environment|g"        dist/$basename$target*
-        sed -i "s|___PLAY_ENVIRONMENT___|$play_environment|g"      dist/$basename$target*
-        sed -i "s|$basename.template.js.map|$basename$target.js.map|g" dist/$basename$target*
+            # Copy files.
+            cp dist/$basename.template.$ext dist/$basename$target.$ext
+            cp dist/$basename.template.$ext.map dist/$basename$target.$ext.map
+
+            # Replace values.
+            sed -i "s|https://FRONTEND.com|$frontend|g"                dist/$basename$target.$ext*
+            sed -i "s|___PAY_ENVIRONMENT___|$pay_environment|g"        dist/$basename$target.$ext*
+            sed -i "s|___PLAY_ENVIRONMENT___|$play_environment|g"      dist/$basename$target.$ext*
+            sed -i "s|$basename.template.$ext.map|$basename$target.$ext.map|g" dist/$basename$target.$ext*
+        done
     done
 }
 create_binaries_for_environment \
@@ -83,4 +89,4 @@ create_binaries_for_environment \
 wait
 
 # Remove template binaries.
-rm dist/*template.js*
+rm -f dist/*template.*js*
--- a/src/publisher-main.ts
+++ b/src/publisher-main.ts
@@ -19,14 +19,12 @@
  * The entry point for publisher runtime (publisher.js).
  */
 
-import {INTERNAL_RUNTIME_VERSION} from './constants';
+import {INTERNAL_RUNTIME_VERSION, IS_ESM} from './constants';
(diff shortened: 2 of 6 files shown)


build_binaries.sh
```diff
@@ -54,15 +54,21 @@ function create_binaries_for_environment() {
     shift 4
 
     for basename in "swg" "swg-basic" "swg-gaa" "publisher"; do
-        # Copy files.
-        cp dist/$basename.template.js dist/$basename$target.js
-        cp dist/$basename.template.js.map dist/$basename$target.js.map
+        for ext in "js" "mjs"; do
```
everyplace, line 57: Content below isn't changed, just indented because of the extra loop above.

vite.config.js
```diff
@@ -109,10 +119,24 @@ const builds = {
   publisher: {
     output: args.minifiedPublisherName || 'publisher.js',
     input: './src/publisher-main.ts',
+    esm: true,
```
everyplace, line 127: Only when esm is present and true do mjs versions of the code also get output.

vite.config.js
```diff
@@ -109,10 +119,24 @@ const builds = {
...
   },
 };
 
-const {input, output} = builds[args.target || 'classic'];
+const target = args.target || 'classic';
+const {input, output, esm} = builds[target];
+const outputs = [
+  {
```
everyplace, line 134: This used to be inlined below, but is now higher up to allow for contextually adding the esm version only when relevant.

src/runtime/publisher-runtime.ts
```diff
@@ -142,18 +145,35 @@ export class PublisherRuntime {
   }
 }
 
-export function installPublisherRuntime(win: Window) {
+interface PublisherWindow extends Window {
```
everyplace, line 148: Adds an internal `PublisherWindow` interface for safer type checking against existing runtime installations, and introduces an optional `autoStart` configuration so ESM imports can avoid running immediate DOM side-effects upon evaluation.

src/runtime/publisher-runtime.ts
```diff
@@ -142,18 +145,35 @@ export class PublisherRuntime {
...
+  options?: {autoStart?: boolean}
+): PreferredSourceApi {
   // Only install the Publisher runtime once.
-  const existingProp = (win as unknown as {[key: string]: unknown})
-    .PREFERRED_SOURCE;
+  const existingProp = (win as PublisherWindow).PREFERRED_SOURCE;
   if (existingProp && !Array.isArray(existingProp)) {
-    return;
+    return (
+      existingProp.api ?? {
```
everyplace, line 166: Returns the existing `.api` singleton reference on duplicate script invocations so module exports do not evaluate to undefined, with a defensive no-op fallback to prevent `TypeError` runtime exceptions if the namespace was improperly initialized.

src/runtime/publisher-runtime.ts
```diff
@@ -169,28 +189,34 @@ export function installPublisherRuntime(win: Window) {
...
 
-  // Handle auto-initialization
-  let autoInit = true;
-  const scripts = win.document.querySelectorAll('script');
-  for (let i = 0; i < scripts.length; i++) {
-    const script = scripts[i];
-    if (script.getAttribute('preferred-sources-control') === 'manual') {
-      autoInit = false;
-      break;
+  if (options?.autoStart !== false) {
```
everyplace, line 204: The DOM scanning logic below is unchanged from the original implementation—it is simply indented inside the new `options?.autoStart` check to bypass automatic DOM queries during passive ESM imports.

src/publisher-main.ts
```diff
@@ -19,10 +19,12 @@
...
-import {INTERNAL_RUNTIME_VERSION} from './constants';
+import {INTERNAL_RUNTIME_VERSION, IS_ESM_BUILD} from './constants';
 import {installPublisherRuntime} from './runtime/publisher-runtime';
 import {log} from './utils/log';
 
 log(`Publisher Runtime: ${INTERNAL_RUNTIME_VERSION}`);
 
-installPublisherRuntime(self);
+export const preferredSource = installPublisherRuntime(self, {
+  autoStart: !IS_ESM_BUILD,
```
everyplace, line 29: Only defaults autoStart to true if the build is not ESM. If it is ESM, auto-start is disabled in favor of manual initialization.

build_binaries.sh
```diff
@@ -83,4 +89,4 @@ create_binaries_for_environment \
 wait
 
 # Remove template binaries.
-rm dist/*template.js*
+rm -f dist/*template.*js*
```
everyplace, line 92: Allows for .js and .mjs

everyplace (commented):
Commented on the bigger changes to help explain the overall narrative.

binaryscales (approved):

PR #4177, opened 2026-08-13
Add mixRrmGisTokens utility to call backend action directly

minyuj:
This PR adds a utility function to call the MixRrmGisTokensAction backend endpoint directly from swg-js.

### DIFF
--- a/src/runtime/gis/gis-utils.ts
+++ b/src/runtime/gis/gis-utils.ts
@@ -1,5 +1,11 @@
+import {Deps} from '../deps';
 import {GisInteropManager} from './gis-interop-manager';
 import {InterventionType} from '../../api/intervention-type';
+import {Message} from '../../proto/api_messages';
+import {StorageKeys} from '../../utils/constants';
+import {XhrFetcher} from '../fetcher';
+import {addQueryParam, parseUrl} from '../../utils/url';
+import {serviceUrl} from '../services';
 
 /**
  * The mode of the GIS.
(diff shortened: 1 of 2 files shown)


src/runtime/gis/gis-utils.ts
```diff
@@ -34,3 +40,58 @@ export function getGisMode(
...
+  const fetcher = new XhrFetcher(deps.win());
+  const publicationId = deps.pageConfig().getPublicationId();
+  const swgUserToken = await deps.storage().get(StorageKeys.USER_TOKEN, true);
+  const rrmOrigin = parseUrl(deps.win().location.href).origin;
+
+  const baseUrl = `/publication/${encodeURIComponent(
+    publicationId
+  )}/mixrrmgistokens`;
+
+  let url = serviceUrl(baseUrl);
```
binaryscales, line 71: consider using feUrl:https://github.com/subscriptions-project/swg-js/blob/main/src/runtime/services.ts#L157 for url+queryParam construction

minyuj: Good catch! However, `feUrl` is specifically designed for UI iframes and hardcodes `/ui/v1` into the path (e.g., `.../swg/ui/v1/...`). Since `mixRrmGisTokens` is calling a backend API action rather than opening an iframe, it needs to use `serviceUrl` which points to `/_/api/v1`. If we used `feUrl` here, it would result in a 404 for this endpoint.

binaryscales (approved):

PR #4196, opened 2026-08-27
feat(swg): implement ESM for swg.js

everyplace:
### Description
This PR adds ESM build support for the `swg.js` target, outputting `dist/swg.mjs` (and corresponding `autopush`/`qual` variants) alongside the existing IIFE `dist/swg.js`.

> [!TIP]
> See prior art when ESM for `publisher.js` was added in https://github.com/subscriptions-project/swg-js/pull/4166. Ironically, this PR is even smaller than the previous one, as it had to add more modifications to the build process to allow for optional esm builds on a specific target. This PR takes advantage of this, adding a single line in `vite.config.js` to enable the new build as a result.

This follows the established multi-output pattern used for `publisher.js` / Preferred Sources:
- Enables `esm: true` for `classic` build target in `vite.config.js`.
- Exports `subscriptions` instance from `src/main.ts` and respects `!IS_ESM_BUILD` for passive startup.
- Updates `installRuntime` in `src/runtime/runtime.ts` to return `SubscriptionsInterface`, accept `options?: {autoStart?: boolean}`, and attach a `.ready()` promise and `.api` property to `self.SWG`.
- Adds unit tests in `src/runtime/runtime-test.js`.
- Fixes linter glob filter and excludes `.d.ts` declarations from ESLint.

> [!IMPORTANT]
> As this PR doesn't change any of the existing tests, backwards compatibility with all existing instantiations for the non-esm `swg.js` output is maintained.

### Testing
- Full unit test suite passed (2,042 tests).
- Linter passed cleanly.

### DIFF
--- a/build-system/config.js
+++ b/build-system/config.js
@@ -60,6 +60,8 @@ module.exports = {
     '!src/api/*.js', // Avoid "unused" prefixes in APIs.
     '!src/proto/*.js', // Auto generated code,
     '!{node_modules,build,dist,third_party,tests_output}/**/*.*',
+    '!src/@types/**/*.*',
+    '!**/*.d.ts',
     '!{testing,examples}/**/*.*',
     '!eslint-rules/**/*.*',
     '!test/coverage/**/*.*',
--- a/build-system/config.js
+++ b/build-system/config.js
@@ -60,8 +60,6 @@ module.exports = {
     '!src/api/*.js', // Avoid "unused" prefixes in APIs.
     '!src/proto/*.js', // Auto generated code,
     '!{node_modules,build,dist,third_party,tests_output}/**/*.*',
-    '!src/@types/**/*.*',
-    '!**/*.d.ts',
     '!{testing,examples}/**/*.*',
     '!eslint-rules/**/*.*',
     '!test/coverage/**/*.*',
--- a/build-system/tasks/lint.js
+++ b/build-system/tasks/lint.js
@@ -189,7 +189,7 @@ function eslintRulesChanged() {
  */
 function setFilesToLint(files) {
   config.lintGlobs = config.lintGlobs
-    .filter((e) => e !== '**/*.js')
+    .filter((e) => e !== '**/*.{js,ts}' && e !== '**/*.js')
     .concat(files);
   if (!isCiBuild()) {
     log(green('INFO: ') + 'Running lint on the following files:');
--- a/build-system/tasks/lint.js
+++ b/build-system/tasks/lint.js
@@ -189,7 +189,7 @@ function eslintRulesChanged() {
  */
 function setFilesToLint(files) {
   config.lintGlobs = config.lintGlobs
-    .filter((e) => e !== '**/*.{js,ts}' && e !== '**/*.js')
+    .filter((e) => e !== '**/*.js')
     .concat(files);
   if (!isCiBuild()) {
     log(green('INFO: ') + 'Running lint on the following files:');
--- a/src/@types/globals.d.ts
+++ b/src/@types/globals.d.ts
@@ -20,9 +20,9 @@ declare global {
     SWG:
       | ((api: Subscriptions) => void)[]
       | {
+          api?: Subscriptions;
           push: (callback: (api: Subscriptions) => void) => Promise<void>;
           ready?: () => Promise<Subscriptions>;
-          api?: Subscriptions;
         };
 
     /**
@@ -34,9 +34,9 @@ declare global {
     SUBSCRIPTIONS:
       | ((api: Subscriptions) => void)[]
       | {
+          api?: Subscriptions;
           push: (callback: (api: Subscriptions) => void) => Promise<void>;
           ready?: () => Promise<Subscriptions>;
-          api?: Subscriptions;
         };
 
     /**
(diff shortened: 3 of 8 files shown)


build-system/config.js
```diff
@@ -60,6 +60,8 @@ module.exports = {
     '!src/api/*.js', // Avoid "unused" prefixes in APIs.
     '!src/proto/*.js', // Auto generated code,
     '!{node_modules,build,dist,third_party,tests_output}/**/*.*',
+    '!src/@types/**/*.*',
```
everyplace: Excludes TypeScript declaration files (`.d.ts`) from the ESLint stream. In glob syntax, `@(...)` in `@types` is interpreted as an extglob pattern by `glob-stream`, which causes linter errors if `.d.ts` files are modified.

build-system/tasks/lint.js
```diff
@@ -189,7 +189,7 @@ function eslintRulesChanged() {
  */
 function setFilesToLint(files) {
   config.lintGlobs = config.lintGlobs
-    .filter((e) => e !== '**/*.js')
+    .filter((e) => e !== '**/*.{js,ts}' && e !== '**/*.js')
```
everyplace: When `config.lintGlobs` was migrated from `**/*.js` to `**/*.{js,ts}`, this filter was not updated. As a result, running `gulp lint --local_changes` or `gulp lint --files=...` failed to remove the global glob and was always scanning the entire codebase instead of just the target files.

src/@types/globals.d.ts
```diff
@@ -17,15 +17,27 @@ declare global {
      * The global `SWG` array contains client ready callbacks. Swgjs calls these when its Subscriptions API is ready.
      * https://github.com/subscriptions-project/swg-js/blob/main/docs/embed-client.md#client-ready-callback
      */
-    SWG: ((api: Subscriptions) => void)[];
+    SWG:
```
everyplace, line 20: Mirrors the `publisher.js` change: https://github.com/subscriptions-project/swg-js/pull/4166/changes#r3718179359

src/@types/globals.d.ts
```diff
@@ -17,15 +17,27 @@ declare global {
...
+      | ((api: Subscriptions) => void)[]
+      | {
+          api?: Subscriptions;
+          push: (callback: (api: Subscriptions) => void) => Promise<void>;
+          ready?: () => Promise<Subscriptions>;
```
everyplace, line 25: Open question: does ready? actually need to be optional, or can it actually be defined?

src/@types/globals.d.ts
```diff
@@ -17,15 +17,27 @@ declare global {
      * The global `SWG` array contains client ready callbacks. Swgjs calls these when its Subscriptions API is ready.
      * https://github.com/subscriptions-project/swg-js/blob/main/docs/embed-client.md#client-ready-callback
      */
-    SWG: ((api: Subscriptions) => void)[];
+    SWG:
+      | ((api: Subscriptions) => void)[]
+      | {
+          api?: Subscriptions;
```
everyplace, line 23: Do we actually need the api additional method, or does ready accomplish the same thing?

src/@types/globals.d.ts
```diff
@@ -17,15 +17,27 @@ declare global {
...
+        };
 
     /**
      * The global `SWG` array contains client ready callbacks. Swgjs calls these when its Subscriptions API is ready.
      * https://github.com/subscriptions-project/swg-js/blob/main/docs/embed-client.md#client-ready-callback
      *
      * This global variable is deprecated in favor of the `window.SWG` global variable.
      */
-    SUBSCRIPTIONS: ((api: Subscriptions) => void)[];
+    SUBSCRIPTIONS:
```
everyplace, line 34: Consider using a defined type for this, rather than duplicating.

src/api/subscriptions.ts
```diff
@@ -56,6 +56,11 @@ export interface Subscriptions {
    */
   clear(): Promise<void> | void;
 
+  /**
+   * Returns a promise that resolves to the Subscriptions instance when ready.
+   */
+  ready?(): Promise<Subscriptions>;
```
everyplace, line 62: Same feedback as above, re: is this necessary to be optional?

src/runtime/runtime.ts
```diff
@@ -137,10 +137,14 @@ export function getRuntime(): Runtime {
...
  */
-export function installRuntime(win: Window): void {
+export function installRuntime(
+  win: Window,
+  options?: {autoStart?: boolean}
+): SubscriptionsInterface {
   // Only install the SwG runtime once.
-  if (win[RUNTIME_PROP] && !Array.isArray(win[RUNTIME_PROP])) {
-    return;
+  const existingSwg = win.SWG;
```
everyplace, line 145: Consider using win[RUNTIME_PROP] instead here to be consistent with 175 below.

src/runtime/runtime.ts
```diff
@@ -168,8 +172,8 @@ export function installRuntime(win: Window): void {
 
   // Queue up any callbacks the publication might have provided.
   const waitingCallbacks = ([] as ((api: Subscriptions) => void)[]).concat(
-    win[RUNTIME_PROP],
-    win[RUNTIME_LEGACY_PROP]
+    Array.isArray(win[RUNTIME_PROP]) ? win[RUNTIME_PROP] : [],
```
everyplace, line 175: Open question: is this already checked earlier to determine if this is an array?

PR #4197, opened 2026-08-27
Updates translations

binaryscales:


kristenwang:
Would be great to add the source of these translations, for example a doc or internal codebase, etc. For referencing

kristenwang (approved):

PR #4201, opened 2026-09-01
feat(gis): defers the yield of control to GIS for non-basic publications until getAvailableInterventions is called and returns no interventions

minyuj:
This PR defers the yield of control to GIS for non-basic publications until `getAvailableInterventions` is called and returns no interventions.

### DIFF
--- a/src/runtime/runtime.ts
+++ b/src/runtime/runtime.ts
@@ -789,10 +789,6 @@ export class ConfiguredRuntime implements Deps, SubscriptionsInterface {
             this.eventManager_
           )
         : undefined;
-
-    if (!integr?.isBasic) {
-      this.gisInteropManager_?.yield();
-    }
   }
 
   creationTimestamp(): number {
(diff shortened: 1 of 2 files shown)


src/runtime/runtime.ts
```diff
@@ -1338,7 +1334,14 @@ export class ConfiguredRuntime implements Deps, SubscriptionsInterface {
 
   async getAvailableInterventions(): Promise<AvailableIntervention[] | null> {
     await this.getEntitlements();
-    return this.entitlementsManager().getAvailableInterventions();
+    const interventions =
+      await this.entitlementsManager().getAvailableInterventions();
+
+    if (!interventions || interventions.length === 0) {
```
kristenwang, line 1341: Add a comment to state that non-basic publications will pass this check and we will do GIS interop yield for those pubs.

minyuj: Actually not all non-basic pub will pass this check. Only if the the pub has no active CTAs.

minyuj: added a comment

kristenwang: I'm confused since the description says "defers the yield of control to GIS for **non-basic publications**"?

minyuj: yeah it kinda does both things. it defers the yield and also filter publications that needs to yield. sorry for the confusion

minyuj: i've explained it in the first comment but i updated the title as well.

kristenwang: Ah I see!

kristenwang (approved):

PR #4204, opened 2026-09-01
Update PS Button Demo Site and Set Min Width for Button

binaryscales:


examples/publisher-button.html
```diff
@@ -1,41 +1,271 @@
 <!DOCTYPE html>
-<html>
```
minyuj, line 2: Curious why removing this? Also, the closing `</html>` still remains at the bottom of the file.

examples/publisher-button.html
```diff
@@ -1,41 +1,271 @@
...
   <style>
+    :root {
+      --bg-color: #f8f9fa;
+      --text-color: #202124;
+      --card-bg: #ffffff;
+      --border-color: #dadce0;
+    }
+
     body {
-      font-family: Arial, sans-serif;
```
minyuj, line 8: Same, is there a reason to remove the font-family for body?

examples/publisher-button.html
binaryscales: bad merge, added back

examples/publisher-button.html
binaryscales: added back

src/ui/add-preferred-source-button-test.js
minyuj, line 352: Nit: also add the checks of title and aria-label like other places?

minyuj (approved):
Nit: consider adding a section (or a few constrained boxes, e.g., widths of 140px, 180px, and 220px) to visually verify:
- Ellipsis truncation works as intended across different string lengths.
- Native hover tooltip (title) displays the full string on hover.
- Sizing stops shrinking once it reaches 140px.

